PR #694, opened 2014-05-31
Feature/exposable query dsl

Mpdreamz:
This PR rips the internal state of the fluent descriptors into interfaces and has the fluent descriptors implement them explicitly. It also now comes with default POCO versions of these interfaces. All of this opens the following new features. 
- `Search()` now has two API's
  1. the original fluent lambda syntax that you either love or hate 
  2. A new initializer syntax [see here for an example](https://github.com/elasticsearch/elasticsearch-net/blob/feature/exposable-query-dsl/src/Tests/Nest.Tests.Unit/Search/InitializerSyntax/InitializerExample.cs#L33) of how this looks exactly
- The initializer syntax is way easier to mock
- Queries and filters now deserialize as well as serialize. This means that you can take elasticsearch query json as string and parse it to C# objects. In fact all the tests in the [/Parsers/](https://github.com/elasticsearch/elasticsearch-net/tree/feature/exposable-query-dsl/src/Tests/Nest.Tests.Unit/QueryParsers) namespace actually take the following approach `serialize => deserialize => serialize => deserialize => property assertion` to be extra sure it's safe to roundtrip multiple times. This opens up a wide array of use cases (saving queries/filters being one of them).
- This also means that you can `GET` an alias with a `filter` configured, rename it and `POST` it back with the same filter, the same goes for warmers and other places that can store a query.
- Queries and filters can now be iterated over using [a special visitor](https://github.com/elasticsearch/elasticsearch-net/blob/feature/exposable-query-dsl/src/Tests/Nest.Tests.Unit/QueryParsers/Visitor/VisitorDemoUseCase.cs#L23). I'm still debating if the visitor API should allow you to actively remove or inject queries/filters in to the graph. Any feedback here would be much appreciated!
- Better raw support!, Prior to this PR if all else failed and you needed to inject custom json in the query in one part of a massive query you'd had to jump to `.RawQuery(string)` instead for the entire query. Now you can just use `.Query(q=>q.Term() && q.Raw())` for the part that really needs custom json.
- Range queries/filters now only support the `gt/gte/lt/lte` properties as these are now the only one documented as of Elasticsearch 1.0. 
- Made a general pass over all the properties to see if they were still relevant.
- Fuzziness is now strongly typed to be able represent `AUTO/int distance/double ratio` properly. Use the static `Fuzziness` class.

Fixes: 
- https://github.com/elasticsearch/elasticsearch-net/issues/667
- https://github.com/elasticsearch/elasticsearch-net/issues/581
- https://github.com/elasticsearch/elasticsearch-net/issues/582
- https://github.com/elasticsearch/elasticsearch-net/issues/612
- https://github.com/elasticsearch/elasticsearch-net/issues/687
- https://github.com/elasticsearch/elasticsearch-net/issues/688

### DIFF
--- a/src/Nest/DSL/Aggregations/CardinalityAggregationDescriptor.cs
+++ b/src/Nest/DSL/Aggregations/CardinalityAggregationDescriptor.cs
@@ -15,36 +15,35 @@ namespace Nest
 	public interface ICardinalityAggregator : IMetricAggregator
 	{
 		[JsonProperty("precision_threshold")]
-		int? _PrecisionThreshold { get; set; }
+		int? PrecisionThreshold { get; set; }
 
 		[JsonProperty("rehash")]
-		bool? _Rehash { get; set; }
+		bool? Rehash { get; set; }
 	}
 
 	public class CardinalityAggregator : MetricAggregator, ICardinalityAggregator
 	{
-		public int? _PrecisionTreshold { get; set; }
-		public bool? _Rehash { get; set; }
+		public int? PrecisionThreshold { get; set; }
+		public bool? Rehash { get; set; }
 	}
 
 	public class CardinalityAggregationDescriptor<T> : MetricAggregationBaseDescriptor<CardinalityAggregationDescriptor<T>, T>, ICardinalityAggregator where T : class
 	{
 		private ICardinalityAggregator Self { get { return this; } }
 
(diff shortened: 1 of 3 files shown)


Draekane:
Do we have any examples of the `Query(q=>q.Raw(<someJsonHere>))` that works?  I have seen this, I have even tried using this, but it doesn't seem to actually work, and I can't find any test cases or examples anywhere...

Issue #701, opened 2014-06-05
Cannot replace ContractResolver, even with a passthrough!


bytenik:
Here's an example test case that DOES work:

```
void Main()
{
    var settings = new ConnectionSettings(new Uri("http://ffazva-es001-01.internal.funnelfire.com:9200/"));
    settings.SetJsonSerializerSettingsModifier(x => { });
    var client = new ElasticClient(settings);
    var results = client.GetMany<JObject>(new[] { "Metabase/17783019068" }, "feed", "feedItem");
    if (results.Count() == 0) throw Exception("Uh oh."); // does not throw
}
```

Now here's two different test cases that do NOT work:

```
void Main()
{
    var settings = new ConnectionSettings(new Uri("http://ffazva-es001-01.internal.funnelfire.com:9200/"));
    settings.SetJsonSerializerSettingsModifier(x => { x.ContractResolver = new ExampleContractResolver(x.ContractResolver); });
    var client = new ElasticClient(settings);
    var results = client.GetMany<JObject>(new[] { "Metabase/17783019068" }, "feed", "feedItem");
    if (results.Count() == 0) throw Exception("Uh oh."); // DOES throw every time, even though it exists
}

class ExampleContractResolver : IContractResolver
{
    private readonly IContractResolver _inner;
    public ExampleContractResolver(IContractResolver inner)
    {
        _inner = inner;
    }

    public JsonContract ResolveContract(Type type)
    {
        return _inner.ResolveContract(type);
    }
}
```

and the second example:

```
void Main()
{
    var settings = new ConnectionSettings(new Uri("http://ffazva-es001-01.internal.funnelfire.com:9200/"));
    settings.SetJsonSerializerSettingsModifier(x => { x.ContractResolver = new ExampleContractResolver(settings); });
    var client = new ElasticClient(settings);
    var results = client.GetMany<JObject>(new[] { "Metabase/17783019068" }, "feed", "feedItem");
    if (results.Count() == 0) throw Exception("Uh oh."); // DOES throw every time, even though it exists
}

public class ExampleContractResolver : Nest.Resolvers.ElasticContractResolver
{
    public ExampleContractResolver(IConnectionSettingsValues cs) : base(cs)
    {
    }
}
```

PR #705, opened 2014-06-07
Wrapped the contract after user-adjustable settings

bytenik:


jimsorock:
@Mpdreamz @bytenik is there an example on how to use/override this contract resolver?

bytenik:
What do you mean? What are you trying to accomplish?

bytenik:
@jimsorock What do you mean? What are you trying to accomplish?

jimsorock:
Trying to replace the contract resolver. I know when setting up the client that I can add json converters, but I was wondering how this fixed the issue in #701.

bytenik:
You can replace the resolver with one that passes through requests for contracts for a NEST type to a NEST resolver.

jimsorock:
Ok but where does that replace actually happen? Do I just subclass my own contract resolver from the new SettingsContractResolver?

PR #941, opened 2014-09-11
Fuzziness alignment for MatchQueryDescriptor

andersosthus:
Allows a Match query to take one of the following for Fuzziness:
- .Fuzziness() (for Auto)
- .Fuzziness(double)
- .Fuzziness(int)

This matches the docs at http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-match-query.html#_boolean

I've seen that there are two ways this kind of fuzziness is implemented in Nest.
There is this style that's used in some of the Suggestors, like the FuzzySuggestDescriptor: https://github.com/elasticsearch/elasticsearch-net/blob/develop/src/Nest/DSL/Suggest/FuzzySuggestDescriptor.cs

And then there is another style, basically taking all fuzziness values and calling .ToString() on it so they can all fit in a String field. This is used in the FuzzyQueryDescriptor: https://github.com/elasticsearch/elasticsearch-net/blob/develop/src/Nest/DSL/Query/FuzzyQueryDescriptor.cs

I feel the IFuzziness way is cleaner :)

This PR does not break anything, since the signature still matches the old signature. The main thing I don't like is the changes I had to do to MatchQueryJsonConverter.cs, to have the Fuzziness deserialize properly. I see that there is a FuzzinessConverter already that does the same thing, but not sure how I could use that from the ReadJson in the MatchQueryJsonConverter.


EmilAlipiev:
Is this fix not applied on 1.7? is it only for 2.0.x versions? I would like to know if there is any workaround on 1.7 version because 2.0 version has a lot of changes and it will take so much time to upgrade from 1.7 to 2.0 and I see that it is not very stable. Can you please advise? thanks

PR #994, opened 2014-10-15
FromAsString & ToAsString properties were not serializing

ashkandaie:
The underlying values for from_as_string & to_as_string properties in JSON.NET are DateTime objects therefor the "reader.value as string" would always be null. Change the code to ALWAYS deserialize as string.

Also updated unit tests to validate both FromAsString and ToAsString properties.


Mpdreamz:
Hi @ashkandaie sorry for the delay I've reimplemented your changes manually as per:

https://github.com/elasticsearch/elasticsearch-net/pull/1191

Ty for reporting and sending a PR for this issue :+1: closing this PR in favor of #1191

PR #1043, opened 2014-11-06
aspnetcore50 (vnext core clr support)

Mpdreamz:
WORK IN PROGRESS
----------------

Opening this PR so that we can communicate/collaborate in the open whilst tackling this. 

Steps taken so far: 

* upgrade to Json.NET 6.0.6 which works on `aspnetcore50` (confirmed locally with some preliminary tests)
* Created `Elasticsearch.aspnetcore50.sln` that loads the vnext projects (using separate `kproj/project.json`).
* Include as many packages I could locate to bring the compile errors from 1300+ to 70~.

Challenges:

* ElasticsearchDynamicValue seems to use the most **forbidden** API's 
* We use HttpWebRequest, supported by the most .net platforms but gone in `aspnetcore50` need to ifdef here to utilize HttpClient` Maybe we can ifdef include `Elasticsearch.Net.Connection.HttpClient` somehow.
* `InvariantCulture` missing from `StringComparison`, which seems odd as it can't be worked around opened a ticket here https://github.com/aspnet/Home/issues/177
* The other missing types/methods/extensions can be ifdef'ed around. 
* Update build script so it utilizes a local kvm/kpm/ etcetera to build the project.
* Figure out how to force our unit tests (nunit) to run in aspnetcore50.


Mpdreamz:
closing in favor of https://github.com/elastic/elasticsearch-net/pull/1324

Issue #1080, opened 2014-11-28
Issues with Nest Connection Pool retry mechanism


satishmallik:
Currently nest is not able to differentiate between Service Unavailable and Timeout exception.

Problem description
1) Think of 3 query nodes (q1,q2,q3) in cluster. One of query node q1 is booted out of cluster for some reasone like OOM.
2) Now all query requests to q1 will fail with error status code 503 service not available.
     A typical response where this situation arises look like,

```
             ConnectionStatus = {StatusCode: 503, 
              Method: POST, 
               Url: http://localhost:9200/codeindex/sourceNoDedupeFileContract/_search?routing=0&pretty=true, 
```

3) Only In this case query should be routed to another query node q2.

Currently nest retires all retryable exception like timeouts on all nodes of the connection pool.
So say a wildcard query is sent to query node q1. If it timesout the query should timeout.
But in this scenario nest still sends this query to another node q2. It again timesout and in turn it is sent to q3. So query times out after 3 mins instead of 1 min.

This also makes a case of DOS attack.

Proper fix should be to differentiate between timeout and ServiceNotAvailable exception. Query should be sent on another node from connection pool only if service is not available on first node.

PR #1093, opened 2014-12-05
Do not retry when timeout is hit

Mpdreamz:
this PR adds a safeguard for retrying after the timeout is hit as reported on #1080 . 

This prevents:
- a query that took 60seconds on one node to be retried on a second node.
- a query that timed out after 30seconds on 2 nodes to be retried a third time.

By default this listens to `SetTimeout()` which itself defaults to `60 seconds`. The max retry timeout can be configured separately by calling `SetMaxRetryTimeout()`.

The MaxRetryException message itself clearly reflects when its thrown because we exceeded the `MaxRetryTimeout`.

### DIFF
--- a/src/Elasticsearch.Net/Connection/Configuration/ConnectionConfiguration.cs
+++ b/src/Elasticsearch.Net/Connection/Configuration/ConnectionConfiguration.cs
@@ -61,6 +61,9 @@ namespace Elasticsearch.Net.Connection
 		private int? _maxDeadTimeout;
 		int? IConnectionConfigurationValues.MaxDeadTimeout { get{ return _maxDeadTimeout; } }
 	
+		private TimeSpan? _maxRetryTimeout;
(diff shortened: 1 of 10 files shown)


ankitkinra:
I want confirm behavior of SetTimeout() from pre 1.3 version to 1.3 version. Say we have the following code in Nest Api *1.0*

`var timeoutInMillis = 100;`
`var maxRetries = 3;`
`var config= new ConnectionSettings(pool);`
`config.SetTimeout(timeoutInMillis);`
`config.MaximumRetries(maxRetries);`

Now imagine the scenario where request times out on the first node in 100 milliseconds, the request will repeat for 3 more times with each timeout being 100 milliseconds, when the same code is migrated to Nest 1.3.1 with the above fix, the request will only be tried for 100 milliseconds and will not be retried as by default MaxRetryTimeout will be equal to Timeout, hence this becomes a breaking change.
I wonder if the default value of MaxRetryTimeout could have been maxRetries * timeoutInMillis

PR #1182, opened 2015-01-02
Fix null inner exception when retrying on WebException

### LINKED ISSUE #1176: Inner Exception not available in MaxRetryException in a couple of scenarios (Related to #935 )
Since #935 is closed, I'm opening a new issue. I've added my comments there. (https://github.com/elasticsearch/elasticsearch-net/issues/935)

A simple scenario to simulate the behavior is - 
1. Initialize NEST Client with 3 connection strings (localhost:9300, localhost:9400, localhost:9500) none of which are really ES endpoints
2. Try doing a get Index with that NEST client

gmarz:
When using connection pooling (and retrying), `WebException`s fall out of the normal exception handling in `RequestHandler` and weren't being added to `SeenExceptions`, ultimately causing `MaxRetryException.InnerException` to be null.

Closes #1176


Mpdreamz:
LGTM :+1:

PR #1184, opened 2015-01-06
Fix exception when deserializing relative uris

### LINKED ISSUE #629: Json.NET throws exception while deserializing relative Uri
This bug was introduced with commit https://github.com/elasticsearch/elasticsearch-net/commit/df2bfc99db9862db53f4ca6012ad7b37d57ade7e.

The problem is that the Uri is created without a UriKind, thus implicitly using UriKind.Absolute, which throws an exception for relative Uris.

Is there a compelling reason to replace the standard [non-buggy] serializer?

gmarz:
By removing the custom `UriJsonConverter` which was put in place to serialize `Uri`s to strings by default, however Json.NET already does this out of the box.

Closes #629


Mpdreamz:
@crunchie84 are you available to confirm this still works in your deployed applications?

crunchie84:
@mpdreamz will have a look at it tomorrow!

Issue #1187, opened 2015-01-07
Query<object>.Indices won't Serialize


abibell:
```
var termQuery = Query<object>.Term("field", "valueText");
var indicesQuery = Query<object>.Indices(i => i.Indices(new[] { "test" }));
//indicesQuery won't Serialize
var q = this.SerializeThenDeserialize(
                f => f.Bool,
                f => f.Bool(b => b
                    .Must(termQuery, indicesQuery)
                    .Should(termQuery, indicesQuery))
            );
```
Any ideas on fix options? Happy to help, this would clean our code

Mpdreamz:
Thats because its [conditionless](https://github.com/elasticsearch/elasticsearch-net/blob/develop/src/Nest/DSL/Query/IndicesQueryDescriptor.cs#L57) it needs a query component. The elasticsearch [docs on indices query](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-indices-query.html) says so as well.

So you'd want to pass the term query like so:

```
.Query(q=>
    q.Indices(i=>i
        .Indices(new [] { "test" })
        .Query(iq=>iq.Term("field", "valueText"))
    )
)
```
Read more about conditionless queries here:

http://nest.azurewebsites.net/nest/writing-queries.html

It also describes how to turn off if you really must :) It also goes into great detail on how you can leverage (`&&`, `||` and `!`) to write terser bool queries. 

Hope this helps!

abibell:
I tried to simplify the problem. Here is the actual test where **should** contains information, but it turns out empty. I think we should reopen as I was trying to simplify the problem. The bigger target is. We are trying to enforce security on certain index-type combination (we can have same type name in different indexes) based on users subscriptions. At the moment we use an ugly Json serialisation and SearchDescriptor raw building, too much code for simple thing.

```csharp
[Test]
	    public void IsClientSideSearchQueryDeserialisable()
	    {
	        var newDescriptor = Deserialize<SearchDescriptor<dynamic>>(original);
            var actual = Serialize(newDescriptor);//Should is empty

            var descriptorJobject = JObject.Parse(actual);
	        var expressionList = new[]
	                             {
	                                 new[] {"size"},//Works
	                                 new[] {"from"},//Works
	                                 new[] {"query", "filtered", "query", "query_string", "fields"},//Works
	                                 new[] {"query", "filtered", "filter", "bool", "must"},//Works
//Can't find contents of should
[...]

PR #1190, opened 2015-01-08
Fix/nomatch indices dsl

Mpdreamz:
We were missing the [Indices Filter](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-indices-filter.html) as reported in #1187 

This PR also adds support for `all` and `none` shortcuts on the `no_match_filter` as well as making sure that that the indices json property is written first since the order matters.


gmarz:
LGTM

PR #1191, opened 2015-01-08
fix #994 known strings should be read using ReadAsString() to prevent da...

Mpdreamz:
...tes being parsed by json.net, ty @ashkandaie


gmarz:
LGTM

PR #1195, opened 2015-01-09
Code cleanup. Removed not used namespaces/using

SychevIgor:
In the code was a lot of not used namespaces. It's minor changes in the code without any kind backward compatibility issues.


SychevIgor:
Agreement is signed.

In fact it's not stylistics in my particulate case. We are experimenting with api now and build of the project was integrated in Continuous Integration process. One step of the build is static code analysis. Based on our default rule set usings/namespaces are must be used or not presented in the code. There were 1000+ warning from this and we want to fix and observe other potential issues without such a noise.

As I see now - building CI process is not compatible with our default rules and it's easy just save the code on a disk and don't care about build process because it will take a lot of time.

PR #1196, opened 2015-01-09
Newtonsoft.Json 6.0.7

SychevIgor:
version Json.NET 6.0.1   576217  Saturday, February 01 2014  was released 1 year ago.
6.0.7 is newer Json.NET 6.0.7 (this version)  84591  Tuesday, December 23 2014  
There is no compatibility issues, just reference update.

Also old versions of packages was deleted, because nothing referenced on it.


Mpdreamz:
Great suggestion!

We have a section on our documentation: http://nest.azurewebsites.net/contributing.html but that might not be as discoverable as a file in the root of the repository.

Just added https://github.com/elasticsearch/elasticsearch-net/blob/develop/contributing.md

PR #1198, opened 2015-01-09
Fix thrift and ssl sniffing

Mpdreamz:
This fixes: https://github.com/elasticsearch/elasticsearch-net/issues/1189

It also alters the thrift connection to generate thrift connections for endpoints on demand instead of in the constructor this allows endpoints that are discovered through sniffs to be also be initialized at any time. 

I ran the indexation profiler a dozen time without hiccups and added two integration tests to ensure sniffing works and new nodes are being utilized.


gmarz:
I was running into a `NullReferenceException` in `Sniffer.FromStream()` because the credentials I was testing with existed, but weren't part of a role that had permission to `_nodes/_all`.  In this scenario, apparently ES returns a 403 Forbidden instead of a 401.

I amended the PR with the above commit to throw an `ElasticsearchAuthenticationException` on both 401 and 403.

Other than that this LGTM :+1:   

Feel free to merge if you're good with my last change.

PR #1204, opened 2015-01-19
Check if RequestTimeout has a value instead of ConnectTimeout during Get...

### LINKED ISSUE #1203: HttpConnection only honors RequestTimeout if ConnectTimeout has a value?
I noticed my request timeout wasn't being honored and found this suspicious line of code
https://github.com/elasticsearch/elasticsearch-net/blob/develop/src/Elasticsearch.Net/Connection/HttpConnection.cs#L391

Is that intended?

aochsner:
...RequestTimeout()

Fixes #1203


gmarz:
Thank you @aochsner!  This was indeed not intended, and a great find on your part :+1:.

PR #1205, opened 2015-01-19
Use ConnectTimeout instead of PingTimeout for thrift connections

gmarz:
Which required adding `ConnectTimeout` to `ConnectionSettings`.  Now the timeout options are a bit more consistent between `ConnectionSettings` and `RequestConfiguration`.

Also cleaned up the xmldocs a bit.


Mpdreamz:
LGTM! thank you for catching those xmldoc discrepancies!

PR #1206, opened 2015-01-20
Expose constructor with inner aggregation dictionary

mausch:
When I want to test a function that converts aggregation results into something else I need to set the inner aggregations of a `SingleBucket` in order to set up the test data. This class however doesn't currently expose the constructors of its base classes so it's not possible.


gmarz:
Thanks @mausch.  This LGTM, but we should probably do this for all of the other aggregation response types that inherit from `AggregationsHelper` as well, since they suffer from the same testability issue.  Thoughts?

mausch:
Sure, I only noticed it for this one because that's what I had to use in my test, but it should be same for other similar classes.

PR #1207, opened 2015-01-20
Build improvements

Mpdreamz:
This PR introduces several build improvements.

## Refactored build.fsx

Our build script did a ton in a single file, not a bad thing if you write top quality f# but as a novice I'm not quite there yet. Split our build features and tried to introduce better typing. 

## Move to paket from nuget 

This is somewhat controversial, paket solves a lot of problems nuget will with its 3.0 release as well.  Paket solves them for us *right now* though.

The feature we really needed is that paket does not fixate reference hint paths at install time, based on the current framework, in `csproj` files. It'll resolve them at build time meaning you can just pass `TargetFrameworkVersion` to msbuild and not worry! HUGE! 

We might move back once nuget 3.0 gets released and if it solves all our problems equally well.

**NOTE**: this does not mean we will stop publishing on nuget! paket is simply a better package manager client but does not replace nuget server.

## Proper .NET 4.5 builds

We now do proper .NET 4.5 builds alongside our default .NET 4.0 builds, without introducing seperate csproj files, ty paket!. This means our nuget packages will now work on `aspnet50`! Do note that we still do not support `aspnetcore50` as of yet!

## WIP

- [x] registering fakelib works on windows but not on mono, it needs `#r "FakeLib.dll"` but from visual studio it needs `#r @"../tools/FAKE/tools/FakeLib.dll"` I can switch to the first and have it work on windows too but then visual studio gets upset. 
- [x] Build should work on Windows/OSX and Linux


gmarz:
Nice work @Mpdreamz !  Huge :+1: to the refactoring of the build script.

PR #1210, opened 2015-01-22
Score function weight can be float

mausch:


mausch:
Hmm, it seems this is the same as https://github.com/elasticsearch/elasticsearch-net/pull/1087

Mpdreamz:
This one sucks big time 

We are under very stringent backward compatibility restraints here. We sign our assemblies and only update our assembly file version in between major versions. This is great to avoid assembly binding redirects and allows folks depending on 1.x releases to even "accidently" load a higher dependency version. 

We are releasing 1.4 and after that we have time allotted to go over all the endpoints in elasticsearch plot all the requests and json bodies (hopefully coming up with a proper schema for those) and evaluate our own mappings in preparation of NEST 2.0

All fine and dandy but its not helping you now, I'll look into exposing the Weight as float tomorrow in a backwards compatible way. This might mean a short lived `.WeightAsFloat` property on `IFunctionScoreFunction` although the `.Weight(float)` syntax can still be added.

mausch:
Thanks, I understand the situation. For my present case it was no big deal, since it was a script I was able to simply include a division in there. But I'll probably use other score functions in the near future so it might be an issue.

PR #1211, opened 2015-01-22
Fix/compress body

Mpdreamz:
This PR also allows you to send gzip'd requests, not just receive them. 

When first implementing this I could swear this was not supported (also some information online that indicates its not) but it is at least from Elasticsearch 1.0 to also send gzip'ed data.


gmarz:
LGTM.  All integration tests pass on ES 1.4.2.

PR #1213, opened 2015-01-23
cat api's interface fix #1212

### LINKED ISSUE #1212: CatIndices not supported with IElasticClient
I need to use CatIndices for my requirement. But I need to use ElasticClient for the same instead of IElasticClient. 

Could you please add CatIndices in IElasticClient interface?

Mpdreamz:
Cat Api's were missing from IElasticClient.


gmarz:
LGTM

PR #1214, opened 2015-01-23
Floating weights on function score query / functions

Mpdreamz:
fix #1210 fix #1087 

Both of the previous PR's submitted a sane way to fix this issue but break backward compatibility. 

This PR is a bit of a hack to make sure if you are using the object initializer syntax we don't break your code. 

Fluent syntax remains `.Weight(long|double)`


gmarz:
LGTM :+1: 

Had to fix a few failing unit tests and also added a TODO so we remember to remove this hack in 2.0.

PR #1218, opened 2015-01-25
Added filter support for HasChild filter.

BurlakovNick:
Hey, folks!

I want to use HasChild filter in my project and to filter childs instead of querying them, but miss this functionality in NEST, so I've added a few lines of code.

Do not judge strictly, I'm a newcomer in elasticsearch community:)


gmarz:
LGTM!  Thanks for the PR @BurlakovNick !

PR #1239, opened 2015-02-06
Added support for the Elastic Search Irish language analyzer.

coreycoto:
Elastic Search has support for a built in Irish language analyzer but NEST is missing the enumeration value. 

http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/analysis-lang-analyzer.html#irish-analyzer


gmarz:
Awesome, thanks @coreycoto !  Looks like we're also missing `latvian` and `sorani`...I'll go ahead add those as well.

PR #1243, opened 2015-02-14
Missing execution parameter in RangeFilter

robertlyson:
Hey guys,

This PR contains missing [`execution`](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-range-filter.html#_execution) parameter for `RangeFilter`, [reported on SO](http://stackoverflow.com/questions/28473743/is-it-possible-to-define-the-execution-option-in-range-filter-via-nest).

I've filled `RangeFilterJsonReader` with `WriteJson` and to be consistent I've renamed class to `RangeFilterJsonConverter`. `IRangeFilter` and `RangeFilterDescriptor` have been changed to handle execution parameter.

I have applied EditorConfig - formatting should be fine.

Let me know if something is missing.


Mpdreamz:
Thank you @robertlyson and @UdiBen very much appreciated!

PR #1248, opened 2015-02-16
fix #1247 add support for precision and orientation on geo shape type ma...

### LINKED ISSUE #1247: Support for Precision and Orientation on GeoShapeMappingDescriptor<T>
The documentation for [geo_shape](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/mapping-geo-shape-type.html) types states that one can specify a precision and orientation for the mapping of a geo_shape type field:

**precision**

> This parameter may be used instead of tree_levels to set an appropriate value for the tree_levels parameter. The value specifies the desired precision and Elasticsearch will calculate the best tree_levels value to honor this precision. The value should be a number followed by an optional distance unit. Valid distance units include: in, inch, yd, yard, mi, miles, km, kilometers, m,meters (default), cm,centimeters, mm, millimeters.

**orientation**

> Optionally define how to interpret vertex order for polygons / multipolygons. This parameter defines one of two coordinate system rules (Right-hand or Left-hand) each of which can be specified in three different ways. 1. Right-hand rule (default): right, ccw, counterclockwise, 2. Left-hand rule: left, cw, clockwise. The default orientation (counterclockwise) complies with the OGC standard which defines outer ring vertices in counterclockwise order with inner ring(s) vertices (holes) in clockwise order. Setting this parameter in the geo_shape mapping explicitly sets vertex order for the coordinate list of a geo_shape field but can be overridden in each individual GeoJSON document

Neither are currently exposed on `GeoShapeMappingDescriptor<T>`, but it would be great if there were so that those using NEST can take adavantage of them.

Mpdreamz:
...pping

### DIFF
--- a/src/Nest/Domain/Geo/GeoPrecision.cs
+++ b/src/Nest/Domain/Geo/GeoPrecision.cs
@@ -1,0 +1,22 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Nest.Resolvers.Converters;
+
+namespace Nest
+{
+	[JsonConverter(typeof(GeoPrecisionConverter))]
+	public class GeoPrecision
+	{
+		public double Precision { get; private set; }
+		public GeoPrecisionUnit Unit { get; private set; }
+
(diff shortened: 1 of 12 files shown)


gmarz:
LGTM

PR #1252, opened 2015-02-17
fix #1251 Smarter untyped baseindextypepath

### LINKED ISSUE #1251: NEST - TypeExists ignores default index
When using a TypeExists method without specifying an Index, the DefaultIndex defined in the ElasticClient is ignored and exception "Index() is not specified" is thrown

Mpdreamz:
on untyped IndexTypePaths steal the type's type information and use it to infer index as well if no index has been explicitly set.


gmarz:
LGTM :+1:

PR #1260, opened 2015-02-25
expose SetTcpKeepAlive() on ConnectionSettings 

Mpdreamz:
so you do not need to subclass HttpConnection and override `AlterServicePoint()` to set it.

Sidenote:

I also added an Obsolete on `HttpClientConnection` indicating its not as well tested as `HttpConnection` and as a heads up that with Elasticsearch.NET/NEST 2.0 we'll be moving the default `HttpConnection` over to `HttpClient` anyway. 

Im still not totally sure about moving over to `HttpClient` though, yes its a nicer interface, but there is a real benefit of having a **true** synchronous code path without the overhead of async dispatching even if that happens on IOCP threads. Calling `.Result` on a task is not really equivalent.

We might endup ifdeffing `HttpConnection` specifically for `aspnetcore50`, still lots to research and measure :)


gmarz:
LGTM :+1:

PR #1263, opened 2015-02-25
When using SniffOnStartup() and newing clients

Mpdreamz:
All the time it will resend a sniff on each `new()` :cry: , this fixes that by doing additional bookkeeping on IConnectionpool.

ty @poulfoged for reporting this [on twitter](https://twitter.com/Mpdreamz/status/570590301041123328)


gmarz:
LGTM

Issue #1254, opened 2015-02-20
Thrift Connection - Open connection issue


lukapor:
When I switch to client version 1.4.x (before I was using 1.3.0) have some serius issue with thrift connection. Here I paste stack trace. With 1.3.0 I didn't have no connection issues. Both clients was tested on ES 1.4.4 with thrift plugin version 2.4.1

2015-02-20T07:34:19.86Z Failed after retrying 2 times: 'POST _bulk?consist
ShortMessage: Failed after retrying 2 times: 'POST _bulk?consistency=all'. 
InnerException: TTransportException, InnerMessage: Failed to connect, InnerStackTrace:    at Elasticsearch.Net.Connection.Thrift.Transport.TSocket.Open()
   at Elasticsearch.Net.Connection.Thrift.Transport.TBufferedTransport.Open()
   at Elasticsearch.Net.Connection.Thrift.ThriftConnection.Execute(RestRequest restRequest, Object requestConfiguration)
InnerException: TTransportException, InnerMessage: Failed to connect, InnerStackTrace:    at Elasticsearch.Net.Connection.Thrift.Transport.TSocket.Open()
   at Elasticsearch.Net.Connection.Thrift.Transport.TBufferedTransport.Open()
   at Elasticsearch.Net.Connection.Thrift.ThriftConnection.Execute(RestRequest restRequest, Object requestConfiguration)
InnerException: TTransportException, InnerMessage: Failed to connect, InnerStackTrace:    at Elasticsearch.Net.Connection.Thrift.Transport.TSocket.Open()
   at Elasticsearch.Net.Connection.Thrift.Transport.TBufferedTransport.Open()
   at Elasticsearch.Net.Connection.Thrift.ThriftConnection.Execute(RestRequest restRequest, Object requestConfiguration)
Elasticsearch.Net.Exceptions.MaxRetryException: Failed after retrying 2 times: 'POST _bulk?consistency=all'. 
InnerException: TTransportException, InnerMessage: Failed to connect, InnerStackTrace:    at Elasticsearch.Net.Connection.Thrift.Transport.TSocket.Open()
   at Elasticsearch.Net.Connection.Thrift.Transport.TBufferedTransport.Open()
   at Elasticsearch.Net.Connection.Thrift.ThriftConnection.Execute(RestRequest restRequest, Object requestConfiguration)
[...]

PR #1265, opened 2015-03-03
Develop - problem with opening thrift connection

lukapor:
This was happened when a lot of simulating non framed thrift connection try to open. When we rebuild our indexes (worker threads - builds index).

After one testing day, a exception described in issue #1254 didn't raise.


Mpdreamz:
LGTM thanks @lukapor (also for testing it out for a day!).

PR #1267, opened 2015-03-11
Fix/trace corrolation

Mpdreamz:
Right now we only TRACE when requests are wrapped up (succesfull/failure/timeout etcetera).

Whats actually more useful is to TRACE the fact we are beginning a request in addition to when ending one.


gmarz:
LGTM :+1:

PR #1269, opened 2015-03-11
fix #995 dont expose geo locations as dumb tuple types

### LINKED ISSUE #995: GeoPolygon Filter points parameter
Something I found pretty confusing about this is that the IEnumerable of points uses Tuple<double,double> for each point which is ambiguous as to whether it is lat/lon or lon/lat.  I think it will be less confusing if a type with explicit lat/lon members was used instead.

Mpdreamz:


gmarz:
LGTM

PR #1271, opened 2015-03-11
split deleteresponse for Delete() and DeleteByQuery() fix #1150

### LINKED ISSUE #1150: DeleteByQuery() should return DeleteByQueryResponse
See #1146

`DeleteResponse` is currently holding both the response for delete and delete by query requests (for bwc reasons).  Instead, we should have a `DeleteByQueryResponse` and move the delete by query specific properties to it.

Mpdreamz:


gmarz:
LGTM :+1:

PR #1283, opened 2015-03-19
Convenience extensions for template exists

### LINKED ISSUE #1280: Add IElasticClient.TemplateExists string overload
I noticed today that I was getting an error when I tried to [delete a template](https://github.com/exceptionless/Exceptionless/blob/master/Source/Core/Repositories/ElasticSearchConfiguration.cs#L43-L44)

So I went and tried wrapping it in an if block and call TemplateExists(string). However, there isn't an overload that just takes a string. GetTemplate, PutTemplate and DeleteTemplate have a string overload.

gmarz:
Closes #1280


Mpdreamz:
LGTM :+1:

PR #1301, opened 2015-04-03
Initial proposal Fixes #1300

### LINKED ISSUE #1300: Include support to parse ElasticSearch time units
This is a pretty small start but the goal is to parse http://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#time-units into a TimeSpan.  

I started down this path when I realized that if I set Timeout("20m") on a request (/_cluster/health for example) that I also have to set the RequestConfiguration.RequestTimeout(<millis>).  Longer term, might be nice to do that be default, but not sure the right place to put that.  

Not sure the right place for this but for my tests to work better, it ended up in Elasticsearch.Net instead of NEST.

Also not quite sure what to do about things bigger than a week.  They start to become dependent on DateTime.Now (months aren't always 31 days, years aren't always 365 days).  Even days are a bit incorrect wrt timezones.  But it's a start...

aochsner:


Mpdreamz:
Great PR @aochsner, love the tests :+1:

PR #1305, opened 2015-04-10
ConditionlessWhen on function_score query

Mpdreamz:
Allows you to specify e.g: `.ConditionlessWhen(string.IsNullOrEmpty(query))` on a function score query to prevent functions to be applied in case of an empty query or fixed references inside the functions


gmarz:
LGTM :+1:

PR #1324, opened 2015-04-23
Feature/corefx

Mpdreamz:
This PR supersedes https://github.com/elastic/elasticsearch-net/pull/1043

# WORK IN PROGRESS

## TODO:

- [ ] Totally untested
- [ ] HttpClient has not been opensourced no clue how well it behaves now, e.g does it buffer reads and writes? No control over nagle algorithm etcereta.
- [ ] HttpClient lack of **real** synchronous interfaces bothers me and the current **faux pas** synchronous implementation will 99% certain have synchronization context bugs. 
- [ ] `.NET 4` support I can add it as a framework in `project.json` and add support for `async/await` with `Microsoft.Bcl.Async` but when I do I get this compiler error `type 'system.runtime.compilerservices.inotifycompletion' is not defined or imported elasticsearch.net httpconnection.cs`. We could `#ifdef` the old webrequest and Begin/End based connection but that would introduce a testing complexity I rather not introduce.


tugberkugurlu:
They have a new package for `HttpClient` for coreclr: System.Net.Http.WinHttpHandler Not sure what the actual distinction is but it seems that ASP.NET team is using it for core: https://github.com/aspnet/Security/blob/30d350da26d5c40e9e160ccb2cd120303da5baf1/src/Microsoft.AspNet.Authentication.OAuthBearer/project.json#L18

[OAuthBearerAuthenticationMiddleware](https://github.com/aspnet/Security/blob/30d350da26d5c40e9e160ccb2cd120303da5baf1/src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs) has some usage examples on HttpClient. IIRC, the APIs should be the same.

Mpdreamz:
Thanks for the links and heads up @tugberkugurlu :smile: 

This scares me a bit: 

https://github.com/aspnet/Security/blob/30d350da26d5c40e9e160ccb2cd120303da5baf1/src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs#L95-L110

I hope this will resolve itself when `HttpClient` and its default `HttpHandler` make it into the corefx repos.

Mpdreamz:
This was sent against `develop` going to reopen it but sent against `2.0`

PR #1325, opened 2015-04-24
Included index property for Indices Filter

abibell:
index property was missing for Indices filter. It is now added.
Refactored the test case


Mpdreamz:
LGTM :+1: please have a look at https://visualstudiogallery.msdn.microsoft.com/c8bccfe2-650c-4b42-bc5c-845e21f96328 

It allows you to have per project editor settings and will automatically set the correct indendation and line endings for our project while leaving your defaults alone. 

@editorconfig is awesome in many ways and supported in many editors :)

PR #1336, opened 2015-04-29
fix id properties configured on connection settings

Mpdreamz:
not utilizing idlookup delegate cache and doing reflection on each lookup :cry:


gmarz:
LGTM :+1:

PR #1347, opened 2015-04-30
Fix bad nuspec output path for .NET 4.0 nuget packages

### LINKED ISSUE #1346: NEST and Elasticsearch.NET cannot be installed in .NET 4.0 projects since 1.4.0
Hi, 

I’m currently working on a project that use NEST whit the .NET framework 4.0 and I noticed that I’m not able to update the nuget package to the latest version which is 1.4.3. The package manager do not show new versions nor available updates. The package is stuck at 1.3.1.

I checked the source code and it seems that the nuspec files are not configured correctly. Since 1.4.0 the Dlls are splitted into .NET 4.0 and .NET 4.5 and the Dlls for 4.0 are misplaced directly in the lib folder but they should be in lib/net40 folder like the .NET 4.5 are in the lib/net45 folder

cstlaurent:
resolve #1346 NEST and Elasticsearch.NET bad nuspec files that makes the package not usable in .NET 4.0

### DIFF
--- a/build/Elasticsearch.Net.Connection.Thrift.nuspec
+++ b/build/Elasticsearch.Net.Connection.Thrift.nuspec
@@ -18,6 +18,10 @@
 		<tags>elasticsearch elastic search lucene thrift nest</tags>
 	</metadata>
 	<files>
+    <file src="output\Elasticsearch.Net.Connection.Thrift\net40\Elasticsearch.Net.Connection.Thrift.dll" target="lib"/>
+    <file src="output\Elasticsearch.Net.Connection.Thrift\net40\Elasticsearch.Net.Connection.Thrift.pdb" target="lib"/>
+    <file src="output\Elasticsearch.Net.Connection.Thrift\net40\Elasticsearch.Net.Connection.Thrift.XML" target="lib"/>
+    
 		<file src="output\Elasticsearch.Net.Connection.Thrift\net40\Elasticsearch.Net.Connection.Thrift.dll" target="lib\net40"/>
 		<file src="output\Elasticsearch.Net.Connection.Thrift\net40\Elasticsearch.Net.Connection.Thrift.pdb" target="lib\net40"/>
 		<file src="output\Elasticsearch.Net.Connection.Thrift\net40\Elasticsearch.Net.Connection.Thrift.XML" target="lib\net40"/>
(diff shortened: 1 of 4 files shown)


gmarz:
Yikes!  Good catch @cstlaurent and thank you for the PR.  Would you mind signing the CLA?

LGTM, but I'd like to get @Mpdreamz 's eyes on this before merging.

cstlaurent:
I just sign it at 2:12 PM PDT, does the status refresh frequently?

Mpdreamz:
On my phone so excuse formatting from the nuget docs:

Mpdreamz:
Sorry for closing on my phone thumb slipped :)

This seems strange to me as I remember testing this on 4.0 projects and the docs mention falling back on contents of `lib`  

https://docs.nuget.org/Create/Enforced-Package-Conventions


-----------



The rule about copying assemblies from only one folder also applies to the root lib folder. Suppose a package has the following folder structure:
\lib
    \MyAssembly.dll (v1.0)
    \MyAssembly.Core.dll (v1.0)
    \Net40
        \MyAssembly.dll (v2.0)
In projects that target the .NET Framework 2.0 and the .NET Framework 3.5, NuGet copies both MyAssembly.dll and MyAssembly.Core.dll. But as was true of the previous example, in projects that target the .NET Framework 4, only MyAssembly.dll *from the *Net40 folder will be copied.

--------------

Would like to be on a PC to test this further but its really late here.

gmarz:
Hmm...I can confirm that 1.3.1 is indeed the latest package that shows up on a 4.0 project.

cstlaurent:
@Mpdreamz I agree, the docs is unclear about .NET 4.0 and 4.5 fallback. I checked folder structure of multiple nuget packages including EntityFramework and Reactive Extensions and both of them have the `lib/net40/` + `lib/net45/` structure, no file directly under `lib/`

Mpdreamz:
Cool then let's get this in ASAP

Will raise an issue with nuget separately

To be on the safe side let's include both root an net40 folder dlls though.

cstlaurent:
@Mpdreamz I'll modify the PR with files at both places

gmarz:
Merged.  Thanks again @cstlaurent

cstlaurent:
@gmarz Thanks to you for your prompt response :smiley:

Issue #1362, opened 2015-05-01
Support for supplying index settings on restore


gmarz:
https://github.com/elastic/elasticsearch/pull/9285

robertlyson:
I'll try to handle it.

gmarz:
Awesome!  Thanks @robertlyson, much appreciated.

PR #1372, opened 2015-05-04
Support for supplying index settings on restore #1362

robertlyson:
PR for https://github.com/elastic/elasticsearch-net/issues/1362 issue.

For `IndexSettings` property on `IRrestoreDescriptor` I reused `IUpdateSettingsRequest`, but I'm not sure if this name fits in this context. Way how we can combine `RestoreDescriptor` with `UpdateSettingsDescriptor` is really nice.

Moreover, I've extracted all setting names to separate class. Avoiding googling for each plain index setting is really convenient.

Any thoughts? Let me know.
Thanks.

### DIFF
--- a/src/Nest/DSL/UpdateSettingsDescriptor.cs
+++ b/src/Nest/DSL/UpdateSettingsDescriptor.cs
@@ -1,116 +1,154 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
 using Elasticsearch.Net;
 using Newtonsoft.Json;
 
 namespace Nest
-{
+{ 
+	public class SettingNames
+	{
+		public const string NumberOfReplicas = "index.number_of_replicas";
+		public const string AutoExpandReplicas = "index.auto_expand_replicas";
+		public const string BlocksReadOnly = "index.blocks.read_only";
+		public const string BlocksRead = "index.blocks.read";
+		public const string BlocksWrite = "index.blocks.write";
+		public const string BlocksMetadata = "index.blocks.metadata";
+		public const string RefreshInterval = "index.refresh_interval";
+		public const string IndexConcurrency = "index.index_concurrency";
+		public const string Codec = "index.codec";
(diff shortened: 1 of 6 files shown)


src/Nest/DSL/UpdateSettingsDescriptor.cs
```diff
@@ -1,116 +1,154 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
 using Elasticsearch.Net;
 using Newtonsoft.Json;
 
 namespace Nest
-{
+{ 
+	public class SettingNames
```
Mpdreamz: Love moving this out of the attributes :+1: suggestion on different name `UpdatableSettings`

Mpdreamz:
PR looks awesome as usual @robertlyson :+1: left a small comment.

src/Nest/DSL/UpdateSettingsDescriptor.cs
robertlyson: @Mpdreamz much more better name :+1: . Maybe we should split this class into three? UpdatableSettings, UnmodifiableSettings and UnremovableSettings? https://github.com/elastic/elasticsearch/blob/master/src/main/java/org/elasticsearch/snapshots/RestoreService.java#L92-105

Mpdreamz: I don't see any settings in `SettingsNames` that are not Updatable though, maybe I need another coffee :smile: 

Where would we use `Unmodifiable` and `Unremovable` ?

robertlyson: Yeah, that's true, looks like I need second coffee :)
I'll change name to `UpdatableSettings` later on.

robertlyson:
Thanks for review. I've pushed change.

Mpdreamz:
ty @robertlyson !

Issue #998, opened 2014-10-16
Variable as suffix


ThomasVer:
When using a variable as a suffix, it doesn't return the JSON you would expect.

For example, this code snippit

```
var dynamicFields = new List<string> { "private-email", "birthday", "fax" };

...
.Query(searchQuery)
.Type(TextQueryType.CrossFields)
.OnFieldsWithBoost(fb =>
	{
		fb
			.Add(f => f.FirstName, 3.0)
			.Add(f => f.FirstName.Suffix("edge_ngram"), 2.0)
			.Add(f => f.LastName, 3.0)
			.Add(f => f.LastName.Suffix("edge_ngram"), 2.0)
			.Add(f => f.EmailAddress, 3.0)
			.Add(f => f.EmailAddress.Suffix("edge_ngram"), 2.0)
			.Add(f => f.Country.Name, 3.0)
			.Add(f => f.Country.Name.Suffix("edge_ngram"), 2.0)
			.Add(f => f.Country.ISO2, 1.0);

		foreach (string suffix in dynamicFields)
		{
			fb.Add(f => f.DynamicFields.Suffix(suffix), 3.0);
		}
	}
)
...
```

results in the following JSON (notice the dynamicFields):

```
...
{
  "type": "cross_fields",
  "query": "demo",
  "fields": [
	"firstName^3",
	"firstName.edge_ngram^2",
	"lastName^3",
	"lastName.edge_ngram^2",
	"emailAddress^3",
	"emailAddress.edge_ngram^2",
	"country.name^3",
	"country.name.edge_ngram^2",
	"country.iSO2^1",
	"suffix.dynamicFields^3",
	"suffix.dynamicFields^3",
	"suffix.dynamicFields^3"
  ]
}
...
```

PR #1380, opened 2015-05-06
Feature/variable suffix

Mpdreamz:
Adds support var using variables in `Suffix()` expressions as requested in #998 

e.g:

```
var prop = "propXY12";
P(p => p.Name.Suffix(prop)).Should().Be("name.propXY12");
```

This PR also fixes a bug where `IDictionary<TKey, TValue>` expressions would fail unexpectedly where `Dictionary<Tkey, TValue> worked fine.


gmarz:
Nice!  LGTM :+1:

PR #1387, opened 2015-05-07
Fix fielddata fields on search and inner hits

### LINKED ISSUE #1366: Support for fielddata_fields on _search and inner hits
http://www.elastic.co/guide/en/elasticsearch/reference/1.5/search-request-fielddata-fields.html

Marking this as separate TODO also on inner hits while I am working on that not to get sidetracked.

gmarz:
This was largely implemented already but with a few issues.  I had to modify the code gen to skip the fielddata_fields param in order to work around https://github.com/elastic/elasticsearch/issues/11025.

Also modified the implementation on inner hits to be more consistent with search (i.e. `IList<PropertyPathMarker>` vs `IEnumerable<string>` and added an overload that accepts property path expressions).

Closes #1366

### DIFF
--- a/src/Nest/DSL/Search/InnerHitsDescriptor.cs
+++ b/src/Nest/DSL/Search/InnerHitsDescriptor.cs
@@ -36,7 +36,7 @@ namespace Nest
 		bool? Version { get; set; }
 
 		[JsonProperty(PropertyName = "fielddata_fields")]
-		IEnumerable<string> FielddataFields { get; set; }
+		IList<PropertyPathMarker> FielddataFields { get; set; }
 
 		[JsonProperty(PropertyName = "script_fields")]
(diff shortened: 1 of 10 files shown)


src/Nest/DSL/Search/InnerHitsDescriptor.cs
```diff
@@ -79,7 +79,7 @@ public class InnerHitsDescriptor<T> : IInnerHits where T : class
 		bool? IInnerHits.Explain { get; set; }
 		ISourceFilter IInnerHits.Source { get; set; }
 		bool? IInnerHits.Version { get; set; }
-		IEnumerable<string> IInnerHits.FielddataFields { get; set; }
+		IList<PropertyPathMarker> IInnerHits.FielddataFields { get; set; }
```
Mpdreamz: ++ for catching this before releasing @gmarz !

PR #1400, opened 2015-05-11
Add 'missing' field to specify what value to use in case the field in…

danp60:
… field_value_factor does not exist in the document.

From  http://www.elastic.co/guide/en/elasticsearch/reference/1.x/query-dsl-function-score-query.html

Also added unit tests.


gmarz:
Awesome, thanks for this @danp60 !

Just one small issue- this is currently targeting `master`.  Would you mind opening a new PR and target the `develop` branch instead?

Mpdreamz:
Pulled this in manually: https://github.com/elastic/elasticsearch-net/commit/ab20ebbeaedf6635ac5505720bc868befe9dc2d0

Thanks as always @danp60 :smile:

PR #1401, opened 2015-05-12
Keep exceptions stack trace in NEST ElasticClient Async calls

### LINKED ISSUE #1399: Nest ElasticClient is losing exception information in Async calls
At [line 145 of `ElasticClient.cs`](https://github.com/elastic/elasticsearch-net/blob/develop/src/Nest/ElasticClient.cs#L145) the current code looks like :

```
if (mr != null)
	throw mr;

var ae = r.Exception.Flatten();
if (ae.InnerException != null)
	throw ae.InnerException;
throw ae;
```

Re-throwing exceptions without wrapping them with `throw ex;` mean that their callstack is completely erased ([more details](http://community.bartdesmet.net/blogs/bart/archive/2006/03/12/3815.aspx)), the code should wrap the exceptions in a new exception and throw that instead.

Note: *I remarked the issue as it happens on a production system and made our investigations of the problem harder than necessary*

vbfox:
This pull request aim to fix #1399 by using the internal `InternalPreserveStackTrace` method of `Exception`.

There is a clean non-reflection way to do in in .Net 4.5 using `ExceptionDispatchInfo` (See [this blog](http://blogs.microsoft.co.il/sasha/2011/10/19/capture-transfer-and-rethrow-exceptions-with-exceptiondispatchinfo-net-45/)) but it would require changing the target version.

The unit test wouldn't catch all causes of exception callstack being erased, only a direct regression.

### DIFF
--- a/src/Elasticsearch.Net/Connection/RequestHandlers/RequestHandlerAsync.cs
+++ b/src/Elasticsearch.Net/Connection/RequestHandlers/RequestHandlerAsync.cs
@@ -204,7 +204,7 @@ namespace Elasticsearch.Net.Connection.RequestHandlers
 
 			//If we are not using any pooling and we see an exception we rethrow
 			if (!requestState.UsingPooling && t.IsFaulted && t.Exception != null && maxRetries == 0)
-				throw t.Exception;
+				t.Exception.RethrowKeepingStackTrace();
 
 
 			var retried = requestState.Retried;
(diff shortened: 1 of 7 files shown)


vbfox:
CLA should have been signed by our CTO and linked to my account, is there still something missing on that front ? (Bran jolly for company "ITG SSF")

Mpdreamz:
Company CLA's are a bit trickier for us to locate under which umbrella you fall but finally found you :)

Mpdreamz:
Im skipping the constants approach until NEST 2.0 when we move to the DNX project system, defining constants becomes that much more easier without worrying about clearing any undefined once using `/p:DefineConstants`

Mpdreamz:
Thanks again for this PR @vbfox !

I added RethrowKeepingStackTrace to all "throw ex" occurrences.

vbfox:
Thank you for looking at it !

PR #1402, opened 2015-05-12
Fix for default index on suggests

### LINKED ISSUE #1310: Suggest<T> doesn't infer index name from T
Currently, `Suggest<T>` doesn't infer the index name from `T`.  You are forced to be explicit:

```csharp
 var result = client.Suggest<ElasticsearchProject>(s => s
        .Index<ElasticsearchProject>()
        // snip ...
);
```

gmoskovicz:
Fixes #1310


gmarz:
Left a few nitpick comments, but otherwise LGTM!  Great job @gmoskovicz !

PR #1405, opened 2015-05-14
Update IdResolver.cs

dobariya:
If idProperty NULL then try to get id from original object itself

### DIFF
--- a/src/Nest/Resolvers/IdResolver.cs
+++ b/src/Nest/Resolvers/IdResolver.cs
@@ -70,7 +70,11 @@ namespace Nest.Resolvers
 			var idProperty = GetInferredId(type);
 			if (idProperty == null)
 			{
-				return null;
+				idProperty = @object.GetType().GetProperty("ID", BindingFlags.Public | BindingFlags.Instance | BindingFlags.IgnoreCase);
+				 if (idProperty == null)
+				 {
+				 	 return null;
+				 }
 			}
 			try
 			{


gmarz:
Hey @dobariya isn't this what [GetInferredId()](https://github.com/elastic/elasticsearch-net/blob/develop/src/Nest/Resolvers/IdResolver.cs#L99) already does?  Is it not working?

dobariya:
Hey @gmarz  GetInferredId() is not working when class,properties are created in memory at run time.Please check this URL http://mironabramson.com/blog/post/2008/06/Create-you-own-new-Type-and-use-it-on-run-time-(C).aspx  which creates class,properties in memory at run time.
Using such object properties at runtime, GetInferredId() returns ID as  null

Mpdreamz:
Hey @dobariya sorry for taking some time to get back to you. 

Before we can pull this in do you mind signing our [CLA](http://www.elasticsearch.org/contributor-agreement/) and updating this PR with a unit test?

dobariya:
@Mpdreamz .I have signed the CLA.Please let me know what is the next step to follow

Mpdreamz:
If you would be able to push an additional commit with a unit test for these runtime classes here:

https://github.com/elastic/elasticsearch-net/blob/develop/src/Tests/Nest.Tests.Unit/Internals/Inferno/IdLookupTests.cs

That would be really really appreciated!

dobariya:
If possible you can get idea from below url
https://github.com/dobariya/elasticsearch-net/commit/188f19a96163b426f08d96dff88ef7de6722f3bc

here i have made path update

PR #1410, opened 2015-05-19
Added overload to DismaxQueryDescriptor's Queries method to take QueryContainer[]

xueye:
This allows it to act like other descriptors that get formatted into an array, such as Bool Query's Should, if you've already composed your QueryContainers elsewhere.


gmarz:
LGTM.  Thanks @xueye !

Issue #1415, opened 2015-05-21
MLT JsonSerializationException : Could not create an instance of type Nest.IMultiGetOperation. Type is an interface or abstract class and cannot be instantiated. Path 'query.mlt.docs[0]._index'


abibell:
# Scenario
Elastic's **More like this** is good to find related documents for a documentId. As a business user I cannot know which fields make results relevant. It could be fields A, B & C today. I may realise that X & Y should be added when I am in production due to customer issues. What I want is that the way to configure the fields. In developer terms, a developer should be able to change the MLT query in a configuration and the system should use the query as template. In order to do this I should be able to serialise/deserialise MLT queries.

# Working sense query
```javascript
POST _search
{
   "query": {
      "mlt": {
         "fields": [
            "name", "content"
         ],
         "docs": [
            {
               "_index": "my_index",
               "_type": "my_type",
               "_id": "SOME_ID"
            }
         ],
         "min_term_freq": 1,
         "max_query_terms": 12         
      }
   },
   "size": 2
}

{
   "query": {
      "more_like_this": {
         "fields": [
            "title",
            "description",
            "abstract",
            "subject",
            "source"
         ],
         "min_term_freq": 1,
         "max_query_terms": 12,
         
      }
   },
   "size": 2
}
```

# Test code that fails in MoreLikeThisQueryJson.cs
```cs
[Test]
public void TestMoreLikeThisQuery()
{
    Func<MoreLikeThisQueryDocumentsDescriptor<dynamic>, MoreLikeThisQueryDocumentsDescriptor<dynamic>> documentsSelector = d => d.Get("SOME_ID", op => op.Index("my_index").Type("my_type"));
    Action<MoreLikeThisQueryDescriptor<ElasticsearchProject>> selector = fz => fz
        .Name("named_query")
        .OnFields(f => f.Name, f => f.Content)
        .LikeText("elasticsearcc")
        .Documents(documentsSelector)//This causes the error
        .MinTermFrequency(1)
        .MaxQueryTerms(12);
    var s = new SearchDescriptor<ElasticsearchProject>()
        .From(0)
        .Size(10)
        .Query(q => q.MoreLikeThis(selector));

	var json = TestElasticClient.Serialize(s);
	const string expected = @"{ from: 0, size: 10, query : 
	{ mlt: { 
		_name: ""named_query"",
		fields : [""name"", ""content"" ],
		like_text : ""elasticsearcc"" ,
        ""docs"": [
                    {
[...]

PR #1416, opened 2015-05-22
IMultiGetOperation can now be deserialized

abibell:
MultiGetOperation can now be deserialized as MultiGetOperationDescriptor<object> is now default
JsonConverter.
Resolved https://github.com/elastic/elasticsearch-net/issues/1415
Added unit tests to verify the outcome.

### DIFF
--- a/src/Nest/DSL/MultiGet/IMultiGetOperation.cs
+++ b/src/Nest/DSL/MultiGet/IMultiGetOperation.cs
@@ -1,10 +1,12 @@
 ﻿using System;
 using System.Collections.Generic;
+using Nest.Resolvers.Converters;
 using Newtonsoft.Json;
 
 namespace Nest
 {
 	[JsonObject(MemberSerialization = MemberSerialization.OptIn)]
+    [JsonConverter(typeof(ReadAsTypeConverter<MultiGetOperationDescriptor<object>>))]
 	public interface IMultiGetOperation
 	{
 		[JsonProperty(PropertyName = "_index")]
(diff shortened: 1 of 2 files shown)


gmarz:
Thank you @abibell.  LGTM :+1: 

There are definitely more instances where round trip serialization fails within the client.  Part of our plan for NEST 2.0 is to go over all of our mappings/types, as well as re-working our [unit/integration tests](https://github.com/elastic/elasticsearch-net/tree/2.0/src/Tests/Nest.Tests.Literate) which will hopefully catch all of these instances.

Side note: would you mind installing [EditorConfig](https://visualstudiogallery.msdn.microsoft.com/c8bccfe2-650c-4b42-bc5c-845e21f96328) for future PRs?  This will fix the formatting/indentation differences without affecting the preferences on other projects.

abibell:
@gmarz I shall do the editor config next time. Yes I came across and reported 2 so far. Are we better off managing the whole query syntax using xsd or lexical grammar Which will generate perfect working code always? For version 2.0, I mean. Hman errors are due to use handcrafting these things.

gmarz:
@abibell Yes we are!  That's something we've been brain storming on for a while now but it comes with a handful of challenges.  We first toyed with the idea of using JSON schema, but it had too many limitations and could not accurately represent the Elasticsearch query DSL.  For instance, you could use pattern properties to represent variable property names, but you aren't able to do things like enforce the existence of a single occurence, etc...

Our idea at the moment is to create a reference implementation using something like Typescript which we could generate our class and interfaces from, and document any parts that require custom serialization.  This would need to be maintained just like the REST spec is today.

Unfortunately though, with all the things already on our plate for 2.0, this will likely be more of a 3.0 thing...

If you have any thoughts/ideas on this, we would love to hear them.

abibell:
I am keen to do that. I will create an Issue so we can track the progress and list all the limitations, so we can address them.

Issue #1144, opened 2014-12-12
Create a Nest.Contrib library for supporting third-party plugins


gmarz:
https://github.com/elasticsearch/elasticsearch-net/issues/1028
https://github.com/elasticsearch/elasticsearch-net/pull/538

niemyjski:
I'm against having a Nest.Contrib library. I think each plugin should have a library so I don't accidentally use a plugin I don't have installed (discovery). Also, it can be aligned with the specific plugin version you are using.

gmarz:
Fair points.  Although, I'd be more concerned with the versioning issue you bring up rather than the accidental usage a plugin.  I think proper namespacing would do a pretty good job in preventing that.

We are definitely not sold on the idea of a Nest.Contrib either, just toying with the idea and haven't really given it much thought yet.

I think the idea was that it could provide a clear and concise way to extend the client, and we would be able to provide the minimum boiler plate code required in order to "hook" into the client properly.  It also gives us some control over the plugins- enforce conventions (like namespacing mentioned above), run unit tests with our builds, etc...

Mpdreamz:
closing this because general concensus is a single contrib library is not the way to go

PR #1417, opened 2015-05-22
Elasticsearch carrot plugin implementation.

akuzhan:
This is plugin implements on the fly data clustering for elastic search results. 

I implemented carrot2 plugin into eleasticsearch.net project for our internal project and wanted to share it with people. 

This plugin is useful if you like to have a RIPL or Foamtree chart. 

https://github.com/carrot2/elasticsearch-carrot2

### DIFF
--- a/src/Elasticsearch.Net/Connection/ITransport.cs
+++ b/src/Elasticsearch.Net/Connection/ITransport.cs
@@ -16,7 +16,9 @@ namespace Elasticsearch.Net.Connection
 			string method, 
 			string path, 
 			object data = null, 
-			IRequestParameters requestParameters = null);
+			IRequestParameters requestParameters = null,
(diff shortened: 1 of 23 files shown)


gmarz:
@akuzhan Thanks for sharing!  This is really interesting.

We've thought about ways to support community plugins (see #1144) and have made a conscience decision not to support them by default within NEST.  So unfortunately we can't pull this in.

Instead, we're leaning towards the approach of having a separate library/package for third party/community plugins- but also understand that it's not always possible to extend the client from outside without making changes to some of the internal bits.  This is something that we're [actually working on improving at the moment](https://github.com/elastic/elasticsearch-net/tree/feature/extensibility), and this PR will help us better understand what else needs to be done, so thank you!

Stay tuned.

PR #1421, opened 2015-05-25
Fixed Suffix for Completion on Mapping

### LINKED ISSUE #1408: Multiple completion suggesters
Hi,

I am trying to create some completion suggesters on some of my fields. My document class looks like this:
```
[ElasticType(Name = "rawfiles", IdProperty = "guid")]
public class RAW
{
	[ElasticProperty(OmitNorms = true, Index = FieldIndexOption.NotAnalyzed, Type = FieldType.String, Store = true)]
    public string guid { get; set; }
			
	[ElasticProperty(OmitNorms = true, Index = FieldIndexOption.Analyzed, Type = FieldType.String, Store = true, IndexAnalyzer = "def_analyzer", SearchAnalyzer = "def_analyzer_search", AddSortField = true)]
	public string filename { get; set; }

	[ElasticProperty(OmitNorms = true, Index = FieldIndexOption.Analyzed, Type = FieldType.String, Store = true, IndexAnalyzer = "def_analyzer", SearchAnalyzer = "def_analyzer_search")]
    public List<string> tags { get { return new List<string>(); } }
}
```

And here is how I am trying to create the completion fields

```
public bool CreateMapping(ElasticClient client, string indexName)
{
    IIndicesResponse result = null;
    try
    {
        result = client.Map<RAW>(
                c => c.Index(indexName)
                    .MapFromAttributes()
                    .AllField(f => f.Enabled(false))
                    .SourceField(s => s.Enabled())
                    .Properties(p => p
                        .Completion(s => s.Name(n => n.tags.Suffix("comp"))
                                    .IndexAnalyzer("standard")
                                    .SearchAnalyzer("standard")
                                    .MaxInputLength(20)
                                    .Payloads()
                                    .PreservePositionIncrements()
                                    .PreserveSeparators())
                        .Completion(s2 => s2.Name(n=>n.filename.Suffix("comp"))
                                    .IndexAnalyzer("standard")
                                    .SearchAnalyzer("standard")
                                    .MaxInputLength(20)
                                    .Payloads()
                                    .PreservePositionIncrements()
                                    .PreserveSeparators())
                        )
                   );
    }
    catch (Exception)
    {

    }

    return result != null && result.Acknowledged;
}
```

My problem is that this is only creating a single completion field named "comp". I was under the impression that this will create two completion fields, one named filename.comp and the other named tags.comp.

I then tried the answer on this [SO question](http://stackoverflow.com/questions/21433586/nest-suggestcompletion-usage-throws-is-not-a-completion-suggest-field-excepti) but this complicated the matter even worse as now my two fields were mapped as a completion field only.

Just to be clear, I want to create a multi-field (field) that has a data, sort and completion fileds. Much like the one in [this example](https://developer.rackspace.com/blog/qbox/)

gmoskovicz:
__This pull request fixes completion mapping with suffix__

:fireworks:  Fixes #1408 :fireworks:

Please review

- [ ] @Mpdreamz / @gmarz

### DIFF
--- a/src/Nest/DSL/PutMappingDescriptor.cs
+++ b/src/Nest/DSL/PutMappingDescriptor.cs
@@ -12,7 +12,7 @@ namespace Nest
 		RootObjectMapping Mapping { get; set; }
 	}
 
-	public interface IPutMappingRequest<T> : IPutMappingRequest where T : class {}
+	public interface IPutMappingRequest<T> : IPutMappingRequest where T : class { }
 
 	internal static class PutMappingPathInfo
 	{
@@ -25,7 +25,7 @@ namespace Nest
 	public partial class PutMappingRequest : IndicesTypePathBase<PutMappingRequestParameters>, IPutMappingRequest
 	{
 		public RootObjectMapping Mapping { get; set; }
-		
+
 		/// <summary>
 		/// Calls putmapping on /_all/{type}
 		/// </summary>
@@ -50,7 +50,7 @@ namespace Nest
 		public PutMappingRequest(IndexNameMarker index, TypeNameMarker type)
 		{
 			this.Type = type;
-			this.Indices = new [] { index };
+			this.Indices = new[] { index };
 		}
 		protected override void UpdatePathInfo(IConnectionSettingsValues settings, ElasticsearchPathInfo<PutMappingRequestParameters> pathInfo)
 		{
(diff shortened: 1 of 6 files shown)


gmoskovicz:
:disappointed: We are still getting this `CLA` error, lets try to fix this early this week :smile:

Mpdreamz:
closing this in favor of the continued discussion on #1408

PR #1423, opened 2015-05-27
Feature/extensibility enhancements

gmarz:
* `ElasticClient` now implements `IHighLevelToLowLevelDispatcher` explicitly so the internal-ish dispatch methods are less exposed.  This required casting to `IHighLevelToLowLevelDispatcher` in order to access the dispatch methods, so I added an internal `Dispatcher { get { return this; } }` convenience property (much like we do with `Self` in the descriptors).

* Renamed `IDomainObject` to the more descriptive`INestSerializable`.

* Changed `CloneFrom()` back to internal since it doesn't need to be exposed.

### DIFF
--- a/src/Elasticsearch.Net/Domain/Response/ElasticsearchResponse.cs
+++ b/src/Elasticsearch.Net/Domain/Response/ElasticsearchResponse.cs
@@ -19,6 +19,7 @@ using Elasticsearch.Net.Serialization;
 
 namespace Elasticsearch.Net
 {
+	// TODO: Make this class internal in 2.0
 	public static class ElasticsearchResponse
 	{
 		internal static ElasticsearchResponse<TTo> CloneFrom<TTo>(IElasticsearchResponse from, TTo to)
--- a/src/Elasticsearch.Net/Domain/Response/ElasticsearchResponse.cs
+++ b/src/Elasticsearch.Net/Domain/Response/ElasticsearchResponse.cs
@@ -19,11 +19,9 @@ using Elasticsearch.Net.Serialization;
 
 namespace Elasticsearch.Net
 {
-	//TODO document and possibly rename some of the properties
-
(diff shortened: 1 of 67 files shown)


src/Nest/Domain/ICustomJson.cs
```diff
@@ -2,14 +2,12 @@
...
 	{
 		object GetCustomJson();
 	}
 
-	public interface IDomainObject
-	{
-		
-	}
+	// Any object that implements this will automatically be added to the serialization contract
+	public interface INestSerializable { }
```
Mpdreamz: lets make this a proper summary: `Any object that implements this interface will automatically have all the JsonProperties of all its implementing interfaces discovered.`

src/Nest/ElasticClient.cs
```diff
@@ -172,17 +157,5 @@ private TRequest ForceConfiguration<TRequest>(TRequest request, Action<IRequestC
 			request.RequestConfiguration = configuration;
 			return request;
 		}
-
-		public static void Warmup()
```
Mpdreamz: This has always been public or not? if not we need to add it back in (sadly)

src/Elasticsearch.Net/Domain/Response/ElasticsearchResponse.cs
```diff
@@ -19,11 +19,9 @@
 
 namespace Elasticsearch.Net
 {
-	//TODO document and possibly rename some of the properties
-
 	public static class ElasticsearchResponse
```
Mpdreamz: Add a todo to make this whole class `internal` in 2.0

src/Nest/ElasticClient.cs
gmarz: It has been public, but I didn't remove it.  Just moved it up.

PR #1434, opened 2015-06-04
Add size parameter to reindex descriptor

shanekenney:
The current scroll search size for reindexing is hard coded. Allow the size to be changed using the ReindexDescriptor.


gmarz:
Nice catch, thanks @shanekenney !  LGTM :+1:

Issue #1433, opened 2015-06-04
Reindex doesn't support multiple Types


dusda:
So ElasticClient.ReIndex<T> is very handy for migrating indexes from one Type mapping to another, but what if you have more than one Type in your Index? ReIndex<T> doesn't let you use it on an existing Index and you can't specify more than one Type when using it, so I'm left with either writing my own implementation to reindex the thing, or re-work my model to move each Type mapping into it's own index.

shanekenney:
@dusda NEST can reindex all types within an index. You should be able to use the following as an example (from ReindexTests):

```
Client.Reindex<object>(r => r
    .FromIndex("sourceindex")
    .ToIndex("targetindex")
```

dusda:
That...works beautifully. Thank you.

What an odd syntax, though, wouldn't it make more sense to have method signature with no generic type, at least externally?

gmarz:
Actually, it looks like we've uncovered a bug here.

Regardless of the type specified, whether some `T` or just `object`, reindex will still [search across all types](https://github.com/elastic/elasticsearch-net/blob/develop/src/Nest/Domain/Observers/ReindexObservable.cs#L58).

`Reindex()<T>` should respect the type when executing the search portion, and we should have a non-generic `Reindex()` as @dusda is suggesting, that defaults to all types, but allows for specifying multiple.

Mpdreamz:
@gmarz @russcam vote to bump to 2.0.0 beta1? :+1: from me

russcam:
@Mpdreamz Agreed :+1:

russcam:
Closed via #1782

Issue #1404, opened 2015-05-13
NEST Count API throws exceptions when default index is not defined


krisze:
Version 1.0.0 beta, 1.4.2
When a default index is not defined upon client creation, even if the index name is defined in the request for Count API, it will throws a `No default index defined` exception.
When I tried to upgrade to 1.4.2, it will throw the an error saying 
`Elastic Nest violates the constraint of type parameter "P"`.

gmarz:
Hey @krisze I'm having trouble reproducing this.  Could you provide an example?

krisze:
Hi @gmarz, I'm tried on both 1.0.0 and 1.4.2:

If no default index was given:
```c#
    var settings = new ConnectionSettings(uri);
     container.Register(
                Component.For<IElasticClient>()
                    .ImplementedBy<ElasticClient>()
                    .DependsOn(new { settings = settings })
                    .LifestyleScoped()
    );
```
then try getting count by running the following query:
````c#

    var response = client.Count<objectName>(count => count
                .Index(indexName)
                .Type(typeName)
                .Query(q => q
                    .Match(m => m
                        .OnField(f => f.id)
                        .Query(id))));
```
for 1.0.0 (beta) the following error occured:
```xml
  
    <Error>
    <Message>An error has occurred.</Message>
        <ExceptionMessage>No default index set on connection!</ExceptionMessage>
        <ExceptionType>System.NullReferenceException</ExceptionType>
        <StackTrace>
    at System.Web.Http.ApiController.<InvokeActionWithExceptionFilters>d__1.MoveNext() --- End of     stack trace from previous location where exception was thrown --- at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task) at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task) at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult() at System.Web.Http.Dispatcher.HttpControllerDispatcher.<SendAsync>d__0.MoveNext()
        </StackTrace>
    </Error>`
```
for 1.4.2 the following error occurred, and if you debug into it, you will see the no index error in the client object at runtime as well:

```xml

    <Error>
     <Message>An error has occurred.</Message>
     <ExceptionMessage>
[...]

PR #1440, opened 2015-06-08
Support reindex with to better syntax

gmoskovicz:
Support for #1433 

- [ ] @gmarz review/ @Mpdreamz

### DIFF
--- a/src/Nest/Domain/Observers/ReindexObservable.cs
+++ b/src/Nest/Domain/Observers/ReindexObservable.cs
@@ -38,6 +38,7 @@ namespace Nest
 			var toIndex = this._reindexDescriptor._ToIndexName;
 			var scroll = this._reindexDescriptor._Scroll ?? "2m";
 			var size = this._reindexDescriptor._Size ?? 100;
+			var allTypes = this._reindexDescriptor._allTypes;
 
 			fromIndex.ThrowIfNullOrEmpty("fromIndex");
 			toIndex.ThrowIfNullOrEmpty("toIndex");
@@ -54,14 +55,10 @@ namespace Nest
 			var page = 0;
 			Func<SearchDescriptor<T>, SearchDescriptor<T>> searchDescriptor = s => s.Index(fromIndex);
 
-			if (typeof(T).Name.Equals(typeof(object).Name))
-			{
+			if (allTypes)
 				searchDescriptor = s => searchDescriptor(s).AllTypes();
-			}
 			else
-			{
 				searchDescriptor = s => searchDescriptor(s).Type<T>();
-			}
 
 
 			var searchResult = this.CurrentClient.Search<T>(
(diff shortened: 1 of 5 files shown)


src/Nest/Domain/Observers/ReindexObservable.cs
```diff
@@ -1,115 +1,125 @@
...
+
+			var createIndexResponse = this.CurrentClient.CreateIndex(
+				toIndex, (c) => settings(c.InitializeUsing(indexSettings.IndexSettings)));
+			if (!createIndexResponse.IsValid)
+				throw new ReindexException(createIndexResponse.ConnectionStatus);
+
+			var page = 0;
+			Func<SearchDescriptor<T>, SearchDescriptor<T>> searchDescriptor = s => s.Index(fromIndex);
+
+			if (typeof(T).Name.Equals(typeof(object).Name))
```
Mpdreamz: I think this fix should be moved to the non generic overload, this might still cause confusion when someone actually calls `Reindex<objecT>()`

src/Nest/ElasticClient-Reindex.cs
```diff
@@ -14,5 +14,11 @@ public IObservable<IReindexResponse<T>> Reindex<T>(Func<ReindexDescriptor<T>, Re
 			var observable = new ReindexObservable<T>(this, _connectionSettings, reindexDescriptor);
 			return observable;
 		}
+
+		/// <inheritdoc />
+		public IObservable<IReindexResponse<object>> Reindex(Func<ReindexDescriptor<object>, ReindexDescriptor<object>> reindexSelector)
```
Mpdreamz: Maybe we can default the type of the non generic to return `IReindexResponse<IDocument>` the lazy deserialization of `IDocument` is slightly more useful then `object` in this case.

src/Nest/Domain/Observers/ReindexObservable.cs
gmoskovicz: @Mpdreamz i did not understand this one :). What should we do if it comes as generic object? I thought the idea was to search all types?

Mpdreamz: Here: https://github.com/elastic/elasticsearch-net/blob/1433-support-generic-reindex/src/Nest/ElasticClient-Reindex.cs#L19 we added a non generic overload to the client. However we can not distinguish here between someone calling the generic or non generic overload. If someone now calls `client.Reindex<object>()` it will reindex on everything where as only `client.Reindex()` should.

src/Nest/ElasticClient-Reindex.cs
```diff
@@ -14,5 +14,11 @@ public IObservable<IReindexResponse<T>> Reindex<T>(Func<ReindexDescriptor<T>, Re
 			var observable = new ReindexObservable<T>(this, _connectionSettings, reindexDescriptor);
 			return observable;
 		}
+
+		/// <inheritdoc />
+		public IObservable<IReindexResponse<IDocument>> Reindex(Func<ReindexDescriptor<object>, ReindexDescriptor<object>> reindexSelector)
```
Mpdreamz: To be truly lazy `IDocument` this signature needs to change to `Reindex(Func<ReindexDescriptor<IDocument>, ReindexDescriptor<IDocument>> reindexSelector)`

gmoskovicz:
@Mpdreamz is this what you expected? I dont know but something is telling me that i've missed something :)

Mpdreamz:
err meant #1404 there :)

Mpdreamz:
No this looks good to me! :+1:

PR #1442, opened 2015-06-09
Support for murmur3 mapping type

### LINKED ISSUE #1441: Add Support for murmur3 Field Type
The new, "experimental" Cardinality Aggregation talks about a murmur3 field type which allows precomputed hashes to speed up high cardinality fields:

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-cardinality-aggregation.html#_pre_computed_hashes

Can this be added to the FieldType enum? Is there some backdoor way to specify a field type by string?

gmarz:
Adding support for the mumur3 mapping type, mainly for use with the cardinality aggregation as explained here: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-cardinality-aggregation.html#_pre_computed_hashes

This feels a bit dirty though IMO, since we have to treat it as an `IElasticCoreType` in order for it to work as a multi field.

We need to rethink the current object graph for mapping types for 2.0, but that's a separate issue.

Closes #1441

### DIFF
--- a/src/Nest/Domain/Mapping/Descriptors/CorePropertiesDescriptor.cs
+++ b/src/Nest/Domain/Mapping/Descriptors/CorePropertiesDescriptor.cs
@@ -81,6 +81,16 @@ namespace Nest
 			return this;
 		}
 
+		public CorePropertiesDescriptor<T> Murmur3Hash(Func<MurmurHashMappingDescriptor<T>, MurmurHashMappingDescriptor<T>> selector)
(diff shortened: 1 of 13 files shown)


Mpdreamz:
For 2.0 i think we need to drop the distinction between `CoreType` and others. This is still a remnant of the the way elasticsearch 0.x did multifield mappings.

PR #1448, opened 2015-06-10
Addition of MultiSortExtensions to allow easy sorting across multiple fields

mgoodfellow:
It can be difficult to build multi-level sorting programmatically through the existing interfaces, so these extensions allow a quick way of adding a list of various `SortFieldDescriptor`s

Usage:

```csharp
// The actual search
var searchResponse =
    this.client.Search<DTO>(
        f => f.Filter(FITLER)
              .Size(SIZE)
              .Query(QUERY)
              .MultiSort(SortMethod(params)));

// This could dynamically build the required sorts based on the parameters passed in (e.g. from UI)
private static IEnumerable<SortFieldDescriptor<DTO>> SortMethod(SortParams params)
{
    IList<SortFieldDescriptor<DTO>> sorts = new List<SortFieldDescriptor<DTO>>();

    if (params.SomeSort.HasValue)
    {
        sorts.Add(new SortFieldDescriptor<DTO>().OnField(f => f.Field).Descending().MissingLast());
    }

    if (params.AnotherSort.HasValue)
    {
        sorts.Add(new SortFieldDescriptor<DTO>().OnField(f => f.Field2).Descending().MissingLast());
    }

    return sorts;
}
```


mgoodfellow:
I've signed the CLA, but I'm not sure how to get the PR to update accordingly

gmarz:
Hey @mgoodfellow thanks for this!  I'm liking it a lot, but I have just a few suggestions...

I don't think these need to be extension methods?  I can't think of a reason for not just including them as instance methods on `SearchDescriptor`.

I think I like `SortMulti` as opposed to `MultiSort`...better Intellisense discovery, etc.  Or maybe just overloading `Sort` would be better?  Not sold on that, just an idea...thoughts?

Lastly, would you mind adding some [unit tests](https://github.com/elastic/elasticsearch-net/blob/develop/src/Tests/Nest.Tests.Unit/Search/Sorting/SortTests.cs) for this?  

Thanks again, this is great!

mgoodfellow:
Hi @gmarz,

Thanks for the feedback! Originally I made them as extension methods as this was the easiest way to enhance the functionality while still using the standard Nuget packages.

I agree with the idea to make it `SortMulti` to enhance the intellisense, but I think overloading `Sort` might get confusing, so it's good to keep some degree of separation between their functionality.

I have added a couple of tests to check the DSL produced.

Any more comments are more than welcome!

gmarz:
Awesome, thank you @mgoodfellow.  LGTM :+1:

Issue #1462, opened 2015-06-23
An item with the same key has already been added (ElasticSearch.NET 1.6.0 vs Newtonsoft.Json 7.0.1)


SalvatoreMeschini:
I updated both Nest packages and ElasticSearch server to latest versions, but now the code is broken.

This line (I simply recompiled the project after updating the packages via NuGet, I'm using Newtonsoft.Json 7.0.1):

  ISearchResponse<ETKDFAQ> response = _client.Search<ETKDFAQ>(s => s.From(0).Size(limit).QueryString(searchString));

Now is leading to:

ERROR 2015-06-23 10:47:14,675 [11   ] .Mvc.Controllers.AbpHandleErrorAttribute - System.ArgumentException: An item with the same key has already been added.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Collections.ObjectModel.KeyedCollection`2.AddKey(TKey key, TItem item)
   at System.Collections.ObjectModel.KeyedCollection`2.InsertItem(Int32 index, TItem item)
   at System.Collections.ObjectModel.Collection`1.Add(T item)
   at Newtonsoft.Json.Utilities.CollectionUtils.AddRange[T](IList`1 initial, IEnumerable`1 collection)
   at Newtonsoft.Json.Serialization.DefaultContractResolver.CreateObjectContract(Type objectType)
   at Newtonsoft.Json.Serialization.DefaultContractResolver.CreateContract(Type objectType)
   at Nest.Resolvers.ElasticContractResolver.CreateContract(Type objectType) in c:\code\elasticsearch-net\src\Nest\Resolvers\ElasticContractResolver.cs:line 29
   at Newtonsoft.Json.Serialization.DefaultContractResolver.ResolveContract(Type type)
   at Nest.Resolvers.SettingsContractResolver.ResolveContract(Type type) in c:\code\elasticsearch-net\src\Nest\Resolvers\SettingsContractResolver.cs:line 34
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.GetContractSafe(Object value)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.Serialize(JsonWriter jsonWriter, Object value, Type objectType)
   at Newtonsoft.Json.JsonSerializer.SerializeInternal(JsonWriter jsonWriter, Object value, Type objectType)
[...]

PR #1465, opened 2015-06-24
Upgrade to Json.NET 7.0.1

gmarz:
This PR updates NEST to Json.NET 7.0.1 and addresses the issue described in #1462.

The problem lied in [ElasticContractResolver.CreateProperties](https://github.com/elastic/elasticsearch-net/blob/fix/update-jsonnet/src/Nest/Resolvers/ElasticContractResolver.cs#L91).  The returned collection of properties was non-unique, causing the [DefaultContractResolver](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/DefaultContractResolver.cs#L392) in Json.NET to blow up [here](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Utilities/CollectionUtils.cs#L65).

There were some recent changes to `DefaultContractResolver`, but I'm not exactly sure what changed specifically caused this behavior and if it's a bug or not- but regardless, `ElasticContractResolver.CreateProperties` should still return a distinct list of properties to be safe.


Mpdreamz:
https://github.com/elastic/elasticsearch-net/blob/fix/update-jsonnet/build/NEST.nuspec#L17

We should upgrade our nuspec dependency as well and introduce `allowedVersions="[7,8)" ` explicitly to not get bitten by this again. 

The fact that major version range is not implied is a nuget fail IMO though.

gmarz:
> We should upgrade our nuspec dependency as well and introduce allowedVersions="[7,8)" explicitly to not get bitten by this again.

Agreed, just added this.

PR #1480, opened 2015-07-07
Fix #1479: Full integral type coverage for field type serialization

### LINKED ISSUE #1479: Mapping of byte and short types in MapFromAttributes
We are using the MapFromAttributes method to create Elasticsearch mapping from classes. We have found out that it doesn't work with properties that have byte and short types.

I've checked the library sources and found out that there is really no support for these types. Here is the part of the GetFieldTypeFromType method from file TypeMappingWriter.cs:

```
switch (propertyType.Name)
{
	case "Int32":
		return FieldType.Integer;
	case "Int64":
		return FieldType.Long;
	case "Single":
		return FieldType.Float;
	case "Decimal":
	case "Double":
		return FieldType.Double;
	case "DateTime":
		return FieldType.Date;
	case "Boolean":
		return FieldType.Boolean;
}
```

Could you please fix the issue?

gmarz:
Closes #1479

### DIFF
--- a/src/Nest/Resolvers/Writers/TypeMappingWriter.cs
+++ b/src/Nest/Resolvers/Writers/TypeMappingWriter.cs
@@ -85,6 +85,7 @@ namespace Nest.Resolvers.Writers
 			using (var ms = new MemoryStream(nestedJson.Utf8Bytes()))
 				return this._elasticSerializer.Deserialize<RootObjectMapping>(ms);
 		}
+
 		internal ObjectMapping ObjectMappingFromAttributes()
 		{
 			var json = JObject.Parse(this.MapFromAttributes());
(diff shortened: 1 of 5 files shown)


src/Nest/Resolvers/Writers/TypeMappingWriter.cs
```diff
@@ -274,8 +274,14 @@ private string GetElasticSearchTypeFromFieldType(FieldType? fieldType)
 				switch (propertyType.Name)
 				{
 					case "Int32":
+					case "UInt32":
+					case "Int16":
+					case "UInt16":
+					case "Byte":
+					case "SByte":
```
Mpdreamz: I think we can do better then map to integer e.g CLR byte to Elasticsearch byte

gmarz: Agreed.  Just pushed another commit that adds support for `byte` and `short.  Also did some minor cleanup of `TypeMappingWriter`.

Issue #1477, opened 2015-07-06
Strongly typed payload for completion suggestion


tkirill:
Currently [SuggestOption.Payload](https://github.com/elastic/elasticsearch-net/blob/4fcecae93dcf16a4ff7e110135da81eed409dda5/src/Nest/Domain/Suggest/SuggestOption.cs#L19) is of type `object`.  We have to cast the value of this field to `JObject` and work with JSON.NET API if we want to get data from payload.

Recently in our project we created strongly-typed copy of `SuggestOption`:

    public class SuggestOption<T>
    {
        public SuggestOption(Nest.SuggestOption original)
        {
            // deserialize Payload
            // copy everything else
        }

        T Payload {get; set; }
        // other fields from original SuggestOptions
    }

Also we created extension method for `IElasticClient` which calls `.SuggestCompletion` and wraps its response in our strongly-typed class.

Is there any other way to do this?  If no do you like the idea to implement similar functionality in Nest?  I would be happy to do a PR!

gmarz:
@tkirill Totally agree.  @Mpdreamz and I have already discussed this and planned on fixing it for 2.0.  In fact, I could have sworn we opened an issue already but I guess not- we can use this one.

Rolling your own class as you've done is the only workaround unfortunately.  A PR for this would be awesome and very much appreciated!  Just make sure to target the `2.0` branch and not `develop`.

tkirill:
@gmarz What version of Visual Studio should I use to work in the branch 2.0?  VS2015?

gmarz:
@tkirill Yes, 2015

gmarz:
Closed via https://github.com/elastic/elasticsearch-net/commit/02dd5353834a80fb5d5c57d1dace6d6e3ad582f5

tkirill:
Cool!  Thanks for implementing this!

PR #1482, opened 2015-07-09
Build error in 2.0 due to missing file

tkirill:
File \src\Nest\Resolvers\Converters\UriJsonConverter.cs was in Nest.csproj but wasn't on disk.

I started #1477 from branch 2.0 and noticed this little error.


Mpdreamz:
Thanks! we jumped to `logical-grouping` as a side branch that grow into 120 commits that should have gone to 2.0. We'll be merging all of those into `2.0` today as well.

Issue #1481, opened 2015-07-09
Literate Tests that we need to write for 2.0


Mpdreamz:
In `2.0` we are redoing our test architecture so that we can we wrtite our unit/integration tests in one go  while they also serve as compilable documentation thanks to some roslyn magic.  We're dubbing this literate testing.

The new tests (and also much of the client) are layout to mimic the official elasticsearch documentation. 

This lists all the literate tests that we need to write for `2.0` based of the elasticsearch documentation. We have more that deal with client concepts (inferring, failover, connectionsettings) that are outside the scope of this issue.

- [ ] ├── aggregations
- [ ] │   ├── bucket
- [x] │   │   ├── children-aggregation.asciidoc
- [x] │   │   ├── datehistogram-aggregation.asciidoc
- [x] │   │   ├── daterange-aggregation.asciidoc
- [x] │   │   ├── filter-aggregation.asciidoc
- [x] │   │   ├── filters-aggregation.asciidoc
- [ ] │   │   ├── geodistance-aggregation.asciidoc
- [ ] │   │   ├── geohashgrid-aggregation.asciidoc
- [ ] │   │   ├── global-aggregation.asciidoc
- [ ] │   │   ├── histogram-aggregation.asciidoc
- [ ] │   │   ├── iprange-aggregation.asciidoc
- [ ] │   │   ├── missing-aggregation.asciidoc
- [ ] │   │   ├── nested-aggregation.asciidoc
- [ ] │   │   ├── range-aggregation.asciidoc
- [ ] │   │   ├── reverse-nested-aggregation.asciidoc
- [ ] │   │   ├── sampler-aggregation.asciidoc
- [ ] │   │   ├── significantterms-aggregation.asciidoc
- [ ] │   │   └── terms-aggregation.asciidoc
- [ ] │   ├── bucket.asciidoc
- [ ] │   ├── metrics
- [ ] │   │   ├── avg-aggregation.asciidoc
- [ ] │   │   ├── cardinality-aggregation.asciidoc
- [ ] │   │   ├── extendedstats-aggregation.asciidoc
- [ ] │   │   ├── geobounds-aggregation.asciidoc
- [ ] │   │   ├── max-aggregation.asciidoc
- [ ] │   │   ├── min-aggregation.asciidoc
- [ ] │   │   ├── percentile-aggregation.asciidoc
- [ ] │   │   ├── percentile-rank-aggregation.asciidoc
[...]

PR #1483, opened 2015-07-10
Strongly-typed payload for completion suggest.

tkirill:
Solves #1477 

## Work in progress

The work isn't finished. I need to write tests.

---

So I created separate class `CompletionOption<T>` as discussed in #1477.  To retrieve these objects the method `GetCompletions` was added to `Suggest`.  I like that the changes are minimal but I looked at other classes that stores response from Elasticsearch and saw that none of them has methods.  It makes me doubt that I chose right implementation.

Maybe it would be better to introduce method `IElasticClient.CompletionSuggest<TIndex, TPayload>` which will copy `Suggest` API except the return type?

Also, I couldn'f find a place where I should put the tests.  Project `Tests` includes a large number of directories but there is no test files in it as I can see...


src/Nest/Search/Suggesters/Suggest.cs
```diff
@@ -14,5 +15,10 @@ public class Suggest
 		public string Text { get; internal set; }
 		[JsonProperty("options")]
 		public IEnumerable<SuggestOption> Options { get; internal set; }
+
+		public IEnumerable<CompletionOption<T>> GetCompletionOptions<T>()
+		{
+			return Options.Select(x => new CompletionOption<T>(x));
```
Mpdreamz: Could use some C# 6 love, `Options?.Select(x => new CompletionOption<T>(x)) ?? Enumerable.Empty<CompletionOption<T>>();` and can be an expression bodied method :+1:

Mpdreamz:
Hi @tkirill thanks for the PR!

I'm fine with changing the signature of `SuggestOption`'s `Payload` from `object` to `T` directly. Since this is targeting `2.0` making this a breaking change is fine by me :+1:

src/Nest/Search/Suggesters/Suggest.cs
tkirill: @Mpdreamz fixed in [e17d7a5a5ce](https://github.com/tkirill/elasticsearch-net/commit/e17d7a5a5ce1c78bcac9b9c376a5884131faa03f#diff-4787c8ec5a0b0d0f96180eb10b40baf5L21)

tkirill:
@Mpdreamz I made `SuggestOption` generic and removed my `CompletionOption` class in e17d7a5a5ce.

Mpdreamz:
@tkirill I agree its strange to have both and it leads me to think `Suggest` such be generic as well, but this will mean the generic type has to trickle from request all the way to the response significantly increasing the scope of this PR. Is this something you are willing to take on?

tkirill:
@Mpdreamz Of course I am willing!  I think I could handle it but I need some guidance how to implement it in a proper way.  I understand the whole idea of propagation the type from request to response.  Maybe we would close this PR and I'll try to think of possible implementation?  And then we will discuss it in issue.

Won't it take you more time than if you would implement it yourself?

Mpdreamz:
True it'll be faster and we will get to it in #1481 but if you think its a fun exercise I'll more then happily guide you through it in an issue, on Skype whatever needs be :).  Fresh set of eyes on any part of the code base is worth tenfold!

michaelbudnik:
Hi guys,

Any updates on it?
Do you plan to include this in the 2.0 release?

tkirill:
@michaelbudnik I haven't finished the work.  I tried but it really requires a lot of changes just as @Mpdreamz said.  And then I just forgot about this, sorry.  If maintainers have no plans on this task I may try again.

Mpdreamz:
Hi guys,

This is something we will tackle as part of the 2.0 release. 

The story becomes a little bit more complicated since payloads are slated to be deprecated in elasticsearch 2.1 in favour of of suggestions v2.

https://github.com/elastic/elasticsearch/issues/10746

We'll have to come up with a good way to backport typed payloads to NEST 1.x and support suggestions v1 and v2 in NEST 2.x

tkirill:
Wow, it looks great!  I think this PR doesn't have any value now in context of suggestions v2.

gmarz:
This has finally been implemented via 02dd5353834a80fb5d5c57d1dace6d6e3ad582f5

Similar to inner hits, payloads are now "lazily" deserialized.

```csharp
new CompletionField<MyPayload>
{
    Input = "Nirvana",
    Output = "Nirvana - Nevermind",
    Payload = new MyPayload
    {
        Id = 1
    }
}
```

Pulling the payload out of  response looks like this:

```csharp
var payload = option.Payload<MyPayload>();
```

PR #1489, opened 2015-07-17
fix #1296 map get response on the update response

### LINKED ISSUE #1296: Update that returns fields from document
I am trying to execute an Update that returns fields from the source document.  It works fine when I do it manually (curl, etc) but when using the .Net client, I am able to specify the fields but Nest.IUpdateResponse doesn't seem to provide any way to get them.  In fact, response.ConnectionStatus.ResponseRaw is just null.  The RequestUri does contain the "&fields=_source" parameter so it seems like it's executing correctly.

Is this something that is just not supported or is there something else I need to do to get these fields from the response?

Mpdreamz:
Update API allows you to get the `_source` or `fields` in one go, this was blatantly missing from our mappings thanks @jcachat for reporting.

For backwards compat reasons this object is untyped.

### DIFF
--- a/src/Nest/Domain/Responses/UpdateResponse.cs
+++ b/src/Nest/Domain/Responses/UpdateResponse.cs
@@ -1,4 +1,6 @@
-﻿using Newtonsoft.Json;
+﻿using System.Collections.Generic;
+using Nest.Domain;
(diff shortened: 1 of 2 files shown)


gmarz:
LGTM

PR #1496, opened 2015-07-17
Fix/force invariant culture on all dates

### LINKED ISSUE #1227: Serialization of DateTime when TimeSeparator is a Dot
I've a unit test where I'm able to repro the problem, I explicitly set Italian culture with a dot as separator for DateTimeFormat.TimeSeparator

```csharp
 var cultureInfo = new CultureInfo("IT-it");
cultureInfo.DateTimeFormat.DateSeparator = ".";
cultureInfo.DateTimeFormat.TimeSeparator = ".";
Thread.CurrentThread.CurrentCulture = cultureInfo;
Thread.CurrentThread.CurrentUICulture = cultureInfo;

//I issue the query here
```

The query fails with an exception

ElasticsearchParseException[failed to parse date field [2015-01-30T08.52.32.443], tried both date format [dateOptionalTime], and timestamp number]; nested: IllegalArgumentException[Invalid format: **"2015-01-30T08.52.32.443"** is malformed at ".32.443"]

I Elastic Search needs semicolon to separate hours, minutes and milliseconds, and rejects the query. When a DateTime is converted to string, localization should not be considered. 

Actually I've patched my code using string instead of DateTime in Nest query.

```csharp
public const String DateTimeFormatForNestQuery = "yyyy-MM-ddThh:mm:ss.fff";

...

String startValueFormatted = startValue.ToString(DateTimeFormatForNestQuery, CultureInfo.InvariantCulture);
String endValueFormatted = endValue.ToString(DateTimeFormatForNestQuery, CultureInfo.InvariantCulture);

.OnField(d => d.PublishDate)
       .GreaterOrEquals(startValueFormatted)
       .LowerOrEquals(endValueFormatted)));   
```

Mpdreamz:
closes #1227

### DIFF
--- a/src/Nest/DSL/Query/RangeQueryDescriptor.cs
+++ b/src/Nest/DSL/Query/RangeQueryDescriptor.cs
@@ -188,25 +188,25 @@ namespace Nest
 
 		public RangeQueryDescriptor<T> Greater(DateTime? from, string format = "yyyy-MM-dd'T'HH:mm:ss.fff")
 		{
-			this.Self.GreaterThan = from.HasValue ? from.Value.ToString(format) : null;
+			this.Self.GreaterThan = from.HasValue ? from.Value.ToString(format, CultureInfo.InvariantCulture) : null;
 			return this;
 		}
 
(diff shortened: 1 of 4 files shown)


gmarz:
LGTM :+1:

PR #1510, opened 2015-08-05
Spike/2.0 pipeline rewrite

Mpdreamz:
This PR represents an ongoing effort to cleaning up some of the internals of NEST for the upcomming 2.0 release. 

Sending a PR because while not completely done, the tests pass and I rather send this now so that I can send smaller PR's against 2.0 afterwards.

### Transport
Completely rewritten from scratch, now models much more closely to our workflow 
clientpipeline6

This new setup also allows us to mock behavior in our unit tests much more explicitly rather then relying on hard to read FakeItEasy setups (as much as I love FakeItEasy!). There is a start of a DSL to tests anything related with the request pipeline in this PR but will build it out further in a second PR.

### IConnection

Now only has two methods on its interface and is directly responsible of return `T` rather then a `Stream` that is later turned into `T`.

### RequestData

All data needed for IConnection to make decisions, also responsible for creating responses and writing data on streams.

### Serializers

`INestSerializer` is gone, `IElasticsearchSerializer` greatly simplified and all Stream based. This means that we also try to serialize directly on the request stream when possible. No more `Stringify` this is now handled by `UrlFormatter` and `IQueryStringValue` objects can implement.

### Generated Code
Now almost all a one liner and following c# 6 patterns.

### Documentation
You can now include images in literate tests and they will be copied properly.

### Settings
Use c# 6 patterns and all timeouts are now timespans.

### Bye bye TPL hello await/async
NEST predates async/await and was written using TPL, making having use of ContinueWith, UnWrap and recursive iteration. This at the time was easier to program against then the Begin*/End* pattern but still grew unwieldy over time. Connection/Transport/Serializer have all been rewritten to use async/await and have become a lot easier to grasp in the process, all async operations allow a cancellationtoken to passed in on IRequestConfiguration.

### `PostData<T>`
Data being posted is now modeled rather then relying on `object` and `is` typechecks

### Object hierarchy

Custom IConnection and IElasticsearchSerializers now need to be passed to `ConnectionSettings` rather then directly on the client.


Mpdreamz:
Since this is a huuuuge PR me and @gmarz reviewed it together. This PR steps a great step in the right direction and is the way we want to move forward.

Issue #1533, opened 2015-08-24
NEST is missing the ability to specify the store property to the _timestamp field


ben-guri:
There isn't a way to specify the store property on the _timestamp field when programmatically creating a new mapping with NEST. TimestampFieldMappingDescriptor misses a property for store.

so that enabling _timestamp field for a mapping would store turned on would be:
ts.Enabled(true).Store(true);
(where ts is TimestampFieldMappingDescriptor)

Mpdreamz:
Thanks for reporting, will take this on board for the next release :+1:

ben-guri:
for when is it scheduled ?
thanks

KodrAus:
I've made a pull request for this, but am not ninja enough to attach it to the issue https://github.com/elastic/elasticsearch-net/pull/1537

gmarz:
@KodrAus FYI - it's just a matter of referencing the issue # in the PR summary or commit message and GitHub will automatically associate them.

gmarz:
Closed via #1537

PR #1537, opened 2015-08-25
add store to timestamp field mapping #1533

KodrAus:
#1533  Any unit tests to implement?

### DIFF
--- a/src/Nest/Domain/Mapping/SpecialFields/TimestampFieldMapping.cs
+++ b/src/Nest/Domain/Mapping/SpecialFields/TimestampFieldMapping.cs
@@ -24,7 +24,7 @@ namespace Nest
 		bool? IgnoreMissing { get; set; }
 		
 		[JsonProperty("store")]
-		bool Store { get; set; }
+		bool? Store { get; set; }
 	}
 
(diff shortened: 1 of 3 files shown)


gmarz:
Thanks for the PR @KodrAus ! I left one minor comment.

We have a few [unit tests for the timestamp field](https://github.com/elastic/elasticsearch-net/blob/develop/src/Tests/Nest.Tests.Unit/Core/Map/TimestampField/TimestampFieldTests.cs).  It would be great if you could include the `store` field in those tests as well.

Also, would you mind signing the [CLA](https://www.elastic.co/contributor-agreement/)?

KodrAus:
I've signed the CLA. Thanks for linking to the unit tests, I'll make the changes and add tests

gmarz:
Excellent, thank you!  One other nitpick I just noticed ^^.

KodrAus:
How's this looking now? It should be verifying that store is not serialized in `TimestampFieldUsingExpression` and that it is in `TimestampFieldUsingString`

gmarz:
LGTM!  Thanks again @KodrAus

KodrAus:
No worries! I'm trying to get more involved in the ES community, since I've been mooching off it for years

PR #1553, opened 2015-09-04
Added HighlightQuery to HighlightFieldDescriptor

Settler:
There are missing HighlightQuery method in HighlightFieldDescriptor and we cannot set query for every highlight field. I fixed that missing behavior.


gmarz:
LGTM.  Thanks @Settler!

PR #1558, opened 2015-09-14
NodeStats: fix typo + add HeapMax to JVM Mem stats

TheFireCookie:
So there is 2 commits. The first one just fix a type.

The second one add some statst to the jvm/mem route that were not mapped.

Also, what's the purpose of the empty string values in this code?

Like this one:
[JsonProperty("non_heap_used")]
public string NonHeapUsed { get; internal set; }

They come null when I use NEST.

### DIFF
--- a/src/Nest/Domain/Stats/NodeStats.cs
+++ b/src/Nest/Domain/Stats/NodeStats.cs
@@ -168,7 +168,7 @@ namespace Nest
 			[JsonProperty("actual_free")]
 			public string ActualFree { get; internal set; }
 			[JsonProperty("actual_free_in_bytes")]
-			public long ActualFreeInbytes { get; internal set; }
+			public long ActualFreeInBytes { get; internal set; }
 			[JsonProperty("actual_used")]
 			public string ActualUsed { get; internal set; }
 			[JsonProperty("actual_used_in_bytes")]
--- a/src/Nest/Domain/Stats/NodeStats.cs
+++ b/src/Nest/Domain/Stats/NodeStats.cs
@@ -167,6 +167,9 @@ namespace Nest
 			public int UsedPercent { get; internal set; }
 			[JsonProperty("actual_free")]
 			public string ActualFree { get; internal set; }
+			[Obsolete]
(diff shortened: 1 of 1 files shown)


gmarz:
@TheFireCookie thanks for the PR!  Unfortunately, simply renaming `ActualFreeInbytes` would be a breaking change in the `develop` branch, but we could definitely pull this change into `2.0`.

For `develop`, we need to keep the misspelled property and mark it `Obsolete` while adding the correctly spelled version.  Would you mind making that change and updating this PR?

> Also, what's the purpose of the empty string values in this code?

What do you mean exactly?  Are they coming back null in the response?

TheFireCookie:
OK I added the Obsolete flag and restored the old property. 

For the problem, here is some screenshots:

nullstring

Maybe ElasticSearch exposed the properties in the past but not anymore?

For this problem, if we agree that there is in fact a problem, I can tag them as Obsolete in develop et remove them in the branch 2.0

Last question: how can I do a new pull request for a clean change in the branche 2.0? (I'm using the GitHub Desktop client) I'm new to the whole branch/fork/git world.

Mpdreamz:
Those properties get set when you pass the `?human` flag to elasticsearch. 

Thanks for updating the PR against two branches! much appreciated :+1:

src/Nest/Domain/Stats/NodeStats.cs
```diff
@@ -167,8 +167,11 @@ public class MemoryStats : BaseMemoryStats
 			public int UsedPercent { get; internal set; }
 			[JsonProperty("actual_free")]
 			public string ActualFree { get; internal set; }
+			[Obsolete]
 			[JsonProperty("actual_free_in_bytes")]
```
Mpdreamz: This `JsonProperty` is a duplicate and needs to be removed, the obsoleted property should get and set the new property `ActualFreeInBytes`.

TheFireCookie:
Ok I guess it's fine then now? :)

For the ?human, I suppose we can then leave the strings fields like that :)

Mpdreamz:
All good! :+1:

PR #1562, opened 2015-09-15
NodeStats: fix typo + add HeapMax to JVM Mem stats

TheFireCookie:
As discussed previously, here is the clean PR for the 2.0


Mpdreamz:
LGTM :+1: thanks @TheFireCookie :)

Issue #1563, opened 2015-09-17
JsonProperty("protected_words_path ")


GokGokalp:
WordDelimiterTokenFilter.cs

Hello,
This property creating extra space and protected document can not work...
JsonProperty("protected_words_path ")

gmarz:
Closed via #1564

PR #1564, opened 2015-09-17
Update WordDelimiterTokenFilter.cs

robertlyson:
This one is fixing #1563.
By the way checking how PR from browser works in github :)


gmarz:
LGTM :+1: Thanks @robertlyson !

PR #1576, opened 2015-09-26
Index.url.cs added to Tests.csproj

robertlyson:
Ups ... Tests.csproj was referencing Indexing.url.cs instead of Index.url.cs.


Mpdreamz:
Thanks :+1: ! Sloppy commit on my end :-1:

Issue #1587, opened 2015-10-07
Support detect_noop in NEST


ppf2:
This is a request to add support for detect_noop option for updates in NEST

https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html?q=update%20a#_literal_detect_noop_literal

gmarz:
Closed via #1588

PR #1588, opened 2015-10-07
Support detect_noop in NEST #1587

robertlyson:
Hi Guys,

This is fix for https://github.com/elastic/elasticsearch-net/issues/1587.
Also I've created PR to address this in api spec https://github.com/elastic/elasticsearch/pull/14004.

### DIFF
--- a/src/CodeGeneration/CodeGeneration.LowLevelClient/ApiEndpoints/update.json
+++ b/src/CodeGeneration/CodeGeneration.LowLevelClient/ApiEndpoints/update.json
@@ -28,6 +28,10 @@
           "options": ["one", "quorum", "all"],
           "description": "Explicit write consistency setting for the operation"
         },
+		"detect_noop": {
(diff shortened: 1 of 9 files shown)


gmarz:
Awesome, thanks @robertlyson.  But since #14004 was merged, it would be better to run the code generator to pick up on `detect_noop` as a query string param.

gmarz:
Nevermind, just realized it was merged into ES 2.x and master, which isn't what the develop branch is pointed at.  In that case, we can just manually add it to https://github.com/elastic/elasticsearch-net/blob/develop/src/CodeGeneration/CodeGeneration.LowLevelClient/ApiEndpoints/update.json, then run the code generator.

robertlyson:
I see, so I'll update it manually, run and will be back with an update.

robertlyson:
Update done. Hope it's fine now.

gmarz:
LGTM!

PR #1593, opened 2015-10-15
Fix for CatAliases API #1591

### LINKED ISSUE #1591: CatAliases API
Hi
I am using CatAliases API to get the list of all aliases .
My request is something like 
ICatAliasesRequest catAliasesRequest = new CatAliasesRequest();
            return m_elasticClient.CatAliases(catAliasesRequest);

and i am trying to read every record received from response(IEnumerable<CatAliasesRecord>). I am observing that the records returned only have alias and index information and returns null for Search and Index routing .

Could anyone please confirm if this is a known issue or point me if i am doing something wrong ?

Thanks in advance !!

robertlyson:
Hi,

This PR fixes #1591.
One more thing fixed - there was no way to specify `search_routing` and `index_routing` parameters when [creating alias](https://www.elastic.co/guide/en/elasticsearch/reference/master/indices-aliases.html#aliases-routing).

### DIFF
--- a/src/Nest/DSL/PutAliasDescriptor.cs
+++ b/src/Nest/DSL/PutAliasDescriptor.cs
@@ -14,6 +14,10 @@ namespace Nest
 	{
 		[JsonProperty("routing")]
 		string Routing { get; set; }
+		[JsonProperty("search_routing")]
+		string SearchRouting { get; set; }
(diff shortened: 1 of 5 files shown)


gmarz:
LGTM :+1: Thank you @robertlyson.

PR #1615, opened 2015-11-12
Fix mixed up doc comments

samlbest:
Should and MustNot methods had their documentation reversed.


gmarz:
LGTM :+1:  Thanks @samlbest

PR #1616, opened 2015-11-16
IdsQueryDescriptor.Values(params long[] values) isn't assigning to Values correctly

jweber:
The `Values(params long[] values)` method in the IdsQueryDescriptor wasn't setting the `IIdsQuery.Values` property with the passed in array of long. This was causing a NullReference exception to be thrown when calling this method.


gmarz:
Yikes, good catch @jweber.  Thanks for the PR!

PR #1618, opened 2015-11-17
Make MultiField property descriptor Obsolete

TheFireCookie:
Not a lot to say. I guess that's what we wanted to do in the issue.


gmarz:
Thanks @TheFireCookie !

PR #1629, opened 2015-11-19
Fix SniffPath format issues against 2.0

koide:
The generated sniff path URL had timeout=00:00:02 by default. 

Elastic complained about no unit specified for timeout, making the call actually fail because it returned no nodes. 

I changed this to specify TotalSeconds seconds and now it "works for me". No formal testing has been performed.

### DIFF
--- a/src/Elasticsearch.Net/Transport/Pipeline/RequestPipeline.cs
+++ b/src/Elasticsearch.Net/Transport/Pipeline/RequestPipeline.cs
@@ -244,7 +244,7 @@ namespace Elasticsearch.Net.Connection
 			}
 		}
 
-		private string SniffPath => "_nodes/_all/settings?flat_settings&timeout=" + this.PingTimeout;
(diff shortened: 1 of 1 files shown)


Mpdreamz:
Hi @koide 

Thanks for the PR! Always appreciated :+1: 

I changed them to TimeSpan early on in the 2.0 branch but we still have a final pass over ConnectionSettings as a TODO before alpha. One of the items there is to change the TimeSpans to `TimeUnitExpression` see:

https://github.com/elastic/elasticsearch-net/blob/2.0/docs/contents/new/CommonOptions/TimeUnit/TimeUnits.asciidoc

This would take care of properly serializing them on the query string:

https://github.com/elastic/elasticsearch-net/blob/2.0/src/Nest/CommonOptions/TimeUnit/TimeUnitExpression.cs#L143

koide:
OK, that looks a bit more involved. I'll wait for your changes and keep my hack in the meantime. :) Thanks for answering!

PR #1628, opened 2015-11-19
Multisearch serialization/deserialization

michaelbudnik:
Still have some problems with setting error responses for inner responses but this can be done later.
This wasn't working at all and now the requests are serializing properly and correct responses are deserialized properly (except errored ones).

2/3 integration tests went from red to green (ExpectStatusCode, ExpectIsValid)

I have implemented ExpectResponse integration test and it still fails, but this is good because it points out some remaining problems (like this with error responses) which should be fixed.

Also...

Should this `new SearchRequest<Developer>("devs") { SearchType = SearchType.Count, Query = new QueryContainer(new MatchAllQuery()) } }` infer the type to be "developer"? Because it isn't and it not equals to the fluent equivalent of this request.

PR #1632, opened 2015-11-20
- fixes #1631 NestSerializer performance problems

### LINKED ISSUE #1631: 2.0 much worse serialization performance compared to 1.x
Sorry guys about that, but I may be reporting more stuff, because I'm evaluating upgrade to 2.0.
May help you with some stuff if I have time.
If you feel that's 2.0 branch is not in a state to report issues related to it just let me know and I'll stop =]

Testing with bulk index request (IndexMany).
2000 objects in a bulk (30 short fields on average per object).

**Used to take less than a second** with ES being the bottleneck but **now it takes about 20-25 seconds** where almost all the time the code is executing is on the client server serializing the request.
Comparing 2.0 (ES & Nest) vs 1.7 (ES & Nest)

Made some performance analysis with VS and it seems CPU hot paths mostly include
Newtonsoft.Json and **ElasticContractResolver**.

Tried with release build but no improvements.

michaelbudnik:
Fixes #1631 by not creating a new Newtosoft serializer for each Serialize/Deserialize method call.
I didn't see any need except formatting.

I think also about removing `SerializationFormatting` parameter from Serialize method and add it as a property to `IElastisearchSerializer` interface or even completely remove it. Who changes indenting per request?
This option should be saved at constructor time. 

What do you think?

### DIFF
--- a/src/Nest/CommonAbstractions/SerializationBehavior/NestSerializer.cs
+++ b/src/Nest/CommonAbstractions/SerializationBehavior/NestSerializer.cs
@@ -1,10 +1,6 @@
-using System;
-using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
-using Elasticsearch.Net;
 using Elasticsearch.Net.Serialization;
 using Nest.Resolvers;
 using Newtonsoft.Json;
@@ -21,7 +17,7 @@ namespace Nest
 		private readonly JsonSerializer _defaultSerializer;
 
 		public NestSerializer(IConnectionSettingsValues settings) : this(settings, null) { }
-		
+
 		/// <summary>
 		/// this constructor is only here for stateful (de)serialization 
 		/// </summary>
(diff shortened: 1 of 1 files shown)


src/Nest/CommonAbstractions/SerializationBehavior/NestSerializer.cs
```diff
@@ -36,24 +32,28 @@ public NestSerializer(IConnectionSettingsValues settings, JsonConverter stateFul
 
 		public void Serialize(object data, Stream writableStream, SerializationFormatting formatting = SerializationFormatting.Indented)
 		{
+			var defaultFormatting = this._defaultSerializer.Formatting;
+			this._defaultSerializer.Formatting = formatting == SerializationFormatting.Indented ? Formatting.Indented : Formatting.None;
```
Mpdreamz: This is not threadsafe we should probably have a dict of `<SerializationFormatting, Serializer>` that we instantiate in the constructor so that here we can just use it e.g `var serializer  = this._defaultSerializers[formatting]`

Mpdreamz:
Left one comment other then that this looks good to me :+1: 

Looking forward to more @michaelbudnik :smile:

Mpdreamz:
Hi @michaelbudnik pulled this and updated it to use a dict thanks for opening this.

lukapor:
Hey I see that issue and I must only share my findings. This findings are only info, I know that client is in development stage.
 I rewrite my application to work with elasticsearch 2.1. All my test degreaded, almost all test are 2 to 3 times slower. I did some profile and I find that main problem is serialize and deserialize. 

getgrouplist

gmarz:
Hey @lukapor thanks for sharing.  Just to confirm, does your build of the client contain https://github.com/elastic/elasticsearch-net/commit/1cb59b1350272ec272e8041db49f30f4f1679ce4?

michaelbudnik:
@lukapor I don't know how you tested it but I believe multisearch is not working at all at the moment in 2.0 (bad serialization/deserialization). I made a pull request #1628 with a partial solution for this (errors still not handled well).
tbh your results don't indicate that this is indeed a "hot path".

lukapor:
@gmarz client contains  #1631 fixes.

@michaelbudnik yes I did similar as you did it in #1628 that the multisearch works. But I saw that ReserializeJsonConverter.Reserialize is called on serialize and that was really slow.

Mpdreamz:
@lukapor @michaelbudnik thanks for reporting this guys, the serialization bits are very much in flux right now as you've noticed :smile: 

We are taking an outward in approach right now to get the API the way we want it and get our test coverage back up before we dive deep into the internals again and can refactor there with confidence. This work is almost done.

One of our goals is to get into a state where we can internalize Json.NET and Elasticsearch.NET dependencies for 2.0, for NEST 3.0 we hope to get into a state where almost all the json parsers/writers we own can be generated.

Both me and @gmarz have been been blocked on working on https://github.com/elastic/azure-marketplace for the past two weeks so we've been slow to respond to PR's and progress has been slow on the 2.0 branch. 

That said please keep reporting issues, I'd hate if we missed something (like the multisearch bug)).

PR #1633, opened 2015-11-22
set default value for the selector in DeleteIndexAsync implementation

jonyadamit:
Very minor change - this is my first contribution ever on github so I keep it simple to make myself familiar with the process (though the change really is needed..)


Mpdreamz:
LGTM :+1: 

In the future cases like these will be caught by our codestandards tests:

https://github.com/elastic/elasticsearch-net/blob/2.0/src/Tests/CodeStandards/ElasticClient.doc.cs#L64

Thanks for your very first PR, much appreciated and first of many I hope :smile:

PR #1635, opened 2015-11-25
- re-add Score property to IHit interface

michaelbudnik:
Cannot find any reasons for removing it.
This prevents getting Score by interface.


Mpdreamz:
:+1: thanks for spotting this @michaelbudnik

PR #1652, opened 2015-12-10
Update Purify.cs

wjt0818:
-1 error


Mpdreamz:
Thank you!, mind attaching a Test with this PR so its documented what this -1 error actually fixes?

Mpdreamz:
Even better would be to send the PR to the purify project: https://github.com/Purify-net/PUrify and add a test to it there. (might be easier as well since it has a purify dedicated test suite already)

PR #1657, opened 2015-12-10
PingTimeout does not use a time unit.

jonyadamit:
Since Elasticsearch 2.0, time units are obligatory.  
This causes connection pooling not to work because of a null reference exception in the `Sniff` method.

I would have liked to use `TimeUnitExpression` but that belongs to `Nest` only.  Maybe it should be moved to `Elasticsearch.net`?

### DIFF
--- a/src/Elasticsearch.Net/Transport/Pipeline/RequestPipeline.cs
+++ b/src/Elasticsearch.Net/Transport/Pipeline/RequestPipeline.cs
@@ -245,7 +245,7 @@ namespace Elasticsearch.Net.Connection
 			}
 		}
 
-		private string SniffPath => "_nodes/_all/settings?flat_settings&timeout=" + this.PingTimeout;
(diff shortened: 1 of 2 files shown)


Mpdreamz:
Just pushed a slightly different approach to fixing the tostring on timespan for sniff paths. 

Leaving this open as a reminder to investigate the string settings after lunch :+1:

russcam:
Hey @jonyadamit, I'm tidying up the branches to rename (branch move and delete)
- `2.0` to `master`
- `develop` (current 1.x client) to `1.x`

and as a consequence, closed this PR as a result of deleting the remote `2.0` branch, sorry! I've reinstated the branch to keep track of the open PRs but feel free to close this PR and open a new one with the change against  `master` :smile:

jonyadamit:
Hey @russcam, actually @Mpdreamz  only kept this PR open to remind him to investigate the string settings issue (see above).  If you fixed it already or have another way to keep track of that settings issue, this PR can be closed..

Mpdreamz:
thanks @jonyadamit node settings are always a string so updated with your other commit as well :+1: 

Also added sniff integration tests

PR #1677, opened 2015-12-21
I'd like to be able to iterate fields from outside nest.

jonyadamit:
For example, given a `SearchRequest`, I'd like to be able to determine which fields were included in the request (using `Source.Include`)


Mpdreamz:
LGTM :+1: thanks!

PR #1651, opened 2015-12-07
UriInfo class is improperly parsing URIs

jweber:
It looks like commit 5b0bd6c refactored this class and lost pieces of the code that respect port numbers as well as the absence of query strings in the given URI.

This was causing `System.ArgumentOutOfRangeException : StartIndex cannot be less than zero.` exceptions to be thrown from projects targeting .NET 4.6.

I just took the same code in the UriInfo class prior to commit 5b0bd6c and re-added it to the refactored class.

PR #1690, opened 2015-12-29
UriInfo class is improperly parsing URIs

jweber:
This is  a re-submit of PR #1651 against the master branch. I've also submitted this change as another pull request to the [PUrify-net](https://github.com/Purify-net/PUrify) repository.

----

It looks like commit 5b0bd6c refactored this class and lost pieces of the code that respect port numbers as well as the absence of query strings in the given URI.

This was causing `System.ArgumentOutOfRangeException : StartIndex cannot be less than zero.` exceptions to be thrown from my project which is targeting .NET 4.6.

I just took the same code in the UriInfo class prior to commit 5b0bd6c and re-added it to the refactored class.

### DIFF
--- a/src/Elasticsearch.Net/Purify/Purify.cs
+++ b/src/Elasticsearch.Net/Purify/Purify.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Globalization;
 using System.Reflection;
 
 namespace Purify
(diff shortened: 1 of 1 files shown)


Mpdreamz:
LGTM :+1: thanks for retargeting!

Issue #1666, opened 2015-12-16
Fix scripted update test using detect_noop


gmarz:
Spotted by @ppf2, [this test](https://github.com/elastic/elasticsearch-net/blob/develop/src/Tests/Nest.Tests.Integration/Core/UpdateTests.cs#L24) peforms a scripted update and specifies [detect_noop](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html?q=update%20a#_detecting_noop_updates), but detect_noop only applies to partial updates (where document merging takes place).  We should move this option over to [this test](https://github.com/elastic/elasticsearch-net/blob/develop/src/Tests/Nest.Tests.Integration/Core/UpdateTests.cs#L69) instead, as ES may eventually be changed to return a 400 in this scenario.

PR #1691, opened 2015-12-30
Move DetectNoop() to DocAsUpsert Test #1666

KodrAus:
For #1666 Looks like my VS might have fiddled with the indentation.


KodrAus:
CLA signed

Mpdreamz:
Thanks @KodrAus :+1: 

true default is fine, we typically follow this pattern in the fluent dsl `BoolProperty(bool property = true)` so `BoolProperty()` reads as switching it on and `BoolProperty(false)` turns it off explicitly. While not specifiying anything results as nothing being send.

PR #1692, opened 2015-12-30
JsonProperty now taken into account when inferring properies

Mpdreamz:
Now we ask the configured `IElasticsearchSerializer` if it can resolve the `MemberInfo` to a fieldname.

Precedence:

* NEST mapping attributes
* Ask serializer (default looks for JsonProperty)
* PropertyNameInferrer(MemberInfo.Name) (default camelCase)

also renamed NestSerializer to JsonNetSerializer,

### DIFF
--- a/src/Elasticsearch.Net/Serialization/ElasticsearchDefaultSerializer.cs
+++ b/src/Elasticsearch.Net/Serialization/ElasticsearchDefaultSerializer.cs
@@ -1,5 +1,6 @@
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -53,7 +54,9 @@ namespace Elasticsearch.Net
 			}
 		}
 
-		public static string RemoveNewLinesAndTabs(string input)
+		public string CreatePropertyName(MemberInfo memberInfo) => null;
+
+		private static string RemoveNewLinesAndTabs(string input)
 		{
 			return new string(input
 				.Where(c => c != '\r' && c != '\n')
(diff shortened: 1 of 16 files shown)


gmarz:
LGTM

russcam:
:+1: to this

RobinRieger:
Quick question, how can I revert to 'camelCase' as the default whilst keeping the JsonProperty as upper case? I.e. I have always had upper case JsonProperty, but my mapping was camelCase (hand made mapping file). With this change it appears I need to change all my JsonProperty tags to camelCase...... Is there a 'option' flag somewhere to set which of the three it uses? For more please see [stackoverflow question](http://stackoverflow.com/questions/34805468/elasticsearch-default-field-name-inferrer)

Mpdreamz:
its a cascade flow (as you guys found out in the comments on the SO answer) :+1: :

1. a hard property mapping on connectingsettings e.g `Rename()`
2. Nest mapping attribute on the property
3. Ask serializer what it thinks the field name should be
4. pass the property name to defaultfield name resolver

Your usecase in my opinion is quite exceptional, however there should be a way for to do what you want. 

I see two options:

1. implement a subclass of our JsonNetSerailizer that returns what you need for `CreatePropertyName`
2. Add a hook at the beginning of the resolve cascading flow. e.g another func on connectionsettings.

I just pushed a small fix to `CreatePropertyName` is virtual see also this test for an full example of how to override this in NEST 2.0:

https://github.com/elastic/elasticsearch-net/blob/6bf8305155c3abceff2572dc4b0c1025db0661d0/src/Tests/ClientConcepts/HighLevel/Inferrence/FieldNames/FieldInference.doc.cs#L243

We'll release an alpha2 today or monday.

gaurharirig:
connectionSettings.DefaultFieldNameInferrer(p => { if (p == "Id") return "_id"; else return p; }); is it not right .i am getting flase acknowldgement but no error

PR #1695, opened 2015-12-31
Fixed potentially incorrect combining

purenewman:
Right container wasn't checked properly.


Mpdreamz:
LGTM :+1: thank you for catching a potentially very nasty bug.

We only had a test if both sides were locked, added some tests locally that check either side only as well now that I will push post merge.

purenewman:
You're welcome. Elasticsearch.Net & NEST are important for me.

PR #1698, opened 2016-01-04
Added Parent and Routing fields to the Hit class

jweber:
I noticed that the `_parent` and `_routing` fields for search responses were not being mapped to the Hit class. This commit simply adds the mapping for those two fields.


gmarz:
Nice catch @jweber, thanks for the PR :+1:

PR #1703, opened 2016-01-06
Overriding of json net no longer a concern of connectionsettings

Mpdreamz:
removed:

```diff
- JsonSerializerSettingsModifier(Action<JsonSerializerSettings> modifier)		
-AddContractJsonConverters(params Func<Type, JsonConverter>[] contractSelectors)		
```

From `ConnectionSettings`. 

It now takes a `Func<TConnectionSettings, IElasticsearchSerializer> factory` to inject subclasses of `JsonNetSerializer` e.g

```charp
public class MyJsonNetSerializer : JsonNetSerializer
{
	public int X { get; set; } = 0;
	public MyJsonNetSerializer(IConnectionSettingsValues settings) : base(settings) { }

	protected override void ModifyJsonSerializerSettings(JsonSerializerSettings settings)
	{
		++X;
	}
}

[U] public void IsCalled()
{
	var connectionPool = new SingleNodeConnectionPool(new Uri("http://localhost:9200"));
	var settings = new ConnectionSettings(connectionPool, new InMemoryConnection(),s => new MyJsonNetSerializer(s));
	var client = new ElasticClient(settings);
	var serializer = ((IConnectionSettingsValues)settings).Serializer as MyJsonNetSerializer;
	serializer.X.Should().BeGreaterThan(0);
}
```

### DIFF
--- a/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
+++ b/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
@@ -38,14 +38,14 @@ namespace Elasticsearch.Net
 			: this(connectionPool, connection, null)
 		{ }
 
-		public ConnectionConfiguration(IConnectionPool connectionPool, IElasticsearchSerializer serializer)
-			: this(connectionPool, null, serializer)
+		public ConnectionConfiguration(IConnectionPool connectionPool, Func<ConnectionConfiguration, IElasticsearchSerializer> serializerFactory)
+			: this(connectionPool, null, serializerFactory)
 		{ }
 
 		// ReSharper disable once MemberCanBePrivate.Global
 		// eventhough we use don't use this we very much would like to  expose this constructor
-		public ConnectionConfiguration(IConnectionPool connectionPool, IConnection connection, IElasticsearchSerializer serializer)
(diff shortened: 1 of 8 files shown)


russcam:
:+1: much cleaner

gmarz:
LGTM

jonyadamit:
OK, so if I have a custom `JsonConverter` that I used to add with `AddContractJsonConverters`, what am I supposed to do now instead?  Create my own implementation of `JsonNetSerializer`, override `ModifyJsonSerializerSettings` and add my converter to the `Converters` collection?

Mpdreamz:
Close!,override:

```csharp
protected virtual IList<Func<Type, JsonConverter>> ContractConverters => null;
```

https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/CommonAbstractions/SerializationBehavior/JsonNetSerializer.cs#L24

jonyadamit:
OK, thank you for the swift reply!

I think this might need a bit more refactoring though.
Now I'm forced to specify the implementation of the `ConnectionPool`, which has no relation to this change. 
Maybe add another constructor overload? 
Move it to a method like the other settings?

Anyway, I love the alpha release, thank you guys :smile:

PR #1704, opened 2016-01-06
Fix/linefeeds

Mpdreamz:
This fixes the linefeeds after introducing `.gitattributes` its either this or removing `.gitattributes` :smile:


Mpdreamz:
:-1: will jump on this again when we are in calmer waters :ship:

PR #1721, opened 2016-01-13
Add back the ability to register callbacks and renamed to OnRequestCo…

### LINKED ISSUE #1712: Callback registered with ConnectionStatusHandler is never invoked
ConnectionConfiguration<T>.ConnectionStatusHandler allows you to register a callback that should be invoked on every request. Using 2.0.0-alpha1, however, the registered callback is never invoked. (The equivalent functionality worked fine with 1.7.1.)

The ConnectionStatusHandler method is a nice and simple way to handle logging needs. Is there an alternative method that could/should be used from version 2.0.0?

gmarz:
…mpleted

Closes #1712

### DIFF
--- a/src/Elasticsearch.Net/Transport/Transport.cs
+++ b/src/Elasticsearch.Net/Transport/Transport.cs
@@ -101,8 +101,7 @@ namespace Elasticsearch.Net
 				if (response == null || !response.Success)
 					pipeline.BadResponse(ref response, requestData, seenExceptions);
 
-				if (this.Settings.OnRequestCompleted != null)
-					this.Settings.OnRequestCompleted(response);
+				this.Settings.OnRequestCompleted?.Invoke(response);
 
 				return response;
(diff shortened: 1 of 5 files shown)


src/Elasticsearch.Net/Transport/Transport.cs
```diff
@@ -100,6 +100,10 @@ public ElasticsearchResponse<TReturn> Request<TReturn>(HttpMethod method, string
 				}
 				if (response == null || !response.Success)
 					pipeline.BadResponse(ref response, requestData, seenExceptions);
+
+				if (this.Settings.OnRequestCompleted != null)
```
Mpdreamz: nitpick: `this.Settings.OnRequestCompleted?.Invoke()`

gmarz: Done. Good call :+1:

PR #1726, opened 2016-01-14
Fix/norms mapping

gmarz:
- Refactoring of norms:
 - `NormsMapping` renamed to `Norms`
 - Implemented `NormsDescriptor` - previously `Norms()` took an object directly.

- Added a way to test only options that can be mapped via the fluent API

- Attribute types now live under their respective type folder instead of all being bunched under a single folder.


Mpdreamz:
LGM :+1:

PR #1738, opened 2016-01-19
Fixes for 2.0

lukapor:
Hi,

Here is my changes of library, that my code is working.

DateRangeExpression fix typo
TypeMapping missing SourceField
MultiSearchRequest operation index, type should override request index, type if exists


russcam:
Thanks for the PR @lukapor and for making me look at multisearch indices and types! :+1:

lukapor:
Thanks for merge.

New logic for setting index and types for multi search is ok, I am setting only operation index, request index/type is null by me.

russcam:
Awesome :smile:

Issue #1742, opened 2016-01-19
Looping Named Buckets


RobinRieger:
Is there a way in 2.0 to loop over named buckets (aggregations)?

In 1.7 I was able to do the following 

    foreach (var row in ((FiltersBucket) searchResultForTimePeriod.Aggregations["repeatBuyers"]).Aggregations.Aggregations) {}

The search looked something like this:

```
Filters
    ("repeatBuyers",
        fil => fil.Filters
            (f => f.Name(((int) RepeatBuyersGrouping.One).ToString())
                .Range
                    (ra => ra.OnField(of => of.TotalOrders)
                        .LowerOrEquals(1)),
            f => f.Name(((int) RepeatBuyersGrouping.TwoFive).ToString())
                .Range
                    (ra => ra.OnField(of => of.TotalOrders)
                        .Greater(1)
                        .LowerOrEquals(5)),
...
```

This way I could use the bucket name and the value in the output. I.e. The name of each bucket is a integer and it maps to an enum which I can then use to get the proper 'display' text' for the output. 

With 2.0, I can't seem to loop the 'NamedBucket'. Do I need to call them separately now or was this accidentally dropped?

russcam:
Here's how you would do it in NEST 2.x

```c#
var names = new[] {
	((int)RepeatBuyersGrouping.One).ToString(),
	((int)RepeatBuyersGrouping.TwoFive).ToString()
};

var response = client.Search<Example>(s => s
	.Aggregations(aggs => aggs
		.Filters("repeatBuyers", fil => fil
			.NamedFilters(f => f
				.Filter(names[0], q => q
					.Range(ra => ra
						.Field(of => of.TotalOrders)
						.LessThanOrEquals(1)))
				.Filter(names[1], q => q
					.Range(ra => ra
						.Field(of => of.TotalOrders)
						.GreaterThan(1)
						.LessThanOrEquals(5))
				)
			)
		)
	)
);

var filters = response.Aggs.Filters("repeatBuyers");
foreach (var name in names)
{
	var filter = filters.NamedBucket(name);
	// do something with filter
}
```

The `Aggegations` property on `FiltersBucket` is `private` - @Mpdreamz / @gmarz, is this an intentional change?

RobinRieger:
To clarify, what you are suggesting is to no longer loop the buckets as such but the array I have that I used to make the different buckets. I tried your code and it seems to work (will check more over coming days)

[...]

PR #1752, opened 2016-01-24
Update FiltersBucket.cs

RobinRieger:
Needed a way to access all the buckets in the result. The named bucket is great for single bucket access but in certain cases you may want to just loop all the buckets with the name and the bucket content in the output. See [here](https://github.com/elastic/elasticsearch-net/issues/1742) for example. 

Just made the Accessibility Level `Public` to allow this.


gmarz:
Thanks for the PR @RobinRieger - LGTM.  I can't remember why we made them private to begin with, but thinking about it now, there's no reason they should be.

PR #1765, opened 2016-01-28
Refactor/expressionvisitor

### LINKED ISSUE #1089: ExpressionVisitor?
I noticed that NEST has an old [`ExpressionVisitor`](https://github.com/elasticsearch/elasticsearch-net/blob/develop/src/Nest/Resolvers/ExpressionVisitor.cs) from the times when it was part of the sample codes  of LINQ. 
This would make sense if NEST were compatible with .NET 3.5, but the DLL that comes with the NuGet package is .NET 4.0 .
Is there any other reason to keep that around?
IMHO .NET's current [ExpressionVisitor](http://referencesource.microsoft.com/#System.Core/Microsoft/Scripting/Ast/ExpressionVisitor.cs,f0615a823a9efe36) is cleaner and prevents some invalid rewrites.
I'm asking not just to potentially clean this up but because I have some small code using ExpressionVisitor that I might contribute.

gmarz:
Closes #1089


gmarz:
ping @Mpdreamz this is ready to be reviewed again

src/Nest/CommonAbstractions/Infer/Field/FieldResolver.cs
```diff
@@ -12,15 +12,15 @@
...
-	//replaces my sloppy 300+ lines (though working!) first attempt, thanks Thomas Levesque.	
-	public class FieldResolver : ExpressionVisitor
+	public class FieldResolver
 	{
 		private readonly IConnectionSettingsValues _settings;
 
 		private readonly ConcurrentDictionary<Field, string> Fields = new ConcurrentDictionary<Field, string>();
 		private readonly ConcurrentDictionary<PropertyName, string> Properties = new ConcurrentDictionary<PropertyName, string>();
 
+		protected Stack<string> Stack { get; set;}
```
Mpdreamz: Can be removed

PR #1771, opened 2016-01-29
Refactor/aggs metadata

gmarz:
Support aggregation meta data https://www.elastic.co/guide/en/elasticsearch/reference/2.1/agg-metadata.html (added by our very own @Mpdreamz  :) )

This PR also introduces another refactoring to aggregations which I think greatly simplifies the object graph (see https://github.com/elastic/elasticsearch-net/commit/8bdcbb3c27051175ec4e2f9865ea9e099b8c2014 and c509515).


src/Nest/Aggregations/Aggregation.cs
```diff
@@ -1,32 +1,26 @@
...
-	/// </summary>
-	[ExactContractJsonConverter(typeof(AggregationJsonConverter))]
-	public interface IAggregation
-	{
-	}
-
-	/// <summary>
-	/// Base for the OIS Aggregation DSL
-	/// </summary>
-	public interface IAggregationBase : IAggregation
```
Mpdreamz: I like this but I'm wondering why we didn't do this in the first place, do you know @gmarz?

src/Nest/Aggregations/AggregationResult.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Nest
+{
+	[ExactContractJsonConverter(typeof(AggregationResultJsonConverter))]
```
Mpdreamz: The JsonConverter name does not match the type its handling

src/Tests/tests.yaml
```diff
@@ -1,5 +1,9 @@
 ﻿# mode either u (unit test), i (integration test) or m (mixed mode)
+<<<<<<< .mine
+mode: i
```
Mpdreamz: Unresolved merge conflict here

src/Nest/Aggregations/Bucket/Bucket.cs
```diff
@@ -1,35 +1,50 @@
...
-		public Bucket(IDictionary<string, IAggregation> aggregations) : base(aggregations) { }
+		public DocCountBucket() { }
+		public DocCountBucket(IDictionary<string, IAggregationItem> aggregations) : base(aggregations) { }
 
 		public IList<TBucketItem> Items { get; set; }
 	}
 
-	public abstract class BucketBase : AggregationsHelper, IBucket
+	// Intermediate object used for deserialization
+	internal class BucketDto : IAggregationResult
```
Mpdreamz: I'm really hating DTO as a suffix but can not suggest a more meaningful name for this :smile:

src/Nest/Aggregations/Aggregation.cs
gmarz: I'm not really sure of original reasoning, but we didn't have to.  I think the new separation is much cleaner.

src/Nest/Aggregations/AggregationResult.cs
gmarz: Yea, I'm having trouble coming up with a batter name for `IAggregationItem` and `IAggregationResult`.  `IAggregation` would be good, but that already exists for requests.

src/Tests/tests.yaml
gmarz: doh, thanks

src/Nest/Aggregations/Bucket/Bucket.cs
gmarz: Likewise.  The entire thing is a code smell IMO...I've made it internal so now we can work on getting rid of it entirely by refactoring the json converter bits.

russcam:
Reviewing now...

russcam:
LGTM :+1:

RobinRieger:
I just downloaded the latest code and it gives me some compile errors re the `DocCount` after I changed all the references from `items` to `buckets`. I think `DocCount` used to be a `long` in the `SingleBucket`/`KeyedBucket` but is now a `long?` in the new `KeyedBucketItem`. 

Is it intentional to be nullable? When will `DocCount` be null? If it can't find anything shouldn't it be zero?

gmarz:
@RobinRieger making them nullable made testing easier, but you're right, they shouldn't be.  Will fix this for the next release, thanks :+1:

Issue #1609, opened 2015-11-09
Indexing JObject documents does not work as expected


JulianRooze:
An issue occurs for us when you try to index a `JObject` as a document (using Nest) like this:

```c#
client.Index<JObject>(finalProposition, i => i
.Index("telecom_propositions")
.Type("proposition")
.Id(docId))
```

The request fails with a 400 because of "malformed content". When you inspect the request being sent, the JSON at first appears to miss its opening bracket. It took me a while to figure it out but it's not a malformed object that's being sent, but rather a list of JSON *properties* concatenated by a newline. Properties don't start with a bracket which explains the missing bracket. 

The reason that it's not serializing the JObject as an object is this code:

https://github.com/elastic/elasticsearch-net/blob/develop/src/Elasticsearch.Net/Connection/RequestHandlers/RequestHandlerBase.cs#L62

It checks if the data is an `IEnumerable<object>`, which is true because a `JObject` *is* an `IEnumerable<object>` and from then on it treats the `JObject` as a list of items to serialize and not a singular object. 

Expected behavior for me would be that it just serialized the whole `JObject` at once and that it would be possible to index a document from a `JObject`. 

The reason this behavior is unexpected to me is that using the search API to get back a JObject works perfectly. It also works for the update document API (we use the DocAsUpsert route as a workaround for this issue) but not with the index API. It also works as expected when you use the bulk API to index `JObject` documents.

Suggested fix would be either a special-case check if `data is JObject` or perhaps for `data is IDictionary<string, object>` in which case it would just serialize the whole object as happens on line 66.

amccool:
I'm getting this when creating bulk descriptors of indexes using JObjects

amccool:
<code>
    private async void blahblah(IEnumerable < JObject >  jos)
            {
            var bdesc = new BulkDescriptor();
            foreach (var jo in jos)
            {
            bdesc.Index<JObject>(i => i
            .Index(Index)
            .Type("Trace")
            .Document(jo));
            }
            //this yeilds the same as the above loop
           //bdesc.IndexMany(jos, (f,g) => f.Index(Index).Type("Trace"));
            var x = Client.Bulk(bdesc);
        }
</code>

russcam:
[I've added some integration tests to NEST 2.x](https://github.com/elastic/elasticsearch-net/commit/4a93517cf1b89baba7e9fbf4b1339effe00586cd) for `JObject` and they get indexed as expected.

[...]

PR #1804, opened 2016-02-05
Align serialization with NEST 2.0 implementation

russcam:
This is to handle serializing `JObject` using the Json.NET contract for it and not treating it as an `IEnumerable<object>`

See https://github.com/elastic/elasticsearch-net/issues/1609

/cc @gmarz @Mpdreamz


Mpdreamz:
:+1: glad a cleaner fix was possible!

PR #1805, opened 2016-02-05
Feature/build improvements

Mpdreamz:
* `build.bat skiptests` now works again
* download of jetbrains profiler tooling and nuget is lazy (only some targets need them)
* depend on prelease version of GitLink that allows us to selectively include the projects to sourcelink, will remove errors on build outputs without pdb's


gmarz:
LGTM

PR #1806, opened 2016-02-05
Leverage aggregation metadata for deserialization

gmarz:
The current method of deserializing aggregation results is complex, difficult to maintain and prone to errors (we've had a number of issues in the past where the aggregation response wasn't being read to completion).

Additionally, since there's no way of knowing the type of aggregation at the time of deserializing, we've had to put hacks in place like the DTO for buckets (`BucketAggregate`) which is a bag of properties for all bucket types.  It is then mapped to the real bucket type via `AggregationsHelper` at the user's request.  One problem with this approach is that one can't cast directly from `IAggregate` to the real bucket type directly from the aggregation dictionary, and instead only to `BucketAggregate` which will contain irrelevant data.

This PR solves these problems and simplifies deserialization by leveraging [aggregation metadata](https://www.elastic.co/guide/en/elasticsearch/reference/2.2/agg-metadata.html) and associating the return type with each aggregation as part of the request.

Each aggregation request object is marked with an `AggregateTypeAttribute` that determines what aggregate type to include in the meta data.  During serialization, this attribute is retrieved and injected into the request.

For instance, the request for a terms aggregation which returns a `TermsAggregate` now looks like this:

```csharp
"my_terms" : {
    "meta": {
        "aggregate_type": "Nest.TermsAggregate"
    },
    "terms": {
        "field": "my_field"
    }
}
```

`AggregateJsonConverter` can now look for the value of `aggregate_type` in the returned metadata for each aggregation result and deserialize directly to that type.

Injected meta data is never returned to the user.


gmarz:
Continuing this work in a [new branch](https://github.com/elastic/elasticsearch-net/tree/feature/aggs-metadata), since this diverged too much from master.

Will open a new PR once it's ready.

PR #1814, opened 2016-02-09
Added a null check to GeoLocation and updated ToString()

niemyjski:
The ToString() must be in a specific format for lat long of which we know... 

Also there is only so much precision you can specify (http://www.geomidpoint.com/latlon.html)

You wouldn't want a number converted to 1.3e6 or 1.12345678901234545666666556 when 8 points of precision gets you to 1/16th of an inch

### DIFF
--- a/src/Nest/QueryDsl/Geo/GeoLocation.cs
+++ b/src/Nest/QueryDsl/Geo/GeoLocation.cs
@@ -82,7 +82,7 @@ namespace Nest
 
 		public override string ToString()
 		{
-            return _latitude.ToString("#0.0#####") + "," + _longitude.ToString("#0.0#####");
+            return _latitude.ToString("#0.0#######") + "," + _longitude.ToString("#0.0#######");
 		}
 
 		public bool Equals(GeoLocation other)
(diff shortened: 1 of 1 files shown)


Mpdreamz:
This LGTM :+1: thanks @niemyjski 

I'd like to open the discussion though that our implicit conversions should not throw but return null as per c# guidelines /cc @gmarz @russcam

niemyjski:
hmm, yeah that's a good point.

niemyjski:
https://msdn.microsoft.com/en-us/library/ms229033.aspx I wasn't aware of that rule..

russcam:
Agreed on the implicit conversions returning `null` instead of throwing; if there is a need for a conversion to throw then it should ideally be `explicit` but then that runs contrary to the value of the implicit conversion :smile:

gmarz:
> Agreed on the implicit conversions returning null instead of throwing

+1 good call @Mpdreamz

niemyjski:
Do you guys want me to fix this.. looks like there is still an issue with this format: https://github.com/exceptionless/Exceptionless.Net/pull/80

PR #1822, opened 2016-02-10
Ignore project.lock.json files

gmarz:
I know there are good reasons to check in project.lock.json, but the official stance is to not.  It makes merging a PITA and adds noise to diffs.

Thoughts @Mpdreamz @russcam ?


Romanx:
I'd add a :+1: to this as it's a pain working locally with them updating on you and having to ignore them if you `git add -p`

Mpdreamz:
project.lock.json really needs to [separate its concerns](https://github.com/aspnet/dnx/issues/2332), since we are never *ever* going to use floating dependencies we could possibly decide to not check these in.

But only because nuget resolves transient dependencies to lowest ([which is a horrible default](https://github.com/NuGet/Home/issues/1192)), unless ofcourse someone's machine has overriden this in a nuget.config at a machine/user profile level using

```xml
<add key="DependencyVersion" value="Highest" />
```

Then our build can possibly be broken on their machine and they'd never know why, repeatable builds are lost if we pull this one.

I'm +1 for pulling this one in though since I'm still holding my breath for either paket and dotnet-cli playing nice or atleast aspnet/dnx#2322 landing before the whole toolchain GA's

Mpdreamz:
@gmarz mind rebasing against master?

russcam:
Happy to keep them out of commits and reduce the noise :+1:

Mpdreamz:
Added this in 3deddcb

Issue #1820, opened 2016-02-10
NullReferenceException from BoolQuery::IsConditionless


WojciechLuczkow:
Exception occurs if any enumerable in BoolQuery contains null values.
`qq.IsConditionless` -> exception if qq == null

Of course QueryContainers should not be null but it could be either ArgumentException thrown in setter or null values ignored

Before update to 2.0 null elements were ignored (at least it worked that way)

gmarz:
This sounds related to #1817 which we already have a fix for.  @russcam can you confirm?

russcam:
@gmarz in a similar area but an additional case that was only partially handled; in a collection of QueryContainers, if the first item was `null` then a `NullReferenceException` would be thrown, but if the first item was not null, no exception would happen and the null query would be filtered out as part of changes in #1817.

@WojciechLuczkow, thanks for reporting as it uncovered a few exceptional cases that also needed to be handled :+1:

PR #1823, opened 2016-02-11
Handle collection of QueryContainer where the first item is null

russcam:
Closes https://github.com/elastic/elasticsearch-net/issues/1820
Additional tests to ensure parity between fluent and OIS


gmarz:
LGTM

Issue #1826, opened 2016-02-11
CreateIndex with Attachment will fail


kyakhta:
```c#
 client.CreateIndex(Index, c => c
                .Mappings(m => m.Map<Docx>(p => p
                .Properties(ps => ps.Attachment(a => a.FileField(t => t.Name(e => e.Foo))))
                    .Properties(ps => ps.String(s => s.Name(o => o.Title))))));
```

will fail on `a => a.FileField(t => t.Name(e => e.Foo)` because `AttachmentPropertyDescriptor.Dictionary` is not initialized and has null value:

```c#
private AttachmentPropertyDescriptor<T> SetMetadataField<TDescriptor, TInterface>(Func<TDescriptor, TInterface> selector, string fieldName)
			where TDescriptor : TInterface, new()
			where TInterface : IProperty
		{
			selector.ThrowIfNull(nameof(selector));
			var type = selector(new TDescriptor());
			**this.Dictionary[fieldName] = type;**
			return this;
		}
```

russcam:
Thanks for reporting @kyakhta :+1:

gmarz:
Closed via #1827

PR #1827, opened 2016-02-12
Fix/mappings

russcam:
See https://github.com/elastic/elasticsearch-net/issues/1826
Ensure `Fields` is initialized before adding attachment metadata fields.
Add overloaded fluent method signatures that use the correct property types for language and content_length.

Add Visit methods for all supported `IProperty` types
Add `NoopMappingVisitor` to make it easier to write a custom `IMappingVisitor` implementation

### DIFF
--- a/src/Nest/Mapping/Visitor/IMappingVisitor.cs
+++ b/src/Nest/Mapping/Visitor/IMappingVisitor.cs
@@ -1,10 +1,13 @@
-﻿namespace Nest
+﻿using System;
+
+namespace Nest
 {
 	public interface IMappingVisitor
 	{
 		int Depth { get; set; }
 		void Visit(TypeMapping mapping);
 		void Visit(StringProperty mapping);
+		[Obsolete("Use Visit(NumberProperty mapping) for number mappings")]
 		void Visit(NumberType mapping);
 		void Visit(DateProperty mapping);
 		void Visit(BooleanProperty mapping);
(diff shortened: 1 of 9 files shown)


src/Nest/Mapping/Visitor/IMappingVisitor.cs
```diff
@@ -1,10 +1,13 @@
...
+﻿using System;
+
+namespace Nest
 {
 	public interface IMappingVisitor
 	{
 		int Depth { get; set; }
 		void Visit(TypeMapping mapping);
 		void Visit(StringProperty mapping);
+		[Obsolete("Use Visit(NumberProperty mapping) for number mappings")]
```
Mpdreamz: Good catch :+1:

src/Tests/Mapping/Types/Specialized/Attachment/AttachmentMappingTests.cs
```diff
@@ -0,0 +1,126 @@
...
+				.ContentTypeField(d => d
+					.Name(n => n.ContentType)
+				)
+				.DateField(d => d
+					.Name(n => n.Date)
+				)
+				.KeywordsField(d => d
+					.Name(n => n.Keywords)
+				)
+				.LanguageField((StringPropertyDescriptor<AttachmentTest> d) => d
```
Mpdreamz: I'm a tad conflicted about this, `LanguageField` now requiring a cast seems wrong. We might need a temporary LangField() mapping?

PR #1835, opened 2016-02-13
Updated documentation

rdehuyss:


Mpdreamz:
:+1: thanks for this, should cause less confusion if you are replaying commands from the readme

Issue #1836, opened 2016-02-15
Error on GeoDistanceSort [2.0.2]


RTodorov:
Hi guys,

I'm adding a GeoDistanceSort to my query but the result json is ignoring GeoUnit and DistanceType:

```
var searchSorting = new List<ISort>();

...

searchSorting.Add(new GeoDistanceSort
{
    Field = "address.location",
    Points = new List<GeoLocation> { new GeoLocation(location.Coordinates.Lat, location.Coordinates.Lon) },
    GeoUnit = DistanceUnit.Kilometers,
    Order = SortOrder.Ascending,
    DistanceType = GeoDistanceType.Plane
});

...

var request = new SearchRequest
{
    Query = myQuery,
    Sort = searchSorting,
};
```

The result json:

```
  "sort": [
    {
      "_geo_distance": {
        "address.location": [
          {
            "lat": -23.549846028909087,
            "lon": -46.70886704698205
          }
        ],
        "order": "asc"
      }
    }
  ],

```

It's missing ```unit``` and ```distance_type```

Any ideas?

Thank you!

Mpdreamz:
Not immediately, we are all travelling and attending [elasticon](https://www.elastic.co/elasticon/conf/2016/sf) the coming days so we might take a tad longer then usual to respond :(

Thanks for reporting, always very much appreciated :+1:

russcam:
Thanks @RTodorov - the type was missing a couple of Json.Net attributes to ensure the properties were serialized when set. This is fixed in PR https://github.com/elastic/elasticsearch-net/pull/1841

RTodorov:
Thank you @russcam ! Enjoy elasticon!

gmarz:
Closed via #1841

PR #1841, opened 2016-02-16
Include unit and distance_type when set

russcam:
See https://github.com/elastic/elasticsearch-net/issues/1836


Mpdreamz:
LGTM :+1:

PR #1858, opened 2016-02-22
Added lock around adding to _enumCache

zeus82:
This is to fix issue 1857


Mpdreamz:
Thanks for reporting this one @zeus82, I pushed a slightly different solution :+1: 

We'll release a patch release later today

PR #1874, opened 2016-03-01
improved dispatch exception to indicate which values are missing

Mpdreamz:
Dispatches into the low level client that do not have enough information to build a route will now indicate which values are missing

> Dispatching Index() from NEST into to Elasticsearch.NET failed
Received a request marked as POST
This endpoint accepts POST,PUT
The request might not have enough information provided to make any of these endpoints:
  - /{index=`<NULL>`}/{type=project}
  - /{index=`<NULL>`}/{type=project}/{id=`<NULL>`}

It now also throws an ArgumentException since its very much a "development time exception" and happens before we even establish a pipeline.


gmarz:
Awesome...left a minor nitpick comment, otherwise looks good.

PR #1875, opened 2016-03-01
Fix #1867 objects can now be serialized in register percolater metadata

### LINKED ISSUE #1867: RegisterPercolatorRequest throws exception when Object is passed in the Metadata
## Test case 1: NEST 1.7	

Success and creates the precolator:

```
	class Program
	{
		private static readonly ILog _log = LogManager.GetLogger(typeof(Program));

		{
			ElasticClient client = new ElasticClient(new ConnectionSettings(new Uri("http://172.19.131.118:9200"), "index"));
			var session = new Session();

			var response = client.RegisterPercolator(new RegisterPercolatorRequest("myindex", Guid.NewGuid().ToString())
			{
				Query = new MatchAllQuery(),
				MetaData = new Dictionary<string, object> 
				{ 
					{ "Name" , "1" }, 
					{"Data", session } // <<< FAIL ON THIS WHEN AN OBJECT IS PASSED 
				}
			});

			var responseRaw = response;
			Console.WriteLine(responseRaw.ConnectionStatus.ResponseRaw.ToString());
		}
	}

	internal class Session
	{
		public object Performers { get; set; }
		public string Composers { get; set; }
	}
```

## Test case 2: NEST 2.0

Fails to serialize class Session

```
	class Program
	{
		private static readonly ILog _log = LogManager.GetLogger(typeof(Program));

		{
			ElasticClient client = new ElasticClient(new ConnectionSettings(new Uri("http://172.19.131.118:9200")));
			var session = new Session();

			var response = client.RegisterPercolator(new RegisterPercolatorRequest("myindex", Guid.NewGuid().ToString())
			{
				Query = new MatchAllQuery(),
				Metadata = new Dictionary<string, object> 
				{ 
					{ "Name" , "1" }, 
					{"Data", session } // <<< FAIL ON THIS WHEN AN OBJECT IS PASSED 
				}
			});

			var responseRaw = response;
			Console.WriteLine(responseRaw.CallDetails);
		}
	}

	internal class Session
	{
		public object Performers { get; set; }
		public string Composers { get; set; }
	}
```

```
An unhandled exception of type 'Elasticsearch.Net.UnexpectedElasticsearchClientException' occurred in Elasticsearch.Net.dll

Additional information: Unsupported type: ConsoleApplication1.Session. Use the JsonSerializer class to get the object's JSON representation. Path ''.
```

Looks like https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Search/Percolator/RegisterPercolator/RegisterPercolatorJsonConverter.cs#L52 is failing

Looks like the class `Session` should implement `JsonConverter` with the ReadJson and WriteJson in order to do this. However it will be nice to automatically conver this class to the Json Representation as we did in 1.x.

gmarz:


src/Nest/Search/Percolator/RegisterPercolator/RegisterPercolatorJsonConverter.cs
```diff
@@ -49,7 +49,14 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
 				foreach(var kv in request.Metadata)
 				{
 					writer.WritePropertyName(kv.Key);
-					writer.WriteValue(kv.Value);
+					if (kv.Value.GetType().IsValueType)
```
Mpdreamz: style wise single method bodies can drop the {} IMO.

PR #1887, opened 2016-03-03
Fix handling of return types.

rossbrower:
Calling a method on the low level client with Stream, VoidResponse, etc. fails with a NullReferenceException. This change fixes the reflection calls which (while seemingly correct) do not work at runtime.


gmarz:
Yikes!  `cs.Body` is null in this case and the `is` operator only works on non-null expressions.

  Nice find @rossbrower and thanks for the PR. :+1:

PR #1888, opened 2016-03-03
Fix #1856 respect DisableDirectStreaming setting on error responses

### LINKED ISSUE #1856: DisableDirectStreaming doesn't seem to make any difference
I have a validate query method which does some Json validation, then fires off a query to elastic to count hits. Unfortunately, when the query is invalid, I can't get the error message back from elastic.

Since this is a validation method, performance is irrelevant.

I'm building the client as follows:
```
private ElasticClient BuildElasticClient(bool disableDirectStreaming=false) {
	var node = new Uri(ElasticServer);
	var config = new ConnectionSettings(node);
	config.DisableDirectStreaming(disableDirectStreaming);
	return new ElasticClient(config);
}
```
And calling it like this:

```
var client = BuildElasticClient(true);
var response = client.LowLevel.Search<Elasticsearch.Net.DynamicResponse>(finalQuery);
```

with a malformed query (I was using "boolean" instead of "bool".

Whether I pass in true/false to my build function, the debug information is exactly the same...

```
Unsuccesful low level call on POST: /_search
 - BadResponse: Node: http://elastic:9200/ Took: 00:00:00.2725346
# OriginalException: System.Net.WebException: The remote server returned an error: (400) Bad Request.
   at System.Net.HttpWebRequest.GetResponse()
   at Elasticsearch.Net.HttpConnection.Request[TReturn](RequestData requestData)
# Request:
{"query": {
  "boolean": {
    "must": [
      {"match": {"Body": "Example"}},
      {"match": {"Author": "Bob"}}
    ]
  }
}, "size": 0}
# Response:
<Response stream not captured or already read to completion by serializer. Set DisableDirectStreaming() on ConnectionSettings to force it to be set on the response.>
```

I've googled repeatedly and can't find anything that explains how to get details of the error returned by elastic. Most of the information I can find is extremely out of date (referencing methods that don't seem to exist any more)

Am I missing something or is this broken?

gmarz:


Mpdreamz:
LGTM but misses a test

PR #1912, opened 2016-03-14
Fix #1904 invoke OnRequestCompleted when exception is thrown

### LINKED ISSUE #1904: OnRequestCompleted doesn't work if ThrowExceptions is enabled
Title says it all. What I expected is that OnRequestCompleted would handle all completed requests, successful and failed. But if ThrowExceptions is enabled, the exception throwing takes place before calling OnRequestCompleted, which defeats the purpose of having a tracing handler.

I checked on your code.

Transport.Request<TReturn> calls

```
if (response == null || !response.Success)
  pipeline.BadResponse(ref response, requestData, seenExceptions);

this.Settings.OnRequestCompleted?.Invoke(response);
```

Would it be a really big problem if you changed the order to call OnRequestCompleted before and not after?

Thanks a huge lot

gmarz:
Closes #1904

### DIFF
--- a/src/Elasticsearch.Net/Transport/Pipeline/RequestPipeline.cs
+++ b/src/Elasticsearch.Net/Transport/Pipeline/RequestPipeline.cs
@@ -459,7 +459,11 @@ namespace Elasticsearch.Net
 				AuditTrail = this.AuditTrail
 			};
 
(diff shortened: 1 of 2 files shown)


Mpdreamz:
LGTM :+1:

PR #1914, opened 2016-03-15
Defer field naming conventions to Elasticsearch

### LINKED ISSUE #1889: ArgumentExceptions in FieldResolver bubble out as UnexpectedElasticsearchClientExceptions
The underlying exception is an [ArgumentException](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/CommonAbstractions/Infer/Field/FieldResolver.cs#L56) and it should bubble out as an `ElasticsearchClientException`.  The problem lies in the fact that `.Resolve()` is called later in the pipeline.

russcam:
Don't throw an ArgumentException if a field name contains a dot (.), instead let Elasticsearch decide.
Fixes #1889

### DIFF
--- a/src/Elasticsearch.Net/Transport/Transport.cs
+++ b/src/Elasticsearch.Net/Transport/Transport.cs
@@ -15,7 +15,7 @@ namespace Elasticsearch.Net
 		public IRequestPipelineFactory PipelineProvider { get; }
 
 		/// <summary>
-		/// Transport coordinates the client requests over the connection pool nodes and is in charge of falling over on different nodes 
+		/// Transport coordinates the client requests over the connection pool nodes and is in charge of falling over on different nodes
 		/// </summary>
 		/// <param name="configurationValues">The connectionsettings to use for this transport</param>
 		public Transport(TConnectionSettings configurationValues)
@@ -23,7 +23,7 @@ namespace Elasticsearch.Net
 		{ }
 
 		/// <summary>
-		/// Transport coordinates the client requests over the connection pool nodes and is in charge of falling over on different nodes 
+		/// Transport coordinates the client requests over the connection pool nodes and is in charge of falling over on different nodes
 		/// </summary>
 		/// <param name="configurationValues">The connectionsettings to use for this transport</param>
 		/// <param name="pipelineProvider">In charge of create a new pipeline, safe to pass null to use the default</param>
@@ -82,6 +82,11 @@ namespace Elasticsearch.Net
 					{
 						pipeline.MarkDead(node);
 						seenExceptions.Add(pipelineException);
+					}
+					catch (ResolveException resolveException)
+					{
+						resolveException.RethrowKeepingStackTrace();
+						return null; //dead code due to call to RethrowKeepingStackTrace()
 					}
 					catch (Exception killerException)
 					{
@@ -142,6 +147,11 @@ namespace Elasticsearch.Net
 					{
 						pipeline.MarkDead(node);
 						seenExceptions.Add(pipelineException);
+					}
+					catch (ResolveException resolveException)
+					{
+						resolveException.RethrowKeepingStackTrace();
+						return null; //dead code due to call to RethrowKeepingStackTrace()
 					}
 					catch (Exception killerException)
 					{
--- a/src/Nest/CommonAbstractions/Infer/Field/FieldResolver.cs
+++ b/src/Nest/CommonAbstractions/Infer/Field/FieldResolver.cs
@@ -51,17 +51,14 @@ namespace Nest
 		{
 			if (property.IsConditionless()) return null;
 			if (!property.Name.IsNullOrEmpty())
-			{
-				if (property.Name.Contains("."))
-					throw new ArgumentException("Property names cannot contain dots.");
 				return property.Name;
-			}
-			string f;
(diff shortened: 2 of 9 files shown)


src/Elasticsearch.Net/Transport/Transport.cs
```diff
@@ -83,6 +83,10 @@ public ElasticsearchResponse<TReturn> Request<TReturn>(HttpMethod method, string
 						pipeline.MarkDead(node);
 						seenExceptions.Add(pipelineException);
 					}
+					catch (ResolveException resolveException)
+					{
+						resolveException.RethrowKeepingStackTrace();
```
Mpdreamz: add a return for analysis sake, eventhough its never hit

src/Nest/CommonAbstractions/Infer/Field/FieldResolver.cs
```diff
@@ -71,7 +71,7 @@ private string Resolve(Expression expression, MemberInfo member, bool toLastToke
 					: null;
 
 			if (name == null)
-				throw new ArgumentException("Could not resolve a name from the given Expression or MemberInfo.");
+				throw new ResolveException("Could not resolve a name from the given Expression or MemberInfo.");
```
Mpdreamz: Lets include a message that it resolved to null

src/Nest/CommonAbstractions/Infer/IndexName/IndexNameResolver.cs
```diff
@@ -15,26 +16,28 @@ public IndexNameResolver(IConnectionSettingsValues connectionSettings)
 
 		public string Resolve(IndexName i)
 		{
-			if (i == null) return this.Resolve((Type)null);
-			return i.Name ?? this.Resolve(i.Type);
+			if (i == null || string.IsNullOrEmpty(i.Name))
+				return this.Resolve((Type)null);
+			if (char.IsUpper(i.Name[0]))
+				throw new ResolveException($"Index names cannot begin with an uppercase character: {i.Name}");
```
Mpdreamz: I "think" an index name can not have uppercase anywhere, not just the first character

src/Nest/CommonAbstractions/Infer/IndexName/IndexNameResolver.cs
```diff
@@ -15,26 +16,28 @@ public IndexNameResolver(IConnectionSettingsValues connectionSettings)
...
-			return this._connectionSettings.DefaultIndex;
+			if (defaultIndices != null && type != null)
+			{
+				string value;
+				if (defaultIndices.TryGetValue(type, out value) && !string.IsNullOrEmpty(value))
+					indexName = value;
+			}
+			if (string.IsNullOrWhiteSpace(indexName))
+				throw new ResolveException($"Could not resolve an index name for the given type and no default index is set.");
+			if (char.IsUpper(indexName[0]))
```
Mpdreamz: Can we move these lines and the ones around line 19 into its own method seeing they do almost the same thing

src/Tests/ClientConcepts/HighLevel/Inferrence/IndexNameInference.doc.cs
```diff
@@ -0,0 +1,76 @@
...
+
+		[U]
+		public void NoIndexThrowsResolveException()
+		{
+			var settings = new ConnectionSettings();
+			var resolver = new IndexNameResolver(settings);
+			Assert.Throws<ResolveException>(() => resolver.Resolve<Project>());
+		}
+	}
+}
```
Mpdreamz: :heart:

src/Tests/ClientConcepts/HighLevel/Inferrence/IndexNameInference.doc.cs
```diff
@@ -0,0 +1,76 @@
...
+			Assert.Throws<ResolveException>(() => resolver.Resolve<Tag>()); // should infer default index
+			Assert.Throws<ResolveException>(() => resolver.Resolve("Foo"));
+		}
+
+		[U]
+		public void NoIndexThrowsResolveException()
+		{
+			var settings = new ConnectionSettings();
+			var resolver = new IndexNameResolver(settings);
+			Assert.Throws<ResolveException>(() => resolver.Resolve<Project>());
```
Mpdreamz: We should assert that the message makes sense too e.g in this case no default global index and no configured index for project directly were specified on ConnectionSettings

src/Nest/CommonAbstractions/Infer/IndexName/IndexNameResolver.cs
gmarz: Doh, you're right.  I've got it confused with `_`.

src/Nest/CommonAbstractions/Infer/IndexName/IndexNameResolver.cs
gmarz: :+1:

src/Tests/ClientConcepts/HighLevel/Inferrence/IndexNameInference.doc.cs
gmarz: Will do, also adding a test to ensure ResolveException bubbles out.

PR #1933, opened 2016-03-21
Alternative to ..\{project}_Net45 folder side-side hack

Mpdreamz:
that allows to develop project.json (the toolchain that produces our
artifacts) and csproj files (the best interface for continuous tests
through NCrunch) in tandem.

The old separate folder hack introduces strict restrictions on our
csproj based solution, adding files was forbidden because this would
expand the wildcards in the csproj and the file would end up in the
wrong folder.

Normally this side by side loading is riddled with nuget biting into
both worlds but luckily we use paket in the csproj world.

We didn't go for this solution initially because NCrunch while not
supporting project.json, it did have an undocumented check for
project.json files to kick off a different project resolution. Since
NCrunch 2.20 (released yesterday) this "special" check is removed and
NCrunch loads our csproj files that live next to project.json normally
again.

Our default work solution can now be Elasticsearch.sln again, while our
command line build still builds Elasticsearch.DNX.sln and we can use the
latter to investigate CoreCLR bugs/features.


Mpdreamz:
@gmarz @russcam pulling this in to `master` and `2.x` so I'm unblocked for the rest of the day. Please still review :smile:

gmarz:
Awesome!  LGTM :+1:

PR #1934, opened 2016-03-21
Implemented missing cat endpoints

Mpdreamz:
* cat snapshots
* cat nodeattrs
* cat help
* cat repositories

Should backport this to 2.x

Also fixes some code generation path issues now the projects live under the same folder and the output for the classic projects are under `bin\Net45\` instead of `bin`


src/CodeGeneration/CodeGeneration.LowLevelClient/ApiEndpoints/cat.snapshots.json
```diff
@@ -8,6 +8,7 @@
       "parts": {
         "repository": {
           "type" : "list",
+          "required" : true,
```
Mpdreamz: have to send a PR in the main repos for this: https://github.com/elastic/elasticsearch/issues/17216

src/Nest/Cat/CatNodeAttributes/CatNodeAttributesRequest.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+#pragma warning disable 612
+
+namespace Nest
+{
+	[Obsolete()]
```
gmarz: :+1:

gmarz:
LGTM

src/Nest/Cat/CatNodeAttributes/CatNodeAttributesRequest.cs
Mpdreamz: ohh I should probably add a better message there ?

gmarz:
Backported to 2.x via e2c990f382ef7fc3a6e7af323b019c2afa3a708b

PR #1944, opened 2016-03-22
fix #1931 script query was not serializing properly

### LINKED ISSUE #1931: Bug in filter aggregation with script
When using filter aggregation with an inline script as filter nest generates an incorrect query to elasticsearch.
 
The following code: 
` .Filter("NAME", htw => htw.Filter(htwFlt => htwFlt.Script(hwScrpt => hwScrpt.Inline("SCRIPT "))))`

generates the following query:

` "NAME": {
      "filter": {
        "script": {
          "inline": "SCRIPT"
        }
      },`

Elasticsearch responds with **[script] query does not support [inline]**
Changing the query to:

` "NAME": {
      "filter": {
        "script": {
          "script": "SCRIPT"
        }
      },`

produces the correct result.

Mpdreamz:
We were serializing script queries as:

script
  - inline

but it needs to be

script
  - script
    - inline

only under function score did we do the correct thing by not including
the additional script tag


gmarz:
LGTM

gmarz:
Only merging this to 2.x as we have #1945 to fix this using `IScript` in master.

PR #1948, opened 2016-03-23
fix #1930 unwrap .Terms(new List<T>) to just T

### LINKED ISSUE #1930: Bug in TermsQueryDescriptor when passing to Terms IEnumerable<TValue>
if i have some descriptor like
`var descr = new QueryContainerDescriptor<SearchProductCard>().Terms(t => t.Field("field").Terms(new List<int> {1,2}));`

it will transfer in 
```
"terms": {
     "field": [
        [
             "1",
             "2"
         ]
      ]
   }
```
This causes an error in elastic
```
"reason": {
               "type": "number_format_exception",
               "reason": "For input string: \"[\""
            }
```

I have noticed that before commit [e1e467e2](https://github.com/elastic/elasticsearch-net/commit/e1e467e2bc91b1752b48048f71f4c08b8333c787) there was an explicit conversion to array
```
public TermsQueryDescriptor<T, K> Terms(IEnumerable<K> terms) => Assign(a =>  
    {
    if (terms.HasAny())
    terms = terms.Where(t => t != null).ToArray();
    a.Terms = terms.Cast<object>();
});
```
and missed later
```
public TermsQueryDescriptor<T, TValue> Terms(IEnumerable<TValue> terms) => Assign(a => a.Terms = terms.Cast<object>());
```

Mpdreamz:
When calling

```csharp
.Terms(new List<T>());
```

The user expects T[] to be written to elasticsearch but due to C#
resolution they end up calling

```csharp
.Terms<List<T>>(params List<T>)
```

 and we serialize to T[][] which is not a valid terms query.

The fluent Terms() method now tries to preempt these cases and unwraps
to T[] when a single IEnumerable is passed as T, special casing `string`
to be exluded from the unwrapping behavior so we do not unwrap string to
char[]

This PR also introduces `QueryDslIntegrationTestsBase`, all of our QueryDSL tests are unit tests but with this new base class we can start introducing more integration tests for the query dsl over time /cc @gmarz @russcam

### DIFF
--- a/src/Tests/QueryDsl/QueryDslIntegrationTestsBase.cs
+++ b/src/Tests/QueryDsl/QueryDslIntegrationTestsBase.cs
@@ -1,0 +1,45 @@
+﻿using System;
+using System.Collections.Generic;
+using Elasticsearch.Net;
+using FluentAssertions;
+using Nest;
+using Tests.Framework;
+using Tests.Framework.Integration;
+using Tests.Framework.MockData;
+using Xunit;
+
+namespace Tests.QueryDsl
+{
+	[Collection(IntegrationContext.ReadOnly)]
+	public abstract class QueryDslIntegrationTestsBase : ApiIntegrationTestBase<ISearchResponse<Project>, ISearchRequest, SearchDescriptor<Project>, SearchRequest<Project>>
+	{
+		protected QueryDslIntegrationTestsBase(IIntegrationCluster cluster, EndpointUsage usage) : base(cluster, usage) { }
+		protected override LazyResponses ClientUsage() => Calls(
+			fluent: (client, f) => client.Search<Project>(f),
+			fluentAsync: (client, f) => client.SearchAsync<Project>(f),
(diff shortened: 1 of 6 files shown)


src/Tests/QueryDsl/QueryDslIntegrationTestsBase.cs
```diff
@@ -0,0 +1,45 @@
...
+using Nest;
+using Tests.Framework;
+using Tests.Framework.Integration;
+using Tests.Framework.MockData;
+using Xunit;
+
+namespace Tests.QueryDsl
+{
+	[Collection(IntegrationContext.ReadOnly)]
+	public abstract class QueryDslIntegrationTestsBase : ApiIntegrationTestBase<ISearchResponse<Project>, ISearchRequest, SearchDescriptor<Project>, SearchRequest<Project>>
```
gmarz: Awesome!

gmarz:
LGTM

gmarz:
merged to 2.x and master

PR #1950, opened 2016-03-23
Some fluent methods still took string for Field

Mpdreamz:
instead of Field directly, binary breaking change should only be pulled
into master


gmarz:
Nice catch, LGTM :+1:

PR #2014, opened 2016-04-06
Changed BucketsPath from internal to public

jonyadamit:
Would like to be able to access the buckets path defined on SingleBucketsPath.

MultiBucketsPath already exposes this via dictionary.


gmarz:
LGTM :+1:

PR #2017, opened 2016-04-06
Elasticsearch 2.3 new API's feature parity

Mpdreamz:
This commit adds support for the new 2.3 API's:

* tasks_list
* tasks_cancel
* update_by_query
* reindex

The spec changes have also been updated in core upstream.

The Reindex API is called `ReindexOnServer()` to offset it against our
current `Reindex()` implementation which is still very useful.

Both the reindex and update_by_query steal the status code from one of
the failures and return a valid json body that we need to deserialize.
To support this this AllowedStatusCodes now accepts a -1.

`.IsValid` for these API's inspects whether any failures have been
returned or not.

### DIFF
--- a/src/CodeGeneration/CodeGeneration.LowLevelClient/ApiEndpoints/cat.snapshots.json
+++ b/src/CodeGeneration/CodeGeneration.LowLevelClient/ApiEndpoints/cat.snapshots.json
@@ -8,7 +8,7 @@
       "parts": {
         "repository": {
           "type" : "list",
-          "required" : true,
+          "required": true,
           "description": "Name of repository from which to fetch the snapshot information"
         }
       },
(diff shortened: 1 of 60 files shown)


gmarz:
LGTM!

> To support this this AllowedStatusCodes now accepts a -1.

Maybe add a TODO to make this nullable instead in master?

> The Reindex API is called ReindexOnServer() to offset it against our
current Reindex() implementation which is still very useful.

:+1: What are your thoughts for 5.0? I'm thinking we should rename ReindexOnServer to Reindex when we remove the old API.

Mpdreamz:
I quite like the idea of keeping both at least untill task result/state are stored in Elasticsearch somehow.

Our roundtripping reindex might still be a lot easier to use and monitor through the non blocking RX observable we expose. Also remapping objects in C# is still alot easier then using scripts which might not even be enabled on the cluster.

gmarz:
@Mpdreamz [tasks cancel tests](https://github.com/elastic/elasticsearch-net/blob/2.x/src/Tests/Cluster/TaskManagement/TasksCancel/TasksCancelApiTests.cs) are failing in unit test mode.  Looks like a setup issue - `SetupTaskIds` is empty because `OnBeforeCall` isn't being ran.

PR #2021, opened 2016-04-07
Verbatim and strict query fixes

### LINKED ISSUE #1876: Expose IsVerbatim and IsStrict on OIS query objects
See #1872 for details.  

The only way to set these on OIS objects right now is to wrap in a `QueryContainer` then cast to `IQueryContainer`.  They should be exposed directly on the base class so one can simply do:

```csharp
new TermQuery
{
    IsStrict = true,
    //snip
}
```

### LINKED ISSUE #1898: IsVerbatim not working with multiple queries
Following up on #1872. 

IsVerbatim works fine on a single TermQuery but i can't seem to get it to work when combining mutiple queries. 

```
var container = new QueryContainer(new TermQuery
{
 Field = "surname",
 Value = ""
});

container &= new TermQuery
{
 Field = "age",
 Value = "10"
}

((IQueryContainer)container).IsVerbatim = true;
ElasticClient.Search<MyType>(new SearchRequest { Query = container });
```

This results in
```
POST /<index>/<type>/_search
{
   "query": {
      "term": {
         "age": {
            "value": "10"
          }
       }
    }
}
```

While I expect
```
POST /<index>/<type>/_search
{
   "query": {
      "bool": {
         "must": [
         {
            "term": {
               "surname": {
                "value": ""
             }
           }
         },
         {
            "term": {
                "age": {
                "value": "10"
                 }
             }
          }
       ]
    }
  }
}
```

gmarz:
- Add ability to set IsStrict and IsVerbatim on OIS query objects and at
  the query level for the fluent syntax.  Setting at the container level
  for both APIs is a noop.

- Fix IsStrict check for initializer syntax.  Previously, an exception was
  never thrown.

Closes #1898
Closes #1876


src/Nest/QueryDsl/Span/SpanQuery.cs
```diff
@@ -39,6 +39,10 @@ public class SpanQuery : ISpanQuery
 	{
 		string IQuery.Name { get; set; }
 		double? IQuery.Boost { get; set; }
+		bool IQuery.IsWritable { get; }
+		public bool IsVerbatim { get; set; }
+		public bool IsStrict { get; set; }
+		public bool IsWritable { get { return this.IsVerbatim || IsConditionless(this); } }
```
russcam: Should this also be 

```c#
public bool IsWritable { get { return this.IsVerbatim || !IsConditionless(this); } }
```

i.e. _negate_ `IsConditionless`?

russcam:
`IsStrict` and `IsVerbatim` are now a noop on the outer container so the expectation in https://github.com/elastic/elasticsearch-net/issues/1898 will not work.

I've added some tests for this.

Also fixed `SpanQuery` `IsWritable`

russcam:
I would have expected this to pass i.e. `IsVerbatim` to cascade from the `bool` query to the contained clauses, but it would fail currently

```c#
public class BoolVerbatimQueryUsageTests : QueryDslUsageTestsBase
{
	protected override bool SupportsDeserialization => false;

	public BoolVerbatimQueryUsageTests(ReadOnlyCluster cluster, EndpointUsage usage) : base(cluster, usage) { }

	protected override object QueryJson => new
	{
		@bool = new
		{
			must = new object[]
			{
				new
				{
					term = new
					{
						description = new
						{
							value = ""
						}
					}
				},
				new
				{
					term = new
					{
						name = new
						{
							value = "foo"
						}
					}
				}
			}
		}
	};

	protected override QueryContainer QueryInitializer =>
		new BoolQuery
		{
			// cascade to clauses?
			IsVerbatim = true,
			Must = new List<QueryContainer>
				{
					new TermQuery
					{
						Field = "description",
						Value = ""
					},
					new TermQuery
					{
						Field = "name",
						Value = "foo"
					}
				}
		};


	protected override QueryContainer QueryFluent(QueryContainerDescriptor<Project> q) => q
		.Bool(b => b
			// cascade to clauses?
			.Verbatim()
			.Must(qt => qt
				.Term(t => t
					.Field(p => p.Description)
					.Value("")
				), qt => qt
				.Term(t => t
					.Field(p => p.Name)
					.Value("foo")
				)
			)
		);
}
```

Is my thinking incorrect here?

src/Nest/QueryDsl/Span/SpanQuery.cs
```diff
@@ -42,7 +42,7 @@ public class SpanQuery : ISpanQuery
 		bool IQuery.IsWritable { get; }
 		public bool IsVerbatim { get; set; }
 		public bool IsStrict { get; set; }
-		public bool IsWritable { get { return this.IsVerbatim || IsConditionless(this); } }
+		public bool IsWritable => this.IsVerbatim || !IsConditionless(this);
```
gmarz: Nice catch :+1:

gmarz:
> IsStrict and IsVerbatim are now a noop on the outer container so the expectation in #1898 will not work.

That's correct.  IsVerbatim/Strict has to now be set at each desired query to obtain this behavior, which wasn't possible before.

> I would have expected this to pass i.e. IsVerbatim to cascade from the bool query to the contained clauses, but it would fail currently

That would be ideal, but not sure if it's possible for the same reason we can't cascade from QueryContainer.  I could be wrong though...

russcam:
> That would be ideal, but not sure if it's possible for the same reason we can't cascade from QueryContainer. I could be wrong though...

OK, apologies if we've discussed this before :smile: That does feel a little bit unintuitive for compound queries like `bool` where `Verbatim()` and `Strict()` are exposed still  as they're inherited from `QueryBase` and `QueryDescriptorBase<TDescriptor, TInterface>`.

Mpdreamz:
Rebased against 2.x appended some small touch ups, please review @gmarz 

LGTM :+1:

PR #2024, opened 2016-04-11
Update code generator so that we can check in XPACK specs

Mpdreamz:
Code generation can now download multiple sources and deal with fixed checked in spec folders too (e.g the xpack related specification files). 

The generator currently ignores the xpack specifications but we can enable them as we add support for graph/shield/watcher/license.

Also needs forward merge to master but the xpack spec needs to be updated post merge


russcam:
Just a couple of small tweaks for the DNX solution :smile: 

Other than that, LGTM :+1:

russcam:
Merged in 59af777e35c30ea6d3b2b78c0a76f98e2daddd3e

PR #2026, opened 2016-04-12
Add not_analyzed to NonStringIndexOption and make it the default option

### LINKED ISSUE #1979: NonStringIndexOption should not default to No
We default the nullable argument for `Index()` on our property descriptors to `No` which doesn't make sense (for instance [here](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Mapping/Types/Core/Date/DateProperty.cs#L68) and [here](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Mapping/Types/Core/Number/NumberProperty.cs#L71)).

The default in ES is to index, so calling `Index()` without passing `No` does the opposite of what you expect.

We should deprecate this in 2.x and remove it the method completely in 5.x.  I think something along the lines of `NotIndexed()` (can't think of a better name at the moment) makes more sense, similar to how we have `NotAnalyzed()`.

gmarz:
This PR is questionable because it changes the default behavior of `Index()`, which previously defaulted to `no`, but now will default to `not_analyzed`.

@Mpdreamz @russcam thoughts?

Closes #1979

### DIFF
--- a/src/Nest/Mapping/NonStringIndexOption.cs
+++ b/src/Nest/Mapping/NonStringIndexOption.cs
@@ -7,6 +7,8 @@ namespace Nest
 	[JsonConverter(typeof(StringEnumConverter))]
 	public enum NonStringIndexOption
 	{
+		[EnumMember(Value = "not_analyzed")]
+		NotAnalyzed,
 		[EnumMember(Value = "no")]
 		No
 	}
--- a/src/Nest/Mapping/Types/Core/Boolean/BooleanProperty.cs
+++ b/src/Nest/Mapping/Types/Core/Boolean/BooleanProperty.cs
@@ -19,7 +19,7 @@ namespace Nest
 		[JsonProperty("fielddata")]
 		INumericFielddata Fielddata { get; set; }
 	}
-	
+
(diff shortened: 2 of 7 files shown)


russcam:
It **is** changing behaviour, albeit correcting an original issue :smile: 

The change is from 

- `no` -
Do not add this field value to the index. With this setting, the field will not be queryable.

to 

- `not_analyzed` -
Add the field value to the index unchanged, as a single term. This is the default for all fields that support this option except for `string` fields. `not_analyzed` fields are usually used with term-level queries for structured search.

So consumers won't be losing functionality by the default change but could potentially be storing more data than they did before. Not a bad thing per se.

I think this is a reasonable behavioural change that [implements the right default according to the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-index.html#mapping-index) so LGTM. We just need to ensure that we highlight it particularly in the release notes.

It'd also be worth adding the descriptions to the XML comments for the `NonStringIndexOption` enum

Mpdreamz:
LGTM :+1: in 5.0 we need to not have any argument defaults for this one.

PR #2028, opened 2016-04-12
The task cancel API test was a bit flakey,

Mpdreamz:
It was setting additional values in the integration setup on the class
itself but this setup is run once for possibly multiple classes e.g

EndpointUsage x;
new TestClass(x).Test1();
new TestClass(x).Test2();
new TestClass(x).Test3();
new TestClass(x).Test4();
new TestClass(x).Test5();

integration setup is an override per TestClass but only one of them will
be called per instance of `x`. So if the callback defines setting
variables on the `TestClass` instance using `this` other TestClass
instances won't see this.

This PR makes CallIsolatedValues a class of its own that allows you to
register extended values that can be shared amongst all TestClass
instances.


src/Tests/Cluster/TaskManagement/TasksCancel/TasksCancelApiTests.cs
```diff
@@ -10,29 +10,31 @@
...
 			public long Id { get; set; }
 			public string Flag { get; set; }
 		}
 
 		public TasksCancelApiTests(OwnIndexCluster cluster, EndpointUsage usage) : base(cluster, usage) { }
 
-		protected IDictionary<string, TaskId> SetupTaskIds { get; } = new Dictionary<string, TaskId>();
-
-		protected override void BeforeAllCalls(IElasticClient client, IDictionary<ClientMethod, string> values)
+		protected override void IntegrationSetup(IElasticClient client, CallUniqueValues values)
```
gmarz: I like the rename.  Makes it clear that this is only applies to integration tests.

src/Tests/Cluster/TaskManagement/TasksCancel/TasksCancelApiTests.cs
```diff
@@ -10,29 +10,31 @@
 
 namespace Tests.Cluster.TaskManagement.TasksCancel
 {
-	// TODO Unit tests will fail in mixed mode because SetupTaskIds isn't setup
 	[Collection(IntegrationContext.OwnIndex)]
 	public class TasksCancelApiTests : ApiIntegrationTestBase<ITasksCancelResponse, ITasksCancelRequest, TasksCancelDescriptor, TasksCancelRequest>
 	{
-		public class Test
+		private TaskId TaskId => this.RanIntegrationSetup ? this.ExtendedValue<TaskId>("taskId") : "foo:1";
```
gmarz: :+1:

gmarz:
LGTM!

PR #2034, opened 2016-04-14
Add support for _license related API's

Mpdreamz:
- License.LoadFromDisk(path)
- Backported elasticsearch node improvements from master branch
- Added bat file generations for the various clusters, allows you to
  manually start one of our test clusters with the right settings
- Make sure tests that inherit ApiTestBase directly do not use a "real"
  client in `mixed` testing mode
- Validate license information on startup, if this fails all tests
  fails. From the commandline we always start a fresh cluster so the
  trial license'll stay valid. If you manually start a node with no
  license but you have `ES_LICENSE_FILE` pointing to a license file we will
  try and register that.


src/Tests/Framework/Integration/Process/ElasticsearchNode.cs
```diff
@@ -128,89 +145,131 @@ string prefix
...
+
+		private void ValidateLicense()
+		{
+			var client = TestClient.GetClient();
+			var license = client.GetLicense();
+			if (license.IsValid && license.License.Status == LicenseStatus.Active) return;
+
+			var exceptionMessageStart = "Server has license plugin installed, ";
+			var licenseFile = Environment.GetEnvironmentVariable("ES_LICENSE_FILE", EnvironmentVariableTarget.Machine);
+			if (!string.IsNullOrWhiteSpace(licenseFile))
```
gmarz: Should we also check for existence of the license here and continue if it isn't there?

gmarz:
One minore comment, otherwise LGTM :+1:

gmarz:
Actually, nevermind my comment.  I think it makes sense to throw the FileNotFound exception if ES_LICENSE_FILE is set.

Mpdreamz:
Manually cherry-picked to 2.x and master

Issue #1997, opened 2016-04-05
Changes to queries


gmarz:
https://www.elastic.co/guide/en/elasticsearch/reference/master/breaking_50_search_changes.html#_changes_to_queries

Any properties that are removed should be marked obsolete in 2.x (#1988).

PR #2045, opened 2016-04-19
Deprecate fuzziness ratio

russcam:
See https://github.com/elastic/elasticsearch-net/issues/1997

It looks like this was deprecated in 1.7 and removed in 2.0 - https://www.elastic.co/guide/en/elasticsearch/reference/1.7/common-options.html#fuzziness
Since it still exists in NEST 2.x, I've marked as deprecated here - will no-op in NEST 2.x


Mpdreamz:
LGTM, +1 for backporting as far back as `1.x`

Issue #2007, opened 2016-04-05
Percolator changes


gmarz:
https://www.elastic.co/guide/en/elasticsearch/reference/master/breaking_50_percolator.html#breaking_50_percolator

Issue #2059, opened 2016-04-25
Re-evaluate Property Type Hierarchy


Mpdreamz:
`text` fields do not support docvalues just as [analyzed string fields](https://github.com/elastic/elasticsearch/issues/10061) did not in `2.x` but now that we have an explicit split of string to `text` and `keyword` this leaves `DocValues()` exposed on our `keyword` mapping.

gmarz:
>  this leaves DocValues() exposed on our `keyword` mapping

I think you mean `text` mapping?

Mpdreamz:
Doh! yeah

russcam:
Started moving `IProperty` properties around to accommodate the `percolator` type in https://github.com/elastic/elasticsearch-net/pull/2063. Be a good opportunity to refactor this further

Mpdreamz:
Comment from @gmarz on PR #2063 

>Few minor nitpicks but overall LGTM :+1: . I'm +1 on ICoreProperty, but I think there are a few instances where it exposes properties that aren't valid for a type (e.g. TextProperty.DocValues). There could be other instances as well. We should re-evaluate all of our types as part of #2059, but I'm cool with this for alpha1.

We should do a full sweep on our mapping API to be make sure its not introducing bad abstract base classes.

gmarz:
Closed via #2337

PR #2063, opened 2016-04-27
Feature/percolate query

russcam:
Closes https://github.com/elastic/elasticsearch-net/issues/2007

@Mpdreamz, I see that you already fixed the percolate stats changes in c5b71db6472c49faa60a8f5f1a5a4c88692b055d  :+1: 

@gmarz, @Mpdreamz , What are your thoughts on `ICoreProperty`? Basically, the `percolator` type only needs name, type name and index name so these remain on `IProperty` and pushed the remaining properties up into `ICoreProperty`. 

Maybe `IAnalyzableProperty` might be better, as the properties relate to analysis?


src/Nest/CommonAbstractions/Infer/Id/IdExtensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Nest
+{
+    public static class IdExtensions
+    {
+		public static bool IsConditionless(this Id id)
```
gmarz: Should we make these extensions internal?  Not just this one in particular, but any `IsCondtionless()` extension (i.e. `TypeName.IsConditionless()`).

src/Nest/Mapping/Types/Core/Text/TextProperty.cs
```diff
@@ -8,7 +8,7 @@
 namespace Nest
 {
 	[JsonObject(MemberSerialization.OptIn)]
-	public interface ITextProperty : IProperty
+	public interface ITextProperty : ICoreProperty
```
gmarz: `ICoreProperty` has `DocValues` which isn't valid for text fields (see #2059).

gmarz:
Few minor nitpicks but overall LGTM :+1: .  I'm +1 on `ICoreProperty`, but I think there are a few instances where it exposes properties that aren't valid for a type (e.g. `TextProperty.DocValues`).  There could be other instances as well.  We should re-evaluate all of our types as part of #2059, but I'm cool with this for alpha1.

src/Nest/CommonAbstractions/Infer/Id/IdExtensions.cs
russcam: ++

src/Nest/Mapping/Types/Core/Text/TextProperty.cs
russcam: Will split DocValues out into a separate interface, and re-evaluate others as part of #2059

PR #2068, opened 2016-05-02
Mega refactor of test framework bits,

Mpdreamz:
This all started to support

```yaml
version: latest
```

So we can start working on `alpha2` without hardcoding` 5.0.0-alpha2-SNAPSHOT` on CI.

In `tests.yaml` to automatically download and run the latest
elasticsearch snapshot. Plugins do not quite install nicely on
elasticsearch snapshots so I needed to disable it if we are running
against a snapshot which led to a shit ton of refactoring down a very
slippery slope.

- Internalize semanticversioning so we do not have to special case
  CoreFX
- ElasticsearchVersion extends SemanticVersion (renamed SemVer.Version)
- rename `do_not_spawn` to `test_against_already_running_elasticsearch`
- Send ContentType as well as Accept (needed for alpha2) @russcam should
  we just use one config setting to send both headers?
- `[Collection(IntegrationContext.ReadOnly)]` is now
  `[Collection(TypeOfCluster.ReadOnly)]` which i think is more
   descriptive
- Split all the filesystem manipulation from `ElasticsearchNode` into
  `INodeFileSystem` of which we now have one implementation
  `TestRunnerFileSystem` but could be used for e.g
  `ProductionFileSystem` if we are intend on shipping this as part of the
  installer.
- Regrouped everything under `Framework` folder into a (hopefully) more
  logical structure
- Tried to minimize the various different ways clients are created.
- UrlTester and ApiTestBase now validate expected urls the same way
- We now deal with elasticsearch start exceptions happening before starting (e.g using bad
  settings) way better and faster, they will surfice as their own
  exception rather then a WaitHandle exception.
- ObservableProcess now spits out `ConsoleOut` instead of `string` and
  `ElasticsearchMessage` extends `ConsoleOut` and is renamed to
  `ElasticsearchConsoleOut`
- Plugins we support are now registered in the static
  `ElasticsearchPlugins`.
- `ClusterBase` definitions or can now be marked with
  `RequiresPluginAttribute` which dictate which plugins MUST be
   available.
- Plugin validation is disabled against SNAPSHOT versions and any
  test part of Cluster Collection with a `RequiresPluginAttribute` are
  automatically skipped.
- Tests now start and run OK when running against alpha2
- Tons of tiny fixes

Tested:

> build
> build integrate 5.0.0-alpha1
> build integrate 5.0.0-alpha2-SNAPSHOT
> build integrate latest

The latter still has 27 failures but atleast the clusters start properly
and plugin tests are ignored!


src/Tests/Framework/EndpointTests/UrlTests.cs
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		private UrlTester Assert(string typeOfCall, IApiCallDetails callDetails)
+		{
+			callDetails.HttpMethod.Should().Be(this.ExpectedHttpMethod, typeOfCall);
+			ComparePathAndQuerstring(this.ExpectedUrl, callDetails.Uri);
+			return this;
+		}
+
+		public static void ComparePathAndQuerstring(string expectedPathAndUri, Uri actualUri)
```
gmarz: Should be ComparePathAndQuer**yS**tring

src/Tests/Framework/ManagedElasticsearch/FileSystem/TestRunnerFileSystem.cs
```diff
@@ -0,0 +1,242 @@
...
+			}
+
+			if (Directory.Exists(this.RepositoryPath))
+			{
+				Console.WriteLine("attempting to delete repositories");
+				Directory.Delete(this.RepositoryPath, true);
+			}
+		}
+	}
+}
```
gmarz: :100: So much cleaner :)

gmarz:
Tested locally and looks good :+1:  I'm getting 27 failures as well when running `build integrate latest`.

Awesome work @Mpdreamz !

russcam:
Re: `ContentType` and `Accept` headers, I was thinking of keeping them separate for two reasons

1. It's obvious which property controls which header
2. There may be situations where a user may want to send in one media/mime type but accept the response in another.

Awesome work on this :+1:

Issue #2072, opened 2016-05-03
Expose TypeName in IGetMappingResponse


russcam:
[The current implementation of `IGetMappingResponse`](https://github.com/elastic/elasticsearch-net/blob/9cab2174dd988fb8cf6b440fdd24d32517eab5b0/src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs#L12) returns a dictionary of index string name and list of `TypeMapping` for the index, but does not expose the `TypeName` to which the `TypeMapping` relates to; so if the request asks for multiple types within a given index, there is no way of correlating the `TypeMapping` back to the type name.

We should expose the `TypeName` along with the `TypeMapping`. Changing the `Mappings` property in 2.x would be a breaking change so we should include another property for this. Maybe

```c#
public interface IGetMappingResponse : IResponse
{
    // .... other properties

    Dictionary<IndexName, IDictionary<TypeName, TypeMapping>> IndexTypeMappings { get; }
}
```

[This came up on stackoverflow](http://stackoverflow.com/questions/36991184/get-elasticsearch-type-mapping-names-with-nest-client#36991184)

gmarz:
@russcam a little confused on the status of this.  #2074 was merged into 2.x and made it into 2.3.1, but #2074 which is targeted at master is still open and in limbo.

Is the change complete for the 2.x branch?  I'm going to remove the 2.3.2 label for now because either the change is done and this should be labeled 2.3.1, or it needs more commits and in that case let's target 2.3.3.

russcam:
#2074 was merged into `2.x` [after the release of 2.3.1](https://github.com/elastic/elasticsearch-net/commits/2.x) so that change will have gone into 2.3.2. The change is complete for 2.x.

Where the `master` change in #2073 is slightly different is that it would be great to return `IndexName` instead of the string for the index name and similarly, `TypeName` instead of a string for the type name. There are a few ways I'd like to explore to see if this would work so have kept #2073 open :smile:

PR #2073, opened 2016-05-03
Map type names along with mappings

russcam:
Closes https://github.com/elastic/elasticsearch-net/issues/2072

### DIFF
--- a/src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs
+++ b/src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs
@@ -1,16 +1,14 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
-using Elasticsearch.Net;
 
 namespace Nest
 {
 	public interface IGetMappingResponse : IResponse
 	{
-		/// <summary>
-		/// TODO dict|indexname, imappings|
-		/// </summary>
-		Dictionary<string, IList<TypeMapping>> Mappings { get; }
+		Dictionary<string, IDictionary<string, TypeMapping>> Mappings { get; }
+
 		TypeMapping Mapping { get; }
+
 		void Accept(IMappingVisitor visitor);
 	}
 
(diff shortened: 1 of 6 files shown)


src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs
```diff
@@ -20,39 +18,30 @@ internal class GetRootObjectMappingWrapping : Dictionary<string, Dictionary<stri
...
-				.SelectMany(kv=>kv.Value)
-				.FirstOrDefault(t=>t != null);
+
+			this.Mapping = this.Mappings.Where(kv => kv.Value.HasAny(v => v.Value != null))
+				.SelectMany(kv => kv.Value)
+				.FirstOrDefault(t => t.Value != null).Value;
 		}
 
-		public Dictionary<string, IList<TypeMapping>> Mappings { get; internal set; } = new Dictionary<string, IList<TypeMapping>>();
+		public Dictionary<IndexName, IDictionary<TypeName, TypeMapping>> Mappings { get; internal set; } = new Dictionary<IndexName, IDictionary<TypeName, TypeMapping>>();
```
Mpdreamz: if we support `IndexName` and `TypeName` on responses we also have to support the following use case:

```csharp
var index = Index<Project>();
var mapping = this.client.GetMapping(new GetMappingRequest(index, AllTypes));
var indexMappings = mapping.Mappings[index];
```

Since the response `Index` is build from `string` and the request `index` is an unresolved `IndexName` it won't be found where you would totally expect it to.

I would love to support this but in 2.x we gave up because we would have to inject ConnectionSettings as state of all responses. 

Ideas here welcome! :smile:

russcam: I was thinking about that :smile: 

The point at which we deserialize the response, we have access to the `ConnectionSettings` on the client and the request; we could use this information to construct the correct `TypeName` for the given index, type mapping and request, without needing to _inject_ the `ConnectionSettings` into the response.

Mpdreamz:
@elastic/microsoft I marked this as research, I would love to support `IndexName` and `TypeName` throughout all responses but we need a way to do this so that unresolved typenames work too e.g


```csharp
IndexName r = typeof(X);
var reponse = client.Call(c=>c.Index(r));
response.Indices[r].......
```

`r` is an unresolved IndexName with no string representation, whereas `Indices` will hold resolved indexnames without type information.

russcam:
I'm going to get this change into `master` _without_ the `IndexName` and `TypeName` support for the moment as it is useful, then can look at ways to support them afterwards, if possible :smile:

PR #2074, opened 2016-05-03
Map type names along with mappings

russcam:
Closes https://github.com/elastic/elasticsearch-net/issues/2072

### DIFF
--- a/src/Tests/Indices/MappingManagement/GetMapping/GetMappingApiTest.cs
+++ b/src/Tests/Indices/MappingManagement/GetMapping/GetMappingApiTest.cs
@@ -38,8 +38,6 @@ namespace Tests.Indices.MappingManagement.GetMapping
 
 		protected override void ExpectResponse(IGetMappingResponse response)
 		{
-			response.IsValid.Should().BeTrue();
-
 			var visitor = new TestVisitor();
 			response.Accept(visitor);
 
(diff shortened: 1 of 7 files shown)


src/Tests/Indices/MappingManagement/GetMapping/GetMappingApiTest.cs
```diff
@@ -38,8 +38,6 @@ public class GetMappingApiTests : ApiIntegrationTestBase<IGetMappingResponse, IG
 
 		protected override void ExpectResponse(IGetMappingResponse response)
 		{
-			response.IsValid.Should().BeTrue();
```
Mpdreamz: Why are we no longer expecting a valid response?

Mpdreamz:
LGTM other then a small comment :+1: +1 also on getting rid of the todo to validate empty responses.

src/Tests/Indices/MappingManagement/GetMapping/GetMappingApiTest.cs
russcam: Looking at  [`ApiIntegrationTestBase`](https://github.com/elastic/elasticsearch-net/blob/2.x/src/Tests/Framework/ApiIntegrationTestBase.cs#L29), there is a check for a valid response in

```c#
[I] protected async Task ReturnsExpectedIsValid() =>
    await this.AssertOnAllResponses(r=>r.IsValid.Should().Be(this.ExpectIsValid));
```

so it seemed superfluous to check it again in `ExpectResponse` :smiley: 

I think we could remove the intermediate `GetRootObjectMappingWrapping` altogether and deserialize to `GetMappingResponse`:

1. if there are no indices, the `_mapping` endpoint returns `{}`
2. if there is a index with no mappings then `{index}/_mapping` returns `{ "index" : { "mappings": {} } }`

PR #2078, opened 2016-05-04
fix #1957 skip http disabled nodes from sniff

### LINKED ISSUE #1957: Issue with SniffingConnectionPool and non-http nodes
We ran into an issue when using the SniffingConnectionPool and having a node in the cluster with http disabled. This gives an exception saying "Failed sniffing cluster state". Looking at the inner exception it turns out that the actual error occurs in `ParseToUri() ` in `SniffResponse` which executes a regex getting an empty httpAddress passed in. 

The issue does not occur when stopping non-http nodes or when using the StaticConnectionPool and explicitly specify the nodes. 

We are using NEST 2.0.4 and ES 2.2.0.

Mpdreamz:


gmarz:
LGTM :+1:

russcam:
LGTM too :+1:

gmarz:
Merged to 2.x and master

PR #2081, opened 2016-05-06
Leverage metadata for deserializing aggregation responses

gmarz:
Continuation of  #1806 

CLR types are no longer burned in the metadata.  Instead, the name of the aggregation is used and a map is maintained in memory for looking up the aggregate type to deserialize to.  This adds some minor overheard, but the request is a bit cleaner:

```
"my_terms" : {
    "meta": {
        "_type": "terms"
    },
    "terms": {
        "field": "my_field"
    }
}
```

versus

```
"my_terms" : {
    "meta": {
        "_type": "Nest.TermsAggregate"
    },
    "terms": {
        "field": "my_field"
    }
}
```

Other notable changes:
- The filters aggregation has been broken up into two separate request objects: [NamedFiltersAggregation](https://github.com/elastic/elasticsearch-net/blob/feature/aggs-metadata/src/Nest/Aggregations/Bucket/Filters/NamedFiltersAggregation.cs) and [AnonymousFiltersAggregation](https://github.com/elastic/elasticsearch-net/blob/feature/aggs-metadata/src/Nest/Aggregations/Bucket/Filters/AnonymousFiltersAggregation.cs), since their responses are completely different.
- Introduced [ILazyDeserialize]() for objects that need the serializer late in the game (similar to [ILazyDocument](https://github.com/elastic/elasticsearch-net/blob/feature/aggs-metadata/src/Nest/CommonAbstractions/LazyDocument/LazyDocument.cs)).  [TopHits](https://github.com/elastic/elasticsearch-net/blob/feature/aggs-metadata/src/Nest/Aggregations/Metric/TopHits/TopHitsAggregate.cs) uses this...it felt better to keep this generic rather than implementing it specifically for [TopHits in the converter](https://github.com/elastic/elasticsearch-net/blob/feature/aggs-metadata/src/Nest/Aggregations/AggregateJsonConverter.cs#L41).

Next, I'd like to convert `AggregationsHelper` to a dictionary and replace `Aggregations` and `Aggs` on the response with a single object.  But that's a separate change from this PR and can be done later.


src/Nest/Aggregations/AggregateJsonConverter.cs
```diff
@@ -10,580 +10,102 @@ namespace Nest
...
 
-			if (reader.TokenType == JsonToken.EndObject)
-				return statsMetric;
+			var meta = metaProperty.Value.ToObject<Dictionary<string, object>>();
+			if (!meta.ContainsKey(AggregationMetadata.Key))
+				throw new Exception($"Cannot deserialization aggregation result. Metadata is missing {AggregationMetadata.Key}.");
 
-			return GetExtendedStatsAggregate(statsMetric, reader);
-		}
+			var type = meta[AggregationMetadata.Key] as string;
```
russcam: nitpick: think we can just do a straight cast here

src/Nest/Aggregations/AggregationMetadata.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Reflection;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace Nest
+{
+	internal static class AggregationMetadata
+	{
+		public static readonly string Key = "_type";
```
russcam: can be a `const`?

src/Nest/Aggregations/AggregationMetadata.cs
```diff
@@ -0,0 +1,71 @@
...
+			{
+				var value = GetElasticsearchName(i);
+				if (Map.ContainsKey(value))
+					return value;
+			}
+			return null;
+		}
+
+		private static Dictionary<string, Type> _map;
+		public static Dictionary<string, Type> Map
```
russcam: This property isn't thread-safe; the map initialization could be done in a static constructor.

I think it might make sense to also return a `ReadOnlyDictionary<string, Type>`?

src/Nest/Aggregations/AggregateJsonConverter.cs
```diff
@@ -10,580 +10,102 @@ namespace Nest
...
-			var extendedStatsMetric = new ExtendedStatsAggregate()
-			{
-				Average = statsMetric.Average,
-				Count = statsMetric.Count,
-				Max = statsMetric.Max,
-				Min = statsMetric.Min,
-				Sum = statsMetric.Sum
-			};
+			var aggregate = jObject.ToObject(AggregationMetadata.Map[type]) as IAggregate;
+			aggregate.Meta = meta.HasAny() ? meta : null;
```
russcam: nitpick: looks like `meta` can be assigned as is here

src/Nest/Aggregations/AggregateJsonConverter.cs
```diff
@@ -10,580 +10,102 @@ namespace Nest
...
-				Value = (reader.Value as double?)
-			};
-			if (valueMetric.Value == null && reader.ValueType == typeof(long))
-				valueMetric.Value = reader.Value as long?;
-
-			if (valueMetric.Value != null)
-			{
-				reader.Read();
-				if (reader.TokenType != JsonToken.EndObject)
+				var bucketsProperty = jObject.Properties().Where(p => p.Name == "buckets").FirstOrDefault();
```
russcam: Can shorten to `.FirstOrDefault(predicate)`

src/Nest/Aggregations/AggregateJsonConverter.cs
```diff
@@ -10,580 +10,102 @@ namespace Nest
...
-			double? toDouble = null, fromDouble = null;
-
-			var readExpectedProperty = true;
-			while (readExpectedProperty)
+			var aggregates = new Dictionary<string, IAggregate>();
+			foreach (var property in properties.Where(p => p.Value.Type == JTokenType.Object))
 			{
-				switch (reader.Value as string)
+				var jObject = property.Value.ToObject<JObject>();
+				var meta = jObject.Properties().Where(p => p.Name == "meta").FirstOrDefault();
```
russcam: Can shorten to `.FirstOrDefault(predicate)`

src/Nest/Aggregations/AggregationContainer.cs
```diff
@@ -287,6 +290,15 @@ public void Accept(IAggregationVisitor visitor)
 		{
 			if (visitor.Scope == AggregationVisitorScope.Unknown) visitor.Scope = AggregationVisitorScope.Aggregation;
 			new AggregationWalker().Walk(this, visitor);
+		}
+
+		internal static void InjectMetadata(IAggregation aggregation)
+		{
+			if (aggregation.Meta == null)
+				aggregation.Meta = new Dictionary<string, object>();
+			if (aggregation.Meta.ContainsKey(AggregationMetadata.Key))
```
russcam: Should a user be prevented from adding that key to metadata to begin i.e. introduce our own metadata dictionary that encapsulates this rule?

src/Nest/CommonAbstractions/SerializationBehavior/LazyDeserialize.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Nest
+{
+	public interface ILazyDeserialize
+	{
+		JsonSerializer Serializer { get; set; }
```
russcam: Do we want to couple (de)serialization so tightly with Json.NET?

russcam:
A few comments, mostly nitpick :smile: 

The part I have a concern about though is the `ILazyDeserialize`. It looks like top hits is the only aggregation that needs this. I don't think we should be exposing `JsonSerializer` on a public property interface, or accept it as a parameter to a public method.

Could we keep a map of `Type` -> `string` on the client to perform the deserialization to the correct type at normal deserialization time? We could then send meta with the top hits aggregation that includes the type to deserialize to, similar to aggregation meta. Happy to add this to the PR?

Looks like the serialization tests need the meta added to `ExpectJson` as well.

This PR will greatly simplify stuff :+1:

src/Nest/Aggregations/AggregateJsonConverter.cs
gmarz: :+1:

src/Nest/Aggregations/AggregationMetadata.cs
gmarz: Yup :+1:

src/Nest/Aggregations/AggregationMetadata.cs
gmarz: :+1:

src/Nest/Aggregations/AggregateJsonConverter.cs
gmarz: No, because if our injected `_type` was the only value in the metadata object, then after we remove it will be assigning `Meta` to an empty dictionary, where it should be null instead.

src/Nest/Aggregations/AggregateJsonConverter.cs
gmarz: :+1:

src/Nest/Aggregations/AggregateJsonConverter.cs
gmarz: :+1:

src/Nest/Aggregations/AggregationContainer.cs
gmarz: Yes...I use to check for that but must have accidentally removed it during refactoring.  Will add it back :+1:.

src/Nest/CommonAbstractions/SerializationBehavior/LazyDeserialize.cs
gmarz: Eh, yea I don't like the `ILazyDeserialize` at all.  I'm going to remove it and treat top hits as a special case in the converter instead.

gmarz:
Thanks for the review @russcam.  I've made most of the suggested changes.

I'm not really sure about using a formatted (ES-like) name as the metadata value anymore.  I'm not sure what it's really buying us.  I've reverted back to using the Nest aggregate type name instead in https://github.com/elastic/elasticsearch-net/pull/2081/commits/69e70b72ce8455f4913377a4d31b6c9b40541493 for comparison.  I think it's overall more elegant, and eliminates the need to maintain a dictionary as well as the use of regular expressions to format the name.

Would love your thoughts @Mpdreamz @russcam.

src/Nest/Aggregations/AggregationContainer.cs
```diff
@@ -287,6 +290,29 @@ public void Accept(IAggregationVisitor visitor)
 		{
 			if (visitor.Scope == AggregationVisitorScope.Unknown) visitor.Scope = AggregationVisitorScope.Aggregation;
 			new AggregationWalker().Walk(this, visitor);
+		}
+
+		internal static ConcurrentDictionary<Type, Type> _aggregateNameCache = new ConcurrentDictionary<Type, Type>();
+		internal static void InjectMetadata(IAggregation aggregation)
+		{
+			aggregation.Meta = aggregation.Meta ?? new Dictionary<string, object>();
```
Mpdreamz: instead of new()'ing a a dictionary here we could keep an default dict per aggregation type.

src/Nest/Aggregations/Metric/TopHits/TopHitsAggregation.cs
```diff
@@ -6,6 +6,7 @@ namespace Nest
 {
 	[JsonObject(MemberSerialization = MemberSerialization.OptIn)]
 	[ContractJsonConverter(typeof(AggregationJsonConverter<TopHitsAggregation>))]
+	[AggregateType(typeof(TopHitsAggregate))]
```
Mpdreamz: I think we need to get rid of the json attributes now we have `AggregateType`. 

Maybe 


```
[AggregationJson(typeof(AggregationJson<TopHitsAggregation, TopHitsAggregate>))]
````

File: src/Nest/Aggregations/AggregationContainer.cs [hunk 11]
```diff
@@ -287,6 +290,29 @@ public void Accept(IAggregationVisitor visitor)
...
+				var attribute = aggregationType.GetTypeInfo().GetInterfaces()
+					.Where(t => typeof(IAggregation).IsAssignableFrom(t))
+					.SelectMany(t => t.GetCustomAttributes(typeof(AggregateTypeAttribute)))
+					.FirstOrDefault() as AggregateTypeAttribute;
+				if (attribute == null)
+					throw new Exception($"{aggregationType} is missing the required AggregateType attribute.");
+				aggregateType = attribute.Type;
+				_aggregateNameCache.TryAdd(aggregationType, aggregateType);
+			}
+			aggregation.Meta.Add(typeKey, aggregateType.FullName);
```
Mpdreamz: I still think it would be nicer to see:

```
"_type" : "extended_stats"
```

in the output json then 

```
"_type" : "Nest.ExtendedStatsAggration"
```

or 

```
"_type" : "Nest.ValueAggregate"
```

I think the first can proof more useful to a user as well.

src/Nest/Aggregations/Metric/TopHits/TopHitsAggregation.cs
Mpdreamz: Would also be something we could write Codestandards test for

Mpdreamz:
:heart: this, exited to get this in ! :smile:

Left some comments.

gmarz:
@Mpdreamz @russcam 

OK, back to using the ES aggregation name with my last commit, but with what I think is a much better solution.  Aggregation types now must override `TypeName` and return their name which is used to default the meta dictionary: `{ "_type", TypeName }`.

This simplifies a lot of the code since now we don't need the `AggregateTypeAttribute`, no regular expressions, and the map from aggregation => aggregate is kept locally to `AggregateJsonConverter`.  We also don't need to explicitly inject the meta value in `AggregationContainer`.

I've also fixed all of the unit tests and removed `AggregationJsonConverter` since it wasn't any different than `ReadAsTypeJsonConverter`.

src/Nest/Aggregations/AggregateJsonConverter.cs
```diff
@@ -10,580 +10,147 @@ namespace Nest
 {
 	internal class AggregateJsonConverter : JsonConverter
 	{
-		private static readonly Regex _numeric = new Regex(@"^[\d.]+(\.[\d.]+)?$");
-
-		public override bool CanWrite => false;
-
-		public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer) => this.ReadAggregate(reader, serializer);
+		private static Dictionary<string, Type> _aggregateTypes = new Dictionary<string, Type>
```
russcam: Missing `percentile_ranks` and `percentiles_bucket` aggregations

src/Nest/Aggregations/Metric/TopHits/TopHitsAggregate.cs
```diff
@@ -7,38 +8,37 @@ namespace Nest
...
+	internal class TopHitsProxy
+	{
+		[JsonProperty("total")]
+		public long Total { get; set; }
+
+		[JsonProperty("max_score")]
+		public double? MaxScore { get; set; }
+
+		[JsonProperty("hits")]
+		public IEnumerable<JObject> Hits { get; set; }
```
russcam: Did you get a chance to look at sending the hits type with the `top_hits` aggregation?

gmarz: I did, but hits could be of multiple types so not sure how that could work?

src/Nest/Aggregations/AggregateJsonConverter.cs
gmarz: Nice catch!

src/Nest/Aggregations/Metric/TopHits/TopHitsAggregate.cs
russcam: top hits is strongly typed to `T`; if the results could be covariant, should we expose (similar to search) and handle this at deserialization, and still return `Hits<T>` where `T` is the request type `T`?

russcam: Keeping a reference to the serializer internal to the `TopHitsAggregate` is a fair compromise; I think the piece I'm not keen on is accepting a `JsonSerializer` to the public `Hits<T>()` and `Documents<T>()`. I understand that we expose elements of the serializer at work in other places :smile: Could it at least be an `IElasticsearchSerializer` and expose the hits in the proxy as nested dictionaries / `dynamic` / `ExpandoObject` (i.e. a non Json.Net type)?

gmarz: Maybe we just shouldn't expose `JsonSerializer` through `Hits<T>()` and `Documents<T>()`.  [ILazyDocument](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/CommonAbstractions/LazyDocument/LazyDocument.cs) doesn't...

src/Nest/Aggregations/AggregateJsonConverter.cs
Mpdreamz: Can we write a CodeStandards tests that reflects this dict for all know aggregation types?

gmarz: [Done](https://github.com/elastic/elasticsearch-net/blob/feature/aggs-metadata/src/Tests/CodeStandards/Aggregations.doc.cs#L18) :+1:

gmarz:
Closing as we decided this is too invasive.  Will revisit again for 6.0.

Issue #1986, opened 2016-04-04
Add support for reindex rethrottle


gmarz:
https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/reindex.rethrottle.json

russcam:
Should be backported to 2.x - https://github.com/elastic/elasticsearch/pull/18020

PR #2083, opened 2016-05-09
Feature/reindex rethrottle

russcam:
Closes https://github.com/elastic/elasticsearch-net/issues/1986

requires backporting to `2.x` branch (dependent on answer to https://github.com/elastic/elasticsearch/issues/17031#issuecomment-217801149)


gmarz:
LGTM :+1:.  My only comment is that we possibly rename `ReindexRethrottle()` to just `Rethrottle()` to make it more obvious that it also applies to update by query.

russcam:
@gmarz Agreed. Will update the PR. Thanks for reviewing :+1:

Issue #2088, opened 2016-05-12
Rescore doesn't support multiple queries


gmoskovicz:
Rescore implementation currently only support a single query. When you use the fluent API you can only add a single rescore query. If you add multiple queries, a single one will be added.

Easy repro:

```
			// NameScoreFunction its just a customer FunctionScoreFunctionBase implementation

			var searchresults =
				client.Search<Session>(q => q
					  .Query(mainQuery)
					  .Rescore(r => r.RescoreQuery(rq => rq
								.Query(qq => qq
										.FunctionScore(fs => fs
											.Functions(new[] { new NameScoreFunction() { QueryName = "fullname", Field = "fullName" } })
								))
								.QueryWeight(0)
								.RescoreQueryWeight(1))
								.WindowSize(200)
							)
							.Rescore(r2 => r2
								.RescoreQuery(rq => rq
									.Query(qq => qq
										.FunctionScore(fs => fs
											.Functions(new[] { new NameScoreFunction() { QueryName = "company name", Field = "companyNames" } 
									})))
									.QueryWeight(.5)
									.RescoreQueryWeight(.5))
								.WindowSize(200)));
```

The result of this is the following: 

```
{
	"rescore": {
		"window_size": 200,
		"query": {
			"rescore_query": {
				"function_score": {
					"functions": [{
						"name_score": {
							"field": "companyNames",
							"query_name": "company name"
						}
					}]
				}
			},
			"query_weight": 0.5,
			"rescore_query_weight": 0.5
		}
	},
	"query": {
		"match": {
			"primary_name": {
				"query": "Joe Shmo"
			}
		}
	}
}
```

Note that the field is `companyNames` which comes from the second rescore query.

russcam:
Thanks for submitting @gmoskovicz :+1: 

This is a bug; you can work around it for the moment using the low level client and specifying the query using an anonymous type

```c#
	var rs = client.LowLevel.Search<Project>("index", "type", new
	{
		query = new
		{
			match = new
			{
				field1 = new
				{
					@operator = "or",
					query = "the quick brown",
					type = "boolean"
				}
			}
		},
        rescore = new object[] {
			new {
				window_size = 100,
				query = new {
					rescore_query = new {
						match = new {
						   field1 = new {
							  query = "the quick brown",
							  type = "phrase",
							  slop = 2
						   }
						}
					},
					query_weight = 0.7,
					 rescore_query_weight = 1.2
				}
			},
			new {
[...]

PR #2095, opened 2016-05-18
Allow Rescore to accept multiple rescore queries

russcam:
The interface for rescore, IRescore, cannot be changed in 2.x as it would break binary compatibility for 2.x.
A new concrete implementation of `IRescore`, `MultiRescore`, is introduced that allows for multiple rescore queries to be specified. When using OIS syntax, a new instance of `MultiRescore` should be used. When using the fluent API, the descriptor takes care of combining multiple rescore queries. A custom json converter takes care of serialization when a single or multiple rescore queries are specified.

The slightly awkward part with having `Rescore` and `MultiRescore` is that when an `IRescore` instance is deserialized from json, it is not known if this is a `Rescore` or `MultiRescore` without checking the type; a consumer would need to know to do this, as `MultiRescore` maps only the first `Rescore`'s properties to properties of the `IRescore` interface.

Add usage tests for both single Rescore and multiple Rescore queries.

Closes https://github.com/elastic/elasticsearch-net/issues/2088

### DIFF
--- a/docs/breaking-changes.asciidoc
+++ b/docs/breaking-changes.asciidoc
@@ -5,7 +5,6 @@
 --
 There are a number of breaking changes when going from 1.x to 2.x
 
-
 * <<elasticsearch-net-breaking-changes, Elasticsearch.Net Breaking Changes>>
 
 * <<nest-breaking-changes, NEST Breaking Changes>>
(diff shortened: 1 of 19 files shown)


Mpdreamz:
Love the bwc solution :+1:

gmoskovicz:
Nice!

Issue #2093, opened 2016-05-17
GetMany doesn't detect default index and type from generic T parameter


oguzhaneren:
**NEST/Elasticsearch.Net version**: 2.3.1

**Elasticsearch version**: 2.32

**Description of the problem including expected versus actual behavior**:

When i call GetMany method, nest thrown exception like below:
Request:
<pre>
_elasticClient.GetMany<Task>(taskIds)
</pre>

Exception:
<pre>
# FailureReason: BadResponse when trying to POST http://localhost:9200/_mget
# Audit trail of this API call:
 - BadResponse: Node: http://localhost:9200/ Took: 00:00:00.0414999
# ServerError: ServerError: 400Type: action_request_validation_exception 
Reason: "Validation Failed: 1: index is missing for doc 0;2: 
index is missing for doc 1;3:
</pre>

But when I add defaultIndex and type to GetMany call everythink is ok.

<pre>
 _elasticClient.GetMany<Task>(taskIds),index:_elasticClient.ConnectionSettings.DefaultIndex,type:"task"]);
</pre>

russcam:
Hey @oguzhaneren, thanks for submitting this, this is a bug.

When `index` and `type` are not specified (`null` values), the `index` and `type` should be inferred from the type `T` on `client.GetMany<T>()`.

gmarz:
Bumped to 2.3.3.

russcam:
Fixed in 2.x with  4b3744b3569179eda3a5f886caacab6d206f6bfe 

Forward ported to 5.x in  fe16bb7cd7aa8e4b7a18578122bb84ba3dda1640

PR #2096, opened 2016-05-18
Infer index and type from T for GetMany

russcam:
Set Index and Type on MultiGetDescriptor. If they have values, the specific index and type will be overridden on each get operation with them.

Fixes https://github.com/elastic/elasticsearch-net/issues/2093


russcam:
Will need forward porting to 5.x

gmarz:
LGTM :+1:  but (as already discussed) needs url tests.

PR #2100, opened 2016-05-19
Update auto-map.asciidoc

ericamick:


russcam:
I've updated the original C# file, thanks for the PR, @ericamick!

PR #2103, opened 2016-05-20
Update bool-dsl.asciidoc

ericamick:


gmarz:
@ericamick thank you for all these awesome doc PRs! :+1:

One issue though - we don't manually edit the asciidoc files directly.  They are generated by our build process from their respective `.doc.cs` files.

Would you mind making these changes to the cs files and opening a single PR with all of your changes?

russcam:
I've made the changes for this PR to the original C# files from which the docs have been generated. [I've also added a multiline comment at the top of each generated asciidoc](https://github.com/elastic/elasticsearch-net/commit/a4a45fba858e5abeeb71f157e6959ac95988cd75) linking back to the file from which each is generated so that it is easier to make a change and submit a PR in future :smile: 

Thanks for the doc PRs, @ericamick!

PR #2107, opened 2016-05-24
Fix double evaluation of IEnumerable methods

### LINKED ISSUE #2101: Code inside Filter executes twice
2.3.1

**Elasticsearch version**:
"version" : {
    "number" : "2.1.1",
    "build_hash" : "40e2c53a6b6c2972b3d13846e450e66f4375bd71",
    "build_timestamp" : "2015-12-15T13:05:55Z",
    "build_snapshot" : false,
    "lucene_version" : "5.3.1"
  },

I've the following code:
```
            var response = client.Search<ContactElastic>(
                s => s.Query(
                    q => q.Bool(
                    b => b
                        .Filter(f =>
                        {
                            var filter = f.Term(p => p.ClientId, 1);

                            return filter;
                        }))));

```

But I see that code inside Filter executes twice, why? Here is a simple code but inside my Filter side I've a sophisticated code that shouldn't to run twice.

gmarz:
The ToListOrNullIfEmpty() extension method called .Any() and .ToList()
which both evaluate the IEnumerable.  This caused funcs to be executed
twice when applied to an IEnumerable<Func<>>.

Closes #2101

### DIFF
--- a/src/Nest/CommonAbstractions/Extensions/Extensions.cs
+++ b/src/Nest/CommonAbstractions/Extensions/Extensions.cs
@@ -161,8 +161,11 @@ namespace Nest
 			foreach (T item in enumerable) handler(item);
 		}
 
-		internal static List<T> ToListOrNullIfEmpty<T>(this IEnumerable<T> enumerable) =>
-			enumerable.HasAny() ? enumerable.ToList() : null;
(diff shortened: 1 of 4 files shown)


gmarz:
Merged to 2.x and cherry picked to master.

PR #2109, opened 2016-05-25
Fix compound query conditionless logic

### LINKED ISSUE #2086: Term Filter with empty value compiles to nothing
2.0.5

**Elasticsearch version**:
"version" : {
    "number" : "2.1.1",
    "build_hash" : "40e2c53a6b6c2972b3d13846e450e66f4375bd71",
    "build_timestamp" : "2015-12-15T13:05:55Z",
    "build_snapshot" : false,
    "lucene_version" : "5.3.1"
  },

I've the following code for filter:
filter &= !f.Verbatim().Term(p => p.MobileNumber, "") && f.Exists(m => m.Field(p => p.MobileNumber));

but in the result search query I see just:
{"exists":{"field":"MobileNumber"}} 
without expected part with:
{"bool":{"must_not":[{"term":{"MobileNumber":{"value":''}}}]}}

Is the way I'm using Verbatim() correct? Don't understand why it doesn't work.

gmarz:
A compound query should only be conditionless when any of its inner
queries are not writable.  Previously we were only checking if the inner
queries were also not conditionless.

Closes #2086

### DIFF
--- a/src/CodeGeneration/CodeGeneration.LowLevelClient/CodeGeneration.LowLevelClient.project.lock.json
+++ b/src/CodeGeneration/CodeGeneration.LowLevelClient/CodeGeneration.LowLevelClient.project.lock.json
@@ -1,6 +1,6 @@
 {
   "locked": false,
-  "version": 2,
+  "version": 1,
   "targets": {
     ".NETFramework,Version=v4.5": {},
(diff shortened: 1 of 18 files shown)


gmarz:
Merged to 2.x and master

PR #2110, opened 2016-05-25
Make .NET core HttpConnection implementation more extensible

### LINKED ISSUE #2108: Allow some means to modify the request message when running on .NET core
**Describe the feature**:

We are using [ElasticSearch on AWS](https://aws.amazon.com/elasticsearch-service/). In our situation we need to be able to sign each request to ElasticSearch by adding an [Authorization header](http://docs.aws.amazon.com/general/latest/gr/sigv4_signing.html).

The HttpConnection class provides a `protected virtual` method named [CreateWebRequest](https://github.com/elastic/elasticsearch-net/blob/master/src/Elasticsearch.Net/Connection/HttpConnection.cs#L35). I can extend HttpConnection and override this method to add the Authorization header. This works great. In fact there's a [Nuget package](https://www.nuget.org/packages/Elasticsearch.Net.Aws/) which does exactly this. See their [AwsHttpConnection implementation](https://github.com/bcuff/elasticsearch-net-aws/blob/c2d59f060f5169e6fe1a303ecf1810cd86c82cf6/src/Elasticsearch.Net.Aws/Elasticsearch.Net.Aws/AwsHttpConnection.cs#L70-L92).

However, ElasticSearch.Net has a [different implementation of HttpConnection](https://github.com/elastic/elasticsearch-net/blob/master/src/Elasticsearch.Net/Connection/HttpConnection-CoreFx.cs#L116) for .NET Core which does not provide this same "hook".  The CreateHttpRequestMessage method is `private static`. I can't figure out a way to add the Authorization header required to use NEST to for ES on AWS.

Is this something that could be added?

gmarz:
Closes #2108

### DIFF
--- a/src/Elasticsearch.Net/Connection/HttpConnection-CoreFx.cs
+++ b/src/Elasticsearch.Net/Connection/HttpConnection-CoreFx.cs
@@ -117,7 +117,14 @@ namespace Elasticsearch.Net
 			return await builder.ToResponseAsync().ConfigureAwait(false);
 		}
 
-		private static HttpRequestMessage CreateHttpRequestMessage(RequestData requestData)
+		protected virtual HttpRequestMessage CreateHttpRequestMessage(RequestData requestData)
+		{
+			var request = this.CreateRequestMessage(requestData);
+			SetBasicAuthenticationIfNeeded(request, requestData);
+			return request;
+		}
+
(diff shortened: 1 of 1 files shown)


russcam:
LGTM :+1:

gmarz:
Merged to 2.x and master

PR #2114, opened 2016-06-01
Fix fuzziness for QueryStringQuery

MiniGab:


karmi:
Hi @MiniGab, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/2114.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

MiniGab:
Hello @karmi done!

gmarz:
LGTM!  Thanks for the PR @MiniGab :+1:

russcam:
Will backport to 2.x in bwc way

Issue #2115, opened 2016-06-02
SearchDescriptor Index(Indices index) doesn't support empty string


xrayf:
NEST Version 2.3.1

Im creating a new SearchDescriptor with following code:

`var desc = new SearchDescriptor<object>();`
When im trying to fill the object with sepcific method calls like:
`desc.Index(string.Empty)`
Im getting an exception here, which tells me "can not parse an empty string to Indices."

Maybe it is right, that an empty string isn't parsable into Indices. But your method summary of the Index Method on SearchDescriptor says:

>         //
        // Summary:
        //     A comma-separated list of index names to search; use `_all` or empty string to
        //     perform the operation on all indices

Thats wrong, because you see, an empty string causes an exception.

Mpdreamz:
Aye, this is expected and tested behavior. We generate as much as we can from:

https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/search.json#L11

Which has that message, we should probably rewrite that while generating.

xrayf:
Yep, anything should be changed. The method description or the restriction of empty strings.

PR #2130, opened 2016-06-20
Convert empty string to _all indices

russcam:
Closes https://github.com/elastic/elasticsearch-net/issues/2115


Mpdreamz:
I don't feel comfertable support `""` to `_all` implicit conversion feels like a slipery slope :) I rather like the exception you get now, passing an empty string is hardly ever intended.

PR #2134, opened 2016-06-22
Move CancellationToken as an optional parameter to all async method

### LINKED ISSUE #2042: CancellationToken Parameter for Async Methods
**Describe the feature**:
Why don't add a CancellationToken Parameter (overload) to the async methods like:

`Task<ISearchResponse<T>> SearchAsync<T>(Func<SearchDescriptor<T>, ISearchRequest> selector = null) where T : class;`

gmarz:
We've always supported CancellationToken but as a request config param, this PR moves it more explicitly to the method as additional argument

Note: I've also removed `CancellationToken` from `IRequestConfiguration`.  Since all async methods now accept a cancellation token, it didn't seem necessary anymore.  Thoughts @Mpdreamz @russcam?

Closes #2042


gmarz:
My last commit removes `CancellationToken` from `RequestData` and instead enforces it on all of the low-level async methods, including `RequestAsync()`.  /cc @Mpdreamz

Mpdreamz:
LGTM :+1: It'd be great if we can add some tests around cancellations by simulating slow parts in various places in the pipeline. We might need to reavaluate if we use `IsCancellationRequested`/`ThrowIfCancellationRequested` too sparsely or not.

But we can tackle that as part of a new PR

PR #2135, opened 2016-06-23
Check the token type of score property when deserializing aggregation

russcam:
The deserialization assumes that a property named "score" is an indicator that an aggregation is a significant terms aggregation, but this may not be the case; "score" could be a user-supplied name for an aggregation.


Mpdreamz:
LGTM :+1:

PR #2154, opened 2016-07-11
Fix gzip compression when running on .NET core

### LINKED ISSUE #2150: EnableHttpCompression doesn't work correctly
NEST/Elasticsearch.Net version : 2.3.3

Elasticsearch version: 2.3.2

Description of the problem including expected versus actual behavior:
I'm trying to use the compression between Client and Server and getting the following Error:

"System.InvalidOperationException: Misused header name. Make sure request headers are used with HttpRequestMessage, response headers with HttpResponseMessage, and content headers with HttpContent objects". 

I have used: "settings.EnableHttpCompression()"; at client and "http.compression:true" at server.

I'm using .net core rc1.

It seems to be a bug, or I'm making something wrong...

gmarz:
- Content-Encoding was being added to the wrong header
 - Stream was being closed pre-maturely

Closes #2150

### DIFF
--- a/src/Elasticsearch.Net/Connection/HttpConnection-CoreFx.cs
+++ b/src/Elasticsearch.Net/Connection/HttpConnection-CoreFx.cs
@@ -144,21 +144,18 @@ namespace Elasticsearch.Net
 			if (data != null)
 			{
 				var stream = requestData.MemoryStreamFactory.Create();
-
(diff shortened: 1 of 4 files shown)


Mpdreamz:
LGTM

gmarz:
Ported to [master](84513971d30d57341049d58d06e1dd692f05919c).

PR #2160, opened 2016-07-13
Return of the reindex helper, this was partially removed in the alpha…

Mpdreamz:
…s while we figured out how it interacted with the new reindex api's

Refactored it a tad and introduced an option to alter the index operation / document when using the `ReindexObserver`

Fixed CountRequest for 5.0 alpha we now default to an explicit match_all query

pending https://github.com/elastic/elasticsearch/issues/19422


gmarz:
LGTM 👍

PR #2161, opened 2016-07-14
Removed trash

babadofar:


Mpdreamz:
This trash actually gets generated from:

```
https://github.com/elastic/elasticsearch-net/blob/master/src/Tests/Aggregations/Pipeline/MovingAverage/MovingAverageHoltWintersAggregationUsageTests.cs#L112
```

We will go over all of these and add more assertions and example response handling. PR's there more then welcomed! :smile:

PR #2162, opened 2016-07-14
Added a general purpose BulkHelper that will partition and paralleliz…

Mpdreamz:
Marked WIP but 98% done

This adds a `.PartitionedBulk()` method to the client that can take really large `IEnumerable<T>` that we itterate and partition lazy and parallelize as bulk request to elasticsearch. 

You can specify bulk parameters such as routing and global pipeline id.

Example usage:

Consider we create a lazy stream of 100k documents

```csharp
var size = 1000;
var pages = 100;
var seenPages = 0;
var numberOfDocuments = size * pages;
var documents = this.CreateLazyStreamOfDocuments(numberOfDocuments);
```

We can then set up our partitioned bulk as followed:

```csharp
//first we setup our cold observable
var observableBulk = this._client.PartitionedBulk(documents, f => f
	.MaxDegreeOfParallelism(8)
	.BackOff(TimeSpan.FromSeconds(10))
	.NumberOfBackOffRetries(2)
	.Size(size)
	.RefreshOnCompleted()
	.Index(IndexName)
);
```

We partition the 100k documents over `size: 1000` meaning we'll send a 100 bulk requests. We setup max degree of parallelism of 8 meaning 8 buffers of size 1000 can only be active at a time. We also setup automatic backoff of requests when we see certain documents failed to index due to 429 errors through `NumberOfBackOffRetries` and `BackOff` which dictate we wait 10 seconds per retry and we wont retry more then 2 times. This is different from the client default cluster failover retries which retry on 502-504 when a node is down. Here we retry with a back off period because the node is busy. 

```csharp
//we set up an observer
var bulkObserver = new PartitionedBulkObserver(
	onError: (e) => { throw e; },
	onCompleted: () => handle.Set(),
	onNext: (b) => Interlocked.Increment(ref seenPages)
);
````
Here we set up a simple observer can be any `IObserver<BulkRequest>` but if we use this one we can track some interesting statistics

````csharp
//when we subscribe the observable becomes hot
observableBulk.Subscribe(bulkObserver);
handle.WaitOne(TimeSpan.FromMinutes(5));

```csharp
seenPages.Should().Be(pages);
var count = this._client.Count<SmallObject>(f => f.Index(IndexName));
count.Count.Should().Be(numberOfDocuments);
```

Here we validate the index contains the expected number of documents and `OnNext()` has been called the expected number of times.

```csharp
bulkObserver.TotalNumberOfFailedBuffers.Should().Be(0);
bulkObserver.TotalNumberOfRetries.Should().Be(0);
```

Here we see some interesting statistics of our coordinated request. FailedBuffers are those that depleeded their retries and did not index everything.


TODO
- [x] wire in CancellationToken
- [x] Im not 100% sold on PartitionedBulk, suggestions?
- [x] BackOff rename to BackOffTime?
- [x] The OnNext() recieves the last bulkresponse of a buffer but not sure how useful that is if we only need to retry 2 items due to a 429 out of an orignal buffer of 1000.


gmarz:
> Im not 100% sold on PartitionedBulk, suggestions?

BulkMany?

>BackOff rename to BackOffTime?

+1.  I'd also rename `NumberOfBackOffRetries` to just `BackOffRetries`

Mpdreamz:
I went with `BulkAll()` quite like it :)

Finished the last open issues ready for review

src/Nest/Document/Multiple/BulkAll/BulkAllRequest.cs
```diff
@@ -0,0 +1,148 @@
+﻿using System.Collections.Generic;
+using Elasticsearch.Net;
+
+namespace Nest
+{
+	public interface IBulkAllRequest<T>
```
gmarz: Would be nice if we included the default values in the xmldocs

src/Nest/Document/Multiple/BulkAll/BulkAllResponse.cs
```diff
@@ -0,0 +1,32 @@
...
+	///
+	/// </summary>
+	public interface IBulkAllResponse
+    {
+        /// <summary>This is the Nth buffer.</summary>
+		long Page { get; }
+
+        /// <summary>The number of back off retries were needed to store this document.</summary>
+		int Retries { get; }
+    }
```
gmarz: What about `Errors` and `ItemsWithErrors`?  Can we accumulate them from the individual `Bulk()` calls?

gmarz:
Few comments but otherwise LGTM.  I love this feature and I also +1 for the name `BulkAll` 👍 

I think we can also back port this to 2.x?

src/Nest/Document/Multiple/BulkAll/BulkAllObservable.cs
```diff
@@ -0,0 +1,219 @@
...
+				}
+			}
+
+			public Task ForEachAsync<TResult>(
+				Func<IList<TDocument>, long, Task<TResult>> taskSelector,
+				Action<IList<TDocument>, TResult> resultProcessor,
+				Action<Task> done
+			)
+			{
+				var semaphore = new SemaphoreSlim(initialCount: _semaphoreSize, maxCount: _semaphoreSize);
```
russcam: Could use a concurrent [`Partitioner`](https://msdn.microsoft.com/en-us/library/system.collections.concurrent.partitioner.create(v=vs.110).aspx) for this, similar to the [`ForEachAsync()` implementation by Stephen Toub](http://blogs.msdn.com/b/pfxteam/archive/2012/03/05/10278165.aspx):

```c#
public static Task ForEachAsync<T>(this IEnumerable<T> source, int dop, Func<T, Task> body) 
{ 
    return Task.WhenAll( 
        from partition in Partitioner.Create(source).GetPartitions(dop) 
        select Task.Run(async delegate { 
            using (partition) 
                while (partition.MoveNext()) 
                    await body(partition.Current); 
        })); 
}
```

Mpdreamz: I'm already using a stephen toub variant :+1:  The method you link relies on the reader to be choped up into `dop` (degree of parallelism) and then itterate over these chunks and do the work on that. 

Won't work with 10.000.000 record dataset chopped into 4 concurrent unit of works. 

In our case the writer (to elasticsearch) is concurrent and whoever feeds us the `IEnumerable` does not have to make any guarantees on the thread safety of the collection or make sure it can witstand calling MoveNext too many times (as would be the case by using a double generator completely lazy approach). 

In *most* cases i reckon the real concurrency benefit is not reading from source (xml, db record set, files) in parallel but parallize the bulk indexing to elasticsearch. 

If it is, folks can still create their own more explicit concurrent read implementation behind the `IEnumerable<T>`

russcam:
Awesome stuff @Mpdreamz!

russcam:
@Mpdreamz: plans to backport this to 2.x?

Misterhex:
This PR is awesome.. i did something similar but on a separate repo.. https://github.com/Misterhex/NESTLogger

PR #2165, opened 2016-07-15
Fix #2163 Add CancellationToken parameter to async extension methods

### LINKED ISSUE #2163: Async extension methods should also directly expose cancellationtoken
As part of #2134 we moved cancellationtokens to more explicit and discoverable method arguments. 

We missed several async extension methods on `IElasticClient` we should get them in line with the rest.

gmarz:


Mpdreamz:
LGTM :+1:

Issue #2008, opened 2016-04-05
Update cpu and suggest stats


gmarz:
https://www.elastic.co/guide/en/elasticsearch/reference/master/breaking_50_index_apis.html#_system_cpu_stats

https://www.elastic.co/guide/en/elasticsearch/reference/master/breaking_50_index_apis.html#_suggest_stats

PR #2180, opened 2016-07-29
Add load average stats to _cat/nodes and indices/stats

TheFireCookie:
Hello,

I saw this commit 1122da55 is not in the branch 2.x. Why?

So there is a pull request to bring it if it's possible.

Thanks,
Matthias.

See https://www.elastic.co/guide/en/elasticsearch/reference/master/breaking_50_index_apis.html#_system_cpu_stats
Closes https://github.com/elastic/elasticsearch-net/issues/2008


karmi:
Hi @TheFireCookie, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/2180.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

gmarz:
This wasn't merged to 2.x because it contains breaking changes that only apply to ES 5.x, namely the new [CPUStats](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Cluster/NodesStats/NodeStats.cs#L106) object in `_nodes/stats`.

Is there any particular reason why you think this should also be in 2.x that maybe we're missing?

TheFireCookie:
The thing I'm missing is os.percent that is available in the cluster nodes stats API of ElasticSearch. I don't understand why it's not in NEST?

Thanks for your answer.

gmarz:
Looks like it's just not mapped.  That's a different change from the attached commit though.  Would you like to send new a PR for this?

TheFireCookie:
Yep I'll do it on monday if I find the time for it.

gmarz:
Awesome :+1: Much appreciated.  No worries if not though, we'll add it ourselves for the next release if you don't get it to it.

Closing this one.  I'll open a new issue for 2.x changes.

Issue #2155, opened 2016-07-12
FieldValues inference converts timestamp strings to date/times then to string in different format


wewebber:
**NEST/Elasticsearch.Net version**: 2.4.1

**Elasticsearch version**: 2.33

**Description of the problem including expected versus actual behavior**:

If we have a string field holding a value that happens to look like a timestamp, and retrieve that field through a Fields() query, the field will get reformatted into a different form of date / time.

E.g. we have the mapping:

```
"mything" : {
    "myname" : {
       "type" : "string"
    }
}
```

fronted by the class:

```c#
public class MyThing
{
    [String]
    public string MyName { get; set; }
}
```

We insert the following mything:

```c#
new MyThing() { MyName = "2016-07-12T06:06:35.203Z" }
```

and do a search like:

```c#
var name = client.Search<MyThing>(sd => sd.Fields(fd => fd.Field(mt => mt.MyName)))
              .Fields.Select(fv => fv.Values<MyThing, string>(mt => mt.MyName)[0]).ToArray().[0]
```

We'll then find that `name` has a value like "12/07/2016 06:06:35" (depending I think on language settings).

The culprit I think is in https://github.com/elastic/elasticsearch-net/blob/cdea256043b9c7fbc7a80c806d94a08a345236a3/src/Nest/CommonAbstractions/Fields/FieldValues.cs#L63, where NEST uses Newtonsoft.Json.Linq.JArray to infer the type of the field.  It infers it to be some sort of DateTime, then does a ToString() on that DateTime.

Mpdreamz:
Thanks for reporting this @wewebber this has now been fixed in 2.x and `5.x` :+1:

wewebber:
Excellent, thanks!

PR #2182, opened 2016-07-31
Use custom connection serialization settings, if available.

mitchknife:
This solves a problem similar to #2155 but for multi-get. My date strings were not round tripping correctly.

Digging into it I found that my custom `JsonSerializationSettings` were being ignored in `JsonNetSerializer` (`DateParseHandling.None` specifically) during a multi-get call.

This is more of a comprehensive solution than #2155 since the problem looks to be a bit more systemic than it did at first.

I was unable to run *all* tests at once (before and after the change), but ran a number of them individually.

### DIFF
--- a/src/Nest/CommonAbstractions/SerializationBehavior/JsonNetSerializer.cs
+++ b/src/Nest/CommonAbstractions/SerializationBehavior/JsonNetSerializer.cs
@@ -24,7 +24,14 @@ namespace Nest
 		private readonly Dictionary<SerializationFormatting, JsonSerializer> _defaultSerializers;
 		private readonly JsonSerializer _defaultSerializer;
 
-		protected virtual void ModifyJsonSerializerSettings(JsonSerializerSettings settings) { }
+		protected virtual void ModifyJsonSerializerSettings(JsonSerializerSettings settings)
+		{
+			// Apply custom serialization settings from the connection serializer, if any; avoiding infinite loop with 'this' check.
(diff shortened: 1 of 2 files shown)


karmi:
Hi @mitchknife, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/2182.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

mitchknife:
Ok, updated my profile with the new email.

karmi:
@mitchknife, cool, the CLA is passing now.

gmarz:
Hey @mitchknife thanks for the PR.  This is definitely an issue, but I'm not quite sold on this solution because it doesn't protect against the case where `ModifyJsonSerializerSettings()` is overridden in a derived class but doesn't call the base implementation first.

The problem stems from the fact that we have a [custom MultiGetConverter](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Document/Multiple/MultiGet/ElasticClient-MultiGet.cs#L60) that disregards any registered custom `JsonNetSerializer` and uses the default.  We should fix the problem there rather than in the `JsonNetSerializer` itself.

Will think of a solution and open a new PR for this, but huge :+1: for raising this.  There maybe other instances (other endpoints) that suffer from this as well, so we also need to do a pass over all the endpoints that use custom converters in this fashion.

gmarz:
@mitchknife see the attached PR above.

PR #2183, opened 2016-08-01
Lazily resolve JsonSerializers in JsonNetSerializer

tsliang:
The current JsonNetSerializer implementation creates the default
serializers
within its constructor, which then calls ModifyJsonSerializerSettings.
This
means that a subclass of JsonNetSerializer cannot modify the
JsonSerializerSettings with fields set by the constructor, because the
order of
operations is:

  1. Call subclass constructor with parameter A, which will set subclass
  field A; field A will be used in ModifyJsonSerializerSettings.
  2. Subclass constructor calls JsonNetSerializer constructor with
     :base()
  3. JsonNetSerializer constructor calls ModifyJsonSerializerSettings
     as part of JsonSerializer.Create
  4. ModifyJsonSerializerSettings makes use of default value for
     field A, because it has not been set yet.
  5. JsonSerializer constructor exits
  6. Subclass constructor sets value of field A from parameter
     A.

  By using the Lazy<T> class, we can ensure that ModifyJsonSerializerSettings is
  called after the subclass constructor has exited, which allows fields to
  be set and thereby used in ModifyJsonSerializerSettings. The unit test
  in Connection.doc.cs has been expanded to demonstrate this capability
  with a (fairly trivial) use case.

### DIFF
--- a/src/Nest/CommonAbstractions/SerializationBehavior/JsonNetSerializer.cs
+++ b/src/Nest/CommonAbstractions/SerializationBehavior/JsonNetSerializer.cs
@@ -19,10 +19,10 @@ namespace Nest
 		protected ElasticContractResolver ContractResolver { get; }
 
 		//todo this internal smells
-		internal JsonSerializer Serializer => _defaultSerializer;
+		internal JsonSerializer Serializer => _defaultSerializer.Value;
 
-		private readonly Dictionary<SerializationFormatting, JsonSerializer> _defaultSerializers;
-		private readonly JsonSerializer _defaultSerializer;
+		private readonly Lazy<Dictionary<SerializationFormatting, JsonSerializer>> _defaultSerializers;
+		private readonly Lazy<JsonSerializer> _defaultSerializer;
 
 		protected virtual void ModifyJsonSerializerSettings(JsonSerializerSettings settings) { }
 		protected virtual IList<Func<Type, JsonConverter>> ContractConverters => null;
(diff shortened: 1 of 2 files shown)


karmi:
Hi @tsliang, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/2183.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

tsliang:
@karmi fixed author and email

gmarz:
Thanks for this @tsliang :+1: 

For the most part, this LGTM.  However, calling a virtual method in a base constructor as we're doing here  when calling `ModifyJsonSerializerSettings()` is generally a bad idea, for reasons exactly like this.  Lazy instantiating the serializers feels like we're working around the problem rather than addressing it...but it might be the only way to fix this though.

Would love to get @Mpdreamz and @russcam s opinion on this.

gmarz:
@tsliang I've proposed an alternate solution in the PR attached above which also ties into the issue raised in #2183.

tsliang:
@gmarz I usually try to avoid construct-with-init (hence the use of Lazy) - IMHO classes should be fully initialized upon construction and not require a separate init method to be called afterwards, lest someone forget to call the init method after construction. Still, this should get the job done. What is the general ES policy on construct-with-init? Are those used a lot in the project?

SayliS:
What about custom resolvers? Right now I have problems setting my own :(

tsliang:
@SayliS what sort of problems?

gmarz:
> IMHO classes should be fully initialized upon construction and not require a separate init method to be called afterwards

@tsliang I totally agree but I think this is a scenario where I can live with it, especially because it gives us a way to solve #2183 as well.  I like the idea of using the lazy constructs, but still feels like we're working around the problem rather than fixing the poor design.

> What is the general ES policy on construct-with-init? Are those used a lot in the project?

We have no policy on this and not something we do a lot at all.  This may be one of the only instances.

tsliang:
@gmarz ok, that makes sense. I have proposed one other solution on #2189. Maybe you can take a look and see what you think? I think it'll solve the problem without requiring any significant changes in how the JsonNetSerializer is used. If it doesn't suit you, though, then let's just do #2189 and I'll close this PR.

gmarz:
@SayliS I'm assuming you mean adding your own custom converters?  You can do this by overriding [ContractConverters](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/CommonAbstractions/SerializationBehavior/JsonNetSerializer.cs#L28) in your custom serializer.  Is that not working for you?

mynkow:
@SayliS  is talking about Newtonsoft-Json Resolvers actually. The problem is that it is impossible to use dependency injection with virtual methods called in the base class.

Here is an example what is my problem and @SayliS 

https://msdn.microsoft.com/en-us/library/ms182331.aspx?f=255&MSPPError=-2147217396

gmarz:
@mynkow gotcha.  Yea, that's exactly what this PR and #2189 is aiming to solve.  Let's close this issue and move the discussion over there to eliminate the redundancy.

@tsliang I'll pull these changes into #2189 if we choose to go this route.

Issue #2181, opened 2016-07-29
cpu_percent missing from NodesStats


gmarz:
Follow up from #2180 

```json
      "os" : {
        "timestamp" : 1469805611911,
        **"cpu_percent" : 18,**
        "load_average" : -1.0,
        "mem" : {
          "total_in_bytes" : 16843710464,
          "free_in_bytes" : 10167721984,
          "used_in_bytes" : 6675988480,
          "free_percent" : 60,
          "used_percent" : 40
        },
        "swap" : {
          "total_in_bytes" : 19930718208,
          "free_in_bytes" : 11793481728,
          "used_in_bytes" : 8137236480
        }
      }
```

This applies to `2.x` only.  /cc @TheFireCookie

gmarz:
Closed via #2186

PR #2185, opened 2016-08-01
Add Percent metric in OperatingSystemStats

TheFireCookie:
Add Percent metric in OperatingSystemStats to solve this issue:

https://github.com/elastic/elasticsearch-net/issues/2181

### DIFF
--- a/src/Nest/Cluster/NodesStats/NodeStats.cs
+++ b/src/Nest/Cluster/NodesStats/NodeStats.cs
@@ -87,8 +87,8 @@ namespace Nest
 		public long Timestamp { get; internal set; }
 		[JsonProperty("load_average")]
 		public float LoadAverage { get; internal set; }
-		[JsonProperty("percent")]
-		public int Percent { get; internal set; }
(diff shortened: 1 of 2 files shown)


gmarz:
@TheFireCookie out of curiosity what version of ES are you running? I tested against version 2.3.4 and the property is `os.cpu_percent`, not `os.percent`.

Also, would be great if you could add an assertion to the integration test here: https://github.com/elastic/elasticsearch-net/blob/2.x/src/Tests/Cluster/NodesStats/NodesStatsApiTests.cs#L98

TheFireCookie:
Sorry, I just went on the doc of ElasticSearch there: https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-stats.html#os-stats

I'll correct my pull request and write a test then.

gmarz:
:+1: thanks!  One last request...would you mind rebasing and squashing this PR to a single commit?

TheFireCookie:
The new one: https://github.com/elastic/elasticsearch-net/pull/2186

PR #2190, opened 2016-08-03
solve _source false issue

pathikdevani:
when i set Source(false) then Nest replace exclude:["\*"]. exclude:["\*"] is effect on performance. it's very slower then false.

### DIFF
--- a/src/Nest/Aggregations/Metric/TopHits/TopHitsAggregation.cs
+++ b/src/Nest/Aggregations/Metric/TopHits/TopHitsAggregation.cs
@@ -18,7 +18,7 @@ namespace Nest
 		IList<ISort> Sort { get; set; }
 
 		[JsonProperty("_source")]
-		ISourceFilter Source { get; set; }
+		object Source { get; set; }
 
 		[JsonProperty("highlight")]
 		IHighlight Highlight { get; set; }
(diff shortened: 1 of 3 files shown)


gmarz:
Hey @pathikdevani thanks for the PR.

We have a [Union](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/CommonAbstractions/Union/Union.cs) type specifically for dealing with properties that can be of multiple types in ES.  For instance, [StopWords](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Analysis/StopWords.cs) can be a `string` or `IEnumerable`, [Like](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/QueryDsl/Specialized/MoreLikeThis/Like/Like.cs) is either a `string` or `ILikeDocument`.

I think leveraging `Union<bool, Fields>` type here instead would be better suited than just `object`. 

Does that make sense?  If you'd like to tackle this and send a new PR that would be awesome, otherwise we're happy to implement ourselves.  Also, FYI this is not something we could fix in the `2.x` branch as it's a breaking change.  It would only go into `5.x` (master).

pathikdevani:
hello @gmarz I create new PR #2199 ans solve this same issue using Union in branch 2.x

gmarz:
Closing in favor of #2200

Issue #2176, opened 2016-07-27
Search phrase suggester collate query (ElasticSearch 2.x)


HarveyWilliams:
Hi, I have been looking into using the collate option of a phrase suggestion to filter out suggestions that would not produce search query results. There is some documentation here: https://www.elastic.co/guide/en/elasticsearch/client/net-api/2.x/suggest-usage.html

I see that the query currently needs to be inline:

```
 .Collate(c => c
     .Query(q => q
         .Inline("{ \"match\": { \"{{field_name}}\": \"{{suggestion}}\" }}")
         .Params(p => p.Add("field_name", "title"))
     )
     .Prune()
)
```

However, this means that I lose any Intellisense, and if I change my mappings in my index, no errors will be thrown until query time.

I can see that there is another another option instead of the inline search descriptor. It appears I can use a previously indexed query by using the `.Indexed()` option:

```
 .Collate(c => c
     .Query(q => q
         .Indexed(/* what goes here? */)
     )
     .Prune()
)
```

Though there doesn't seem to be any documentation on this and no tests either. Can someone point me in the right direction?

Thanks in advance.

gmarz:
> I see that the query currently needs to be inline ... However, this means that I lose any Intellisense

The reason for this is that `Query()` is not just any `QueryDescriptor` as it is most everywhere else.  It's actually a [template query](https://www.elastic.co/guide/en/elasticsearch/reference/2.3/query-dsl-template-query.html) which is based on the Mustache templating system.  `Inline` being the query template itself.  We could perhaps type `Inline` as a `Func<QueryDescriptor, IQuery>`, but then it may not be obvious that it's a template.

The bigger issue here is that the collate [Query](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Search/Suggesters/PhraseSuggester/PhraseSuggestCollate.cs#L12) option is typed incorrectly as `IScript` instead of `ITemplateQuery`.  Unfortunately, we can only fix this in 5.x as it would be a breaking change.  But fortunately, `IScript` and `ITemplateQuery` are pretty much identical, so we can get away with it.  This is probably what caused the mix up to begin with.

This explains the confusion around `Indexed`, which is part of `IScript` and used to referenced an indexed script.  It is the equivalent of referencing a [stored template](https://www.elastic.co/guide/en/elasticsearch/reference/2.3/query-dsl-template-query.html#_stored_templates).

Hope that makes sense.  Sorry for the confusion!

[...]

PR #2192, opened 2016-08-04
"collate_match" for phrase suggestions

HarveyWilliams:
Fixed issue #2176 where the "collate_match" boolean wasn't accessible through the the SuggestOption.

### DIFF
--- a/src/Nest/Search/Suggesters/SuggestOption.cs
+++ b/src/Nest/Search/Suggesters/SuggestOption.cs
@@ -31,7 +31,10 @@ namespace Nest
 		[JsonProperty("highlighted")]
 		public string Highlighted { get; internal set; }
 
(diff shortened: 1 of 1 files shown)


HarveyWilliams:
Just signed the agreement.

karmi:
Hi @HarveyWilliams, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/2192.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

HarveyWilliams:
@karmi Thanks Karmi, didn't realise that. I'm having trouble with adding the other email (it's added but the check is still failing). I'm just gonna try pushing with the correct email and I'll make a new PR.

PR #2197, opened 2016-08-05
Fix/serialization bind

Mpdreamz:
Yet another take on 

#2189
#2183
#2182

That does not require locking patterns and construct then init like composition.

This introduces an ISerializerFactory that gets set on IConnectionSettings which we can then use to instantiate the IElasticsearchSerializer the user provided throughout, getting rid of hardcoded `new JsonNetSerializer()` in some custom converters. 

Obsoletes the func based factory and wraps it to our default `SerializerFactory` 

The way you have to reset serializers is a tad clunky though

https://github.com/elastic/elasticsearch-net/compare/2.x...fix/serialization-bind?expand=1#diff-37181b379c228340d8420dc00e71cd63R57

And implementers should now also implement the stateful constructor to be correct. 

If they don't we fall back on the current (broken but bwc) behaviour.

Keen to hear your thoughts @gmarz @tsliang @mitchknife @russcam

### DIFF
--- a/src/Nest/CommonAbstractions/ConnectionSettings/ConnectionSettingsBase.cs
+++ b/src/Nest/CommonAbstractions/ConnectionSettings/ConnectionSettingsBase.cs
@@ -19,16 +19,23 @@ namespace Nest
 			: this(new SingleNodeConnectionPool(uri ?? new Uri("http://localhost:9200"))) { }
 
 		public ConnectionSettings(IConnectionPool connectionPool)
-			: this(connectionPool, null, null) { }
+			: this(connectionPool, null, new SerializerFactory()) { }
 
 		public ConnectionSettings(IConnectionPool connectionPool, IConnection connection)
-			: this(connectionPool, connection, null) { }
+			: this(connectionPool, connection, new SerializerFactory()) { }
 
 		public ConnectionSettings(IConnectionPool connectionPool, Func<ConnectionSettings, IElasticsearchSerializer> serializerFactory)
+#pragma warning disable CS0618 // Type or member is obsolete
 			: this(connectionPool, null, serializerFactory) { }
+#pragma warning restore CS0618 // Type or member is obsolete
 
+		public ConnectionSettings(IConnectionPool connectionPool, IConnection connection, ISerializerFactory serializerFactory)
+			: base(connectionPool, connection, serializerFactory, s => serializerFactory.Create(s)) { }
+
+		[Obsolete("Please use the constructor taking ISerializerFactory instead of a Func")]
 		public ConnectionSettings(IConnectionPool connectionPool, IConnection connection, Func<ConnectionSettings, IElasticsearchSerializer> serializerFactory)
-			: base(connectionPool, connection, serializerFactory) { }
+			: base(connectionPool, connection, null, s => serializerFactory?.Invoke(s)) { }
+
 	}
 
 	/// <summary>
@@ -37,7 +44,7 @@ namespace Nest
 	[Browsable(false)]
 	[EditorBrowsable(EditorBrowsableState.Never)]
 	public abstract class ConnectionSettingsBase<TConnectionSettings> : ConnectionConfiguration<TConnectionSettings>, IConnectionSettingsValues
-		where TConnectionSettings : ConnectionSettingsBase<TConnectionSettings>
+		where TConnectionSettings : ConnectionSettingsBase<TConnectionSettings>, IConnectionSettingsValues
 	{
 		private string _defaultIndex;
 		string IConnectionSettingsValues.DefaultIndex => this._defaultIndex;
@@ -63,16 +70,35 @@ namespace Nest
 		private readonly FluentDictionary<MemberInfo, IPropertyMapping> _propertyMappings = new FluentDictionary<MemberInfo, IPropertyMapping>();
 		FluentDictionary<MemberInfo, IPropertyMapping> IConnectionSettingsValues.PropertyMappings => _propertyMappings;
 
-		protected ConnectionSettingsBase(IConnectionPool connectionPool, IConnection connection, Func<TConnectionSettings, IElasticsearchSerializer> serializerFactory)
-			: base(connectionPool, connection, serializerFactory)
+		private readonly ISerializerFactory _serializerFactory;
+		ISerializerFactory IConnectionSettingsValues.SerializerFactory => _serializerFactory;
+
+		protected ConnectionSettingsBase(
+			IConnectionPool connectionPool,
+			IConnection connection,
+			ISerializerFactory serializerFactory,
+			Func<TConnectionSettings, IElasticsearchSerializer> serializerFactoryFunc
+			)
+			: base(connectionPool, connection, serializerFactoryFunc)
 		{
 			this._defaultTypeNameInferrer = (t => t.Name.ToLowerInvariant());
 			this._defaultFieldNameInferrer = (p => p.ToCamelCase());
 			this._defaultIndices = new FluentDictionary<Type, string>();
 			this._defaultTypeNames = new FluentDictionary<Type, string>();
+			this._serializerFactory = serializerFactory ?? new SerializerFactory();
 
 			this._inferrer = new Inferrer(this);
 		}
+
+		protected ConnectionSettingsBase(
+			IConnectionPool connectionPool,
+			IConnection connection,
+			Func<TConnectionSettings, IElasticsearchSerializer> serializerFactoryFunc
+			)
+			: this(connectionPool, connection, null, serializerFactoryFunc) { }
+
+		IElasticsearchSerializer IConnectionSettingsValues.StatefulSerializer(JsonConverter converter) =>
+			this._serializerFactory.CreateStateful(this, converter);
 
 		/// <summary>
 		/// The default serializer for requests and responses
@@ -92,7 +118,7 @@ namespace Nest
 		/// <summary>
 		/// The default index to use when no index is specified.
 		/// </summary>
-		/// <param name="defaultIndex">When null/empty/not set might throw 
+		/// <param name="defaultIndex">When null/empty/not set might throw
 		/// <see cref="NullReferenceException"/> later on when not specifying index explicitly while indexing.
 		/// </param>
 		public TConnectionSettings DefaultIndex(string defaultIndex)
@@ -177,10 +203,10 @@ namespace Nest
 			var mapper = new PropertyMappingDescriptor<TDocument>();
 			propertiesSelector(mapper);
 			ApplyPropertyMappings(mapper.Mappings);
-			return (TConnectionSettings) this;
+			return (TConnectionSettings)this;
 		}
 
-		private void ApplyPropertyMappings<TDocument>(IList<IClrTypePropertyMapping<TDocument>> mappings) 
+		private void ApplyPropertyMappings<TDocument>(IList<IClrTypePropertyMapping<TDocument>> mappings)
 			where TDocument : class
 		{
 			foreach (var mapping in mappings)
@@ -198,7 +224,7 @@ namespace Nest
 				{
 					var newName = mapping.NewName;
 					var mappedAs = _propertyMappings[memberInfo].Name;
-					var typeName = typeof (TDocument).Name;
+					var typeName = typeof(TDocument).Name;
 					if (mappedAs.IsNullOrEmpty() && newName.IsNullOrEmpty())
 						throw new ArgumentException("Property mapping '{0}' on type is already ignored"
 							.F(e, newName, mappedAs, typeName));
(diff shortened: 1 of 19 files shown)


Mpdreamz:
This also attaches some commits to highlight the chicken egg problem.

[Happy flow for overriding `ModifyJsonSerializerSettings`](https://github.com/elastic/elasticsearch-net/blob/c39b4c5107a5c2682e590db4c4885039ac341c2a/src/Tests/ClientConcepts/Serializer/ModifySerializationSettingsTests.cs)
[Where that goes wrong with custom constructors](https://github.com/elastic/elasticsearch-net/blob/c39b4c5107a5c2682e590db4c4885039ac341c2a/src/Tests/ClientConcepts/Serializer/CustomConstructorSerializerSettingsTests.cs#L50)
[How the factory implementation can combat it](https://github.com/elastic/elasticsearch-net/blob/c39b4c5107a5c2682e590db4c4885039ac341c2a/src/Tests/ClientConcepts/Serializer/FactorySerializerTests.cs)

src/Nest/CommonAbstractions/ConnectionSettings/ConnectionSettingsBase.cs
```diff
@@ -63,17 +68,36 @@ public abstract class ConnectionSettingsBase<TConnectionSettings> : ConnectionCo
...
 		}
 
+		protected ConnectionSettingsBase(
+			IConnectionPool connectionPool,
+			IConnection connection,
+			Func<TConnectionSettings, IElasticsearchSerializer> serializerFactoryFunc
+			)
+			: this(connectionPool, connection, null, serializerFactoryFunc) { }
+
+		IElasticsearchSerializer IConnectionSettingsValues.CreateStatefulSerializer(JsonConverter converter) =>
```
gmarz: For parity reasons, I think we should rename this to just `StatefulSerializer` (e.g.  `settings.Serializer` and `settings.StatefulSerializer`).

src/Nest/CommonAbstractions/SerializationBehavior/SerializerFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+		}
+
+		public IElasticsearchSerializer Create(IConnectionSettingsValues settings)
+		{
+			return this._serializerFactoryFunc?.Invoke(settings) ?? new JsonNetSerializer(settings);
+		}
+
+		public IElasticsearchSerializer CreateStateful(IConnectionSettingsValues settings, JsonConverter converter)
+		{
+			return new JsonNetSerializer(settings, converter);
```
gmarz: Nit: condense with lambda expression

src/Nest/CommonAbstractions/SerializationBehavior/SerializerFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+using Elasticsearch.Net;
+using Newtonsoft.Json;
+
+namespace Nest
+{
+	public interface ISerializerFactory
+	{
+		IElasticsearchSerializer Create(IConnectionSettingsValues settings);
+
+		IElasticsearchSerializer CreateStateful(IConnectionSettingsValues settings, JsonConverter converter);
```
gmarz: I'm wondering if we really need an additional method for this.  Can we just add `JsonConverter` as a nullable parameter on `Create`?  Stateful converters are really an internal thing, and I think it's best to "hide" it from the user.

gmarz:
@Mpdreamz this is great!  A few comments but otherwise LGTM.  Much better approach than #2189.

tsliang:
+1! I like that we're taking the improvements one level up to the ConnectionSettings and removing the Func parameter. LGTM!

src/Nest/CommonAbstractions/ConnectionSettings/ConnectionSettingsBase.cs
```diff
@@ -63,17 +68,36 @@ public abstract class ConnectionSettingsBase<TConnectionSettings> : ConnectionCo
 		private readonly FluentDictionary<MemberInfo, IPropertyMapping> _propertyMappings = new FluentDictionary<MemberInfo, IPropertyMapping>();
 		FluentDictionary<MemberInfo, IPropertyMapping> IConnectionSettingsValues.PropertyMappings => _propertyMappings;
 
-		protected ConnectionSettingsBase(IConnectionPool connectionPool, IConnection connection, Func<TConnectionSettings, IElasticsearchSerializer> serializerFactory)
-			: base(connectionPool, connection, serializerFactory)
+		private readonly ISerializerFactory _serializerFactory;
+		ISerializerFactory IConnectionSettingsValues.SerializerFactory { get; }
```
russcam: Should line 72 be

```c#
ISerializerFactory IConnectionSettingsValues.SerializerFactory => _serializerFactory;
```

src/Nest/CommonAbstractions/SerializationBehavior/SerializerFactory.cs
russcam: If they're only an internal implementation detail, I think we shouldn't expose a public method contract for creating them - maybe split this interface into two (SOL**I**D), `ISerializerFactory` and `IStatefulSerializerFactory`, the latter being a non-public interface?

src/Nest/CommonAbstractions/SerializationBehavior/JsonNetSerializer.cs
```diff
@@ -15,41 +15,64 @@ public class JsonNetSerializer : IElasticsearchSerializer
...
 		public JsonNetSerializer(IConnectionSettingsValues settings) : this(settings, null) { }
 
 		/// <summary>
-		/// this constructor is only here for stateful (de)serialization 
+		/// this constructor is only here for stateful (de)serialization
 		/// </summary>
-		internal JsonNetSerializer(IConnectionSettingsValues settings, JsonConverter stateFullConverter)
+		protected internal JsonNetSerializer(
+			IConnectionSettingsValues settings,
+			JsonConverter stateFullConverter
```
russcam: nitpick: `statefulConverter`

src/Nest/CommonAbstractions/SerializationBehavior/JsonNetSerializer.cs
```diff
@@ -15,41 +15,64 @@ public class JsonNetSerializer : IElasticsearchSerializer
...
 			var indentedSerializer = JsonSerializer.Create(this.CreateSettings(SerializationFormatting.Indented));
-			//indentedSerializer.Formatting = Formatting.Indented; 
 			this._defaultSerializers = new Dictionary<SerializationFormatting, JsonSerializer>
 			{
 				{ SerializationFormatting.None, this._defaultSerializer },
 				{ SerializationFormatting.Indented, indentedSerializer }
 			};
 		}
 
+		protected void CreateCustomJsonSerializers(Action<JsonSerializerSettings, IConnectionSettingsValues> s)
```
russcam: nitpick: can this have a more meaningful name e.g. settingsModifier? I think an XML comment would also really help here

russcam:
This looks good on first pass, good stuff @Mpdreamz :+1: 

We'll need to [improve the serialization documentation](https://www.elastic.co/guide/en/elasticsearch/client/net-api/2.x/connecting.html#_overriding_default_json_net_behavior) to explain how to use this, perhaps moving it out into a separate section in the NEST docs :smile:

src/Nest/CommonAbstractions/SerializationBehavior/SerializerFactory.cs
Mpdreamz: I chose to explicitly expose this nasty quality of our serializer here, anyone who implements the factory but does not support the stateful method is not fully adhering to the contract and might get bit (as per the linked multiget issues). To that end i do not think its more solid to split these, these two coexist in a pair.

src/Nest/CommonAbstractions/ConnectionSettings/ConnectionSettingsBase.cs
Mpdreamz: good catch :+1:

Mpdreamz:
Addressed comments and merged to `2.x` and `master`

jonyadamit:
I'm sorry I don't get it.
Previously I had:
```
public class SomeCustomSerializer : JsonNetSerializer
	{
		protected override void ModifyJsonSerializerSettings(JsonSerializerSettings settings)
		{
			base.ModifyJsonSerializerSettings(settings);
			settings.PreserveReferencesHandling = PreserveReferencesHandling.All;   //for instance.. (no pun intended :))
		}
	}
```
How should this simple thing be implemented with the new approach?

jonyadamit:
@Mpdreamz , @russcam any thoughts?

russcam:
The following in 2.4.5 will work

```c#
void Main()
{
	var pool = new SingleNodeConnectionPool(new Uri("http://localhost:9200"));
	var connectionSettings = new ConnectionSettings(
		pool, 
		new HttpConnection(), 
		new SerializerFactory(s => new SomeCustomSerializer(s)));
				
	var client = new ElasticClient(connectionSettings);
}

public class SomeCustomSerializer : JsonNetSerializer
{
	public SomeCustomSerializer(IConnectionSettingsValues settings) 
		: base(settings)
	{
		
	}
	
	protected override void ModifyJsonSerializerSettings(JsonSerializerSettings settings)
	{
		base.ModifyJsonSerializerSettings(settings);
		settings.PreserveReferencesHandling = PreserveReferencesHandling.All;   
		//for instance.. (no pun intended :))
	}
}
```

jonyadamit:
@russcam  thanks. Would work, but issues an obsolete warning. Was that method unintentionally decorated?

russcam:
I think this is the better way

```c#
void Main()
{
	var pool = new SingleNodeConnectionPool(new Uri("http://localhost:9200"));
	var connectionSettings = new ConnectionSettings(
		pool, 
		new HttpConnection(), 
		new SerializerFactory(s => new SomeCustomSerializer(s)));
				
	var client = new ElasticClient(connectionSettings);
}

public class SomeCustomSerializer : JsonNetSerializer
{
	public SomeCustomSerializer(IConnectionSettingsValues settings) 
		: base(settings)
	{
		OverwriteDefaultSerializers(ModifyJsonSerializerSettings);
	}

	private void ModifyJsonSerializerSettings(JsonSerializerSettings settings, IConnectionConfigurationValues connectionSettings)
	{
		settings.PreserveReferencesHandling = PreserveReferencesHandling.All;   
		//for instance.. (no pun intended :))
	}
}
```

This method was introduced to allow state to be passed in the derived serializer constructor, to modify either `JsonSerializerSettings` or `IConnectionConfigurationValues`

/cc @Mpdreamz, @gmarz

jonyadamit:
How would you feel about having a constructor overload for `SerializerFactory` that would take this "settings modifier action"?
If all I want is to change that sort of setting, it would change my code from this:
```
void Main()
{
    var pool = new SingleNodeConnectionPool(new Uri("http://localhost:9200"));
    var connectionSettings = new ConnectionSettings(
        pool, 
        new HttpConnection(), 
        new SerializerFactory(s => new SomeCustomSerializer(s)));

    var client = new ElasticClient(connectionSettings);
}

public class SomeCustomSerializer : JsonNetSerializer
{
    public SomeCustomSerializer(IConnectionSettingsValues settings) 
        : base(settings)
    {
        OverwriteDefaultSerializers(ModifyJsonSerializerSettings);
    }

    private void ModifyJsonSerializerSettings(JsonSerializerSettings settings, IConnectionConfigurationValues connectionSettings)
    {
        settings.PreserveReferencesHandling = PreserveReferencesHandling.All;   
        //for instance.. (no pun intended :))
    }
}
```
to that:
```
void Main()
{
    var pool = new SingleNodeConnectionPool(new Uri("http://localhost:9200"));
    var connectionSettings = new ConnectionSettings(
        pool, 
        new HttpConnection(), 
        new SerializerFactory(s => s.PreserveReferencesHandling = PreserveReferencesHandling.All));

    var client = new ElasticClient(connectionSettings);
}
```
:sweat_smile:

Mpdreamz:
@jonyadamit looking forward to a PR 😉

jonyadamit:
Gladly :smile:   #2247
I'm very uncomfortable with the `OverwriteDefaultSerializers` method though.
If we create the settings and serializers twice, something here is inherently wrong..
I can see why you did it this way though, and without going too deep in the code I have a feeling that avoiding this would require a complete overhaul of `JsonNetSerializer` and it's dependents..

Mpdreamz:
Thanks @jonyadamit merged and ported to master. 

The call to `OverwriteDefaultSerializers` does not sit well with me either but given all alternative attempts (see original PR comment at the top) relied on threading constructs its the lesser of two evils. 

Zooming wayyyyy back out our long term plan to fix this is to generate **all** *our* json reads and writes and only offload to JSON.NET for user content (documents/fields).  

We have something in the works that codifies all elasticsearch requests and responses using typescript and we intend to use that as a compilable source of truth from which we can generate our C# classes and readers/writers from a stream. This is all POC and not a company wide effort (yet).

Spijkerboer:
The SerializerFactory doesn't seem to work for me on 2.5.2.
I tried to following to get rid of the timezone while serializing. 

`var config = new ConnectionSettings(pool, new HttpConnection(),
new SerializerFactory((jsonSettings, nestSettings) =>
		jsonSettings.DateTimeZoneHandling = DateTimeZoneHandling.Unspecified));`

When I look with Fiddler I see that the +01:00 is being added.
In a consoleapplication with just Newtonsoft.Json the timezone indication issn't there:


```
JsonSerializerSettings settings = new JsonSerializerSettings()
{
  DateTimeZoneHandling = DateTimeZoneHandling.Unspecified
};

var now = DateTime.Now;
Console.WriteLine(JsonConvert.SerializeObject(now, settings));
```

I tried to two CustomSerializer examples from this issue too, those didn't work for me.

PR #2199, opened 2016-08-05
_source false  issue solved with Union<>

pathikdevani:
when set source false, Nest consider as exclude:["*"]. it's effect on performance.

i solve using Union<>.


karmi:
Hi @pathikdevani, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/2199.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

pathikdevani:
Hi @karmi, #2199 commit is create by my Github account. I am using Github on Visual Studio 2015 and  i have only one common email address.

pathikdevani:
@karmi i find my mistake, problem with visual studio. so i close this PR and create new from my email.

PR #2200, opened 2016-08-05
solved _source false issue using Union<>

pathikdevani:
when source to set false, Nest consider exclude["*"]. it's effect on performance.

i solve this using Union<>.

### DIFF
--- a/src/Nest/Aggregations/Metric/TopHits/TopHitsAggregation.cs
+++ b/src/Nest/Aggregations/Metric/TopHits/TopHitsAggregation.cs
@@ -18,7 +18,7 @@ namespace Nest
 		IList<ISort> Sort { get; set; }
 
 		[JsonProperty("_source")]
-		ISourceFilter Source { get; set; }
+        Union<bool, ISourceFilter> Source { get; set; }
 
 		[JsonProperty("highlight")]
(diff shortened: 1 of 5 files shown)


russcam:
Hey @pathikdevani, we maintain backwards binary compatibility within major versions of NEST and  unfortunately this solution would break backwards binary compatibility (bwc) for `2.x`; We would need to come up with another way of handling this for `2.x` that maintains bwc.

This fix would be great to get into `master` though that is tracking against Elasticsearch 5.x; would you be able to close this PR and retarget the change against `master`?

pathikdevani:
@russcam thanks for replay. 

It is very serious issue when we use aggregation. If we can add change in 2.x then it's very useful.
Anything i can do for add change in 2.x?

And I create new PR for master branch (5.x).

gmarz:
@pathikdevani I just opened the above PR (#2223) which proposes a fix that's backwards compatible with 2.x, which will likely make it into the next release.

I'll work on incorporating your PR into master.  Thanks for raising this!

gmarz:
@pathikdevani this is now pushed to master with the above commit.  Thanks again for the PR!

pathikdevani:
Thank you @gmarz

PR #2201, opened 2016-08-08
Move Analyze parameters to json body

### LINKED ISSUE #2169: Analyze operation legacy request
**Describe the feature**:

When an analyze operation is performed, the request generated to ES contains the text to analyze as the body of the request and the rest of parameters in the query string. This is maintained by ES for backward compatibility and has limitations (text cannot start by '{') as described here: https://www.elastic.co/guide/en/elasticsearch/reference/2.3/indices-analyze.html. I have looked in your code base and seen this is done in: ElasticClient (https://github.com/elastic/elasticsearch-net/blob/2.x/src/Nest/Indices/Analyze/ElasticClient-Analyze.cs).

I would suggest to pass all the parameters in the body of the request as json as described in the above ES documentation link.

russcam:
Closes #2169

### DIFF
--- a/src/Elasticsearch.Net/Domain/RequestParameters/AnalyzeRequestParameters.cs
+++ b/src/Elasticsearch.Net/Domain/RequestParameters/AnalyzeRequestParameters.cs
@@ -1,0 +1,39 @@
+﻿using System;
+
+namespace Elasticsearch.Net
+{
+	public partial class AnalyzeRequestParameters
+	{
+		///<summary>The name of the analyzer to use</summary>
+		[Obsolete("Deprecated. Specify the analyzer to use in the body of the request.")]
+		public AnalyzeRequestParameters Analyzer(string analyzer) => this;
+
+		///<summary>Deprecated : A comma-separated list of character filters to use for the analysis</summary>
+		[Obsolete("Deprecated. Specify the char filters to use in the body of the request.")]
+		public AnalyzeRequestParameters CharFilters(params string[] char_filters) => this;
+
+		///<summary>A comma-separated list of character filters to use for the analysis</summary>
+		[Obsolete("Deprecated. Specify the char filters to use in the body of the request.")]
+		public AnalyzeRequestParameters CharFilter(params string[] char_filter) => this;
+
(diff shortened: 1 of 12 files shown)


src/Elasticsearch.Net/Domain/RequestParameters/RequestParameters.Generated.cs
```diff
@@ -1613,30 +1613,6 @@ public class AnalyzeRequestParameters : FluentRequestParameters<AnalyzeRequestPa
 	{
 		public override HttpMethod DefaultHttpMethod => HttpMethod.POST;
 		
-		///<summary>The name of the analyzer to use</summary>
-		public AnalyzeRequestParameters Analyzer(string analyzer) => this.AddQueryString("analyzer", analyzer);
-		
-
```
gmarz: Removing these methods breaks bwc :\

gmarz:
This LGTM for `master`, but not backwards compatible for `2.x`.  Can we come up with a way to mark the methods on `AnalyzRequestParameters` as obsolete?

src/Elasticsearch.Net/Domain/RequestParameters/RequestParameters.Generated.cs
russcam: Thanks for taking a look @gmarz. I've changed the generated `RequestParameters` to be partial classes so that I can patch these methods back in. Thoughts?

gmarz: LGTM :+1:

src/Elasticsearch.Net/Domain/RequestParameters/AnalyzeRequestParameters.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+
+namespace Elasticsearch.Net
+{
+	public partial class AnalyzeRequestParameters
+	{
+		///<summary>The name of the analyzer to use</summary>
+		[Obsolete("Deprecated. Specify the analyzer to use in the body of the request.")]
```
Mpdreamz: :+1: binary bwc is a pain.

Mpdreamz:
LGTM

Issue #2194, opened 2016-08-04
GetAlias ignores the alias parameter in descriptor


plmaheu:
**NEST/Elasticsearch.Net version**:
2.4.2

**Elasticsearch version**:
2.3.5

**Description of the problem including expected versus actual behavior**:

This overload is ignoring the alias name parameter.
`IGetAliasesResponse GetAlias(Func<GetAliasDescriptor, IGetAliasRequest> selector = null)`
Was working before in 1.7.2.

I'm getting /_alias in url while it should be /_alias/somealiasname

This overload, `IGetAliasesResponse GetAlias(IGetAliasRequest request)`, works as expected.

**Steps to reproduce**:
 1. `elasticClient.GetAlias(x => x.Alias("testalias"));`
 2. Look at request url (I sniffed requests using Fiddler). Will be /_alias instead of /_alias/testalias

**Provide `ConnectionSettings` (if relevant)**:

**Provide `DebugInformation` (if relevant)**:

russcam:
Thanks for reporting @plmaheu! The following will do what you need, using `.Name(Names name)`

```c#
client.GetAlias(a => a
    .Name("alias-name")
);
```

which yields

```
GET http://localhost:9200/_alias/alias-name
```

It looks like `.Alias(string alias)` is an artifact of the 1.x refactoring and should have been removed. I'll mark it as deprecated and map it to the same route value as `Name` so that it still works in 2.x. I'll remove it in `master`, tracking 5.x

russcam:
[I've just merged a PR into 2.x to address this](https://github.com/elastic/elasticsearch-net/pull/2203); it will go into the next release.

Thanks again fr reporting @plmaheu :+1:

PR #2203, opened 2016-08-09
Deprecate Alias on GetAlias and GetAliases

russcam:
Closes https://github.com/elastic/elasticsearch-net/issues/2194

### DIFF
--- a/src/Nest/Indices/AliasManagement/GetAlias/ElasticClient-GetAlias.cs
+++ b/src/Nest/Indices/AliasManagement/GetAlias/ElasticClient-GetAlias.cs
@@ -5,12 +5,10 @@ using Elasticsearch.Net;
 
 namespace Nest
 {
-	using GetAliasesConverter = Func<IApiCallDetails, Stream, GetAliasesResponse>;
-
 	public partial interface IElasticClient
 	{
 		/// <summary>
-		/// The get index alias api allows to filter by alias name and index name. This api redirects to the master and fetches 
+		/// The get index alias api allows to filter by alias name and index name. This api redirects to the master and fetches
 		/// the requested index aliases, if available. This api only serialises the found index aliases.
 		/// <para> Difference with GetAlias is that this call will also return indices without aliases set</para>
 		/// <para> </para>http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/indices-aliases.html#alias-retrieving
@@ -35,10 +33,10 @@ namespace Nest
 			this.GetAlias(selector.InvokeOrDefault(new GetAliasDescriptor()));
 
 		/// <inheritdoc/>
-		public IGetAliasesResponse GetAlias(IGetAliasRequest request) => 
+		public IGetAliasesResponse GetAlias(IGetAliasRequest request) =>
 			this.Dispatcher.Dispatch<IGetAliasRequest, GetAliasRequestParameters, GetAliasesResponse>(
 				request,
-				new GetAliasesConverter(DeserializeGetAliasesResponse),
+				DeserializeGetAliasesResponse,
 				(p, d) => this.LowLevelDispatch.IndicesGetAliasDispatch<GetAliasesResponse>(p)
 			);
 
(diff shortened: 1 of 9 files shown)


Mpdreamz:
LGTM :+1:

russcam:
Will _port_ to `master`

tabro:
@gmarz @russcam @Mpdreamz I believe that this PR may have caused an issue with NEST throwing a deserilazation error when performing the following code on a non-existing alias when providing AllowedStatusCodes.:

 ```
await ElasticClient.GetAliasAsync(d => d
                                .RequestConfiguration(r => r
                                .AllowedStatusCodes((int)HttpStatusCode.NotFound)
                                .CancellationToken(CancellationToken.None))
                   .Name("does-not-exist"))
                   .ConfigureAwait(false);
```

Removing the `.AllowedStatusCode(404)` makes NEST correctly return an empty list of aliases. In previous versions this would throw an exception as expected.

**Comment by tabro** [comment]
Stacktrace:

```
Elasticsearch.Net.UnexpectedElasticsearchClientException : Error converting value "alias [does-not-exist] missing" to type 'System.Collections.Generic.Dictionary`2[System.String,System.Collections.Generic.Dictionary`2[System.String,Nest.AliasDefinition]]'. Path 'error', line 1, position 41.
  ----> Newtonsoft.Json.JsonSerializationException : Error converting value "alias [does-not-exist] missing" to type 'System.Collections.Generic.Dictionary`2[System.String,System.Collections.Generic.Dictionary`2[System.String,Nest.AliasDefinition]]'. Path 'error', line 1, position 41.
  ----> System.ArgumentException : Could not cast or convert from System.String to System.Collections.Generic.Dictionary`2[System.String,System.Collections.Generic.Dictionary`2[System.String,Nest.AliasDefinition]].
   at Elasticsearch.Net.Transport`1.<RequestAsync>d__15`1.MoveNext() in C:\Users\russ\source\elasticsearch-net-2.x\src\Elasticsearch.Net\Transport\Transport.cs:line 150
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Nest.ElasticClient.<Nest-IHighLevelToLowLevelDispatcher-DispatchAsync>d__220`4.MoveNext() in C:\Users\russ\source\elasticsearch-net-2.x\src\Nest\ElasticClient.cs:line 0
```

**Comment by russcam** [comment]
@tabro do you have an example json response where this throws?

**Comment by russcam** [comment]
@tabro I've looked into this further. This change didn't introduce the error that you're seeing, that throws an exception on the commit c5206075fb2701ab61251f1e7f1a9b5da9a7e219 prior to this change.

By allowing a status code in `.AllowedStatusCodes()`, you're telling NEST that if the status code comes back as one in the range `>= 200` and `< 300` or as one in the allowed status codes, then the response is _valid_ and the call is _successful_. At this point, NEST will deserialize the body of the response into the strong type for the request made, which in this case is `IGetAliasesResponse`. This throws an exception as the response looks like

```
HTTP/1.1 404 Not Found
Content-Type: application/json; charset=UTF-8
Content-Length: 67

{
  "error" : "alias [does-not-exist] missing",
  "status" : 404
}
```
and the way alias responses are handled is to [first deserialize to a rather verbose `Dictionary<string, Dictionary<string, Dictionary<string, AliasDefinition>>>` structure.](https://github.com/elastic/elasticsearch-net/blob/2.x/src/Nest/Indices/AliasManagement/GetAliases/ElasticClient-GetAliases.cs#L75)

I'd recommend not setting `.AllowedStatusCodes((int)HttpStatusCode.NotFound)` then either

- check `.IsValid` on the response to know if the response is valid, optionally checking the status code for 404 if this is particular response you want to handle
- set `.ThrowExceptions()` on `ConnectionSettings` and handle the thrown exception in the case of a 404.

PR #2204, opened 2016-08-11
Allow for modifying HttpClientHandler on .NET Core

### LINKED ISSUE #2198: HTTPS Cert Validation Callback in .Net Core
In the .net framework version, it is allowed to check the HTTPS cert with method overriding.
```csharp
 public class HttpsConnection : HttpConnection
    {
        protected override HttpWebRequest CreateHttpWebRequest(RequestData requestData)
        {
            var request = base.CreateHttpWebRequest(requestData);
            request.ServerCertificateValidationCallback = (sender, cert, chain, error) =>
            {
 /// perform the checking on the self-signed cert
                return true;
            };
            return request;
        }
    }
```
The .Net team has changed the code now. It is handled in HttpClientHandler. Is it possible do the similar way in .Net Core version? It seems this is the only way to check cert chain manually when we do not have permission to install the cert in the server.

Ref:
https://github.com/dotnet/corefx/issues/4476
https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/HttpClientHandler.Windows.cs#L167
https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.cs#L1511

gmarz:
This allows for setting things like the `ServerCertificateValidationCallback` on Core CLR which previously wasn't supported:

```csharp
public class SecureHttpConnection : HttpConnection
{
    protected override HttpClientHandler CreateHttpClientHandler(RequestData requestData)
    {
        var handler = base.CreateHttpClientHandler(requestData);
        handler.ServerCertificateCustomValidationCallback = (sender, cert, chain, errors) => true;
        return handler;
    }
}
```

Closes #2198

### DIFF
--- a/src/Elasticsearch.Net/Connection/HttpConnection-CoreFx.cs
+++ b/src/Elasticsearch.Net/Connection/HttpConnection-CoreFx.cs
@@ -6,6 +6,8 @@ using System.IO.Compression;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


russcam:
LGTM :+1:

PR #2221, opened 2016-08-22
Fix for sniffing throwing error for multiple threads.

lostllama:
In using ES, I found that the following scenario could throw an error, and it seems to be down to the way the semaphores are handled. This is the scenario:

1) Thread A blocks the semaphore, sniffs, and releases it.
2) Thread B waits for the semaphore, A releases, B blocks, and then, not needing to sniff, simply returns.
3) Thread C waits for the semaphore, A releases, B blocks, and then C times out because B never released it.

My solution is two-fold:

1) Thread B will release the semaphore, so that C can enter.
2) In case Thread C times out before B can release, it checks if sniffing is needed before throwing the exception, and, if it isn't, it simply returns (as opposed to throwing the exception).

### DIFF
--- a/src/Elasticsearch.Net/Transport/Pipeline/RequestPipeline.cs
+++ b/src/Elasticsearch.Net/Transport/Pipeline/RequestPipeline.cs
@@ -121,9 +121,17 @@ namespace Elasticsearch.Net
 		public void FirstPoolUsage(SemaphoreSlim semaphore)
 		{
 			if (!this.FirstPoolUsageNeedsSniffing) return;
-			if (!semaphore.Wait(this._settings.RequestTimeout))
-				throw new PipelineException(PipelineFailure.CouldNotStartSniffOnStartup, (Exception)null);
-			if (!this.FirstPoolUsageNeedsSniffing) return;
+            if (!semaphore.Wait(this._settings.RequestTimeout))
+            {
+                if (this.FirstPoolUsageNeedsSniffing)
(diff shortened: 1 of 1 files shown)


russcam:
@lostllama this looks good; would you be able to add a test as well?

lostllama:
Hi Russ

Maybe this evening :) I'm at work at the moment and beyond fixing the issue
(which was causing us a problem), I'm not sure they'd like me working on
tests in work time. I'll see what I can do when I get home.

John

karmi:
Hi @lostllama, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/2221.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

lostllama:
@karmi I've made my e-mail address visible. It is the same e-mail address, however.

lostllama:
@karmi I've done that now.

russcam:
@lostllama I'll look at adding some tests for this; we have some in this area but looks like it didn't catch this.

russcam:
I've added a test to verify that no pipeline exception is thrown, indicating that Thread B has released the `SemaphoreSlim`.

Thanks again for the PR, we're looking to put a release out this week so it will be included in that.

lostllama:
@russcam Thanks :)

russcam:
[This is now on nuget in 2.4.4](https://www.nuget.org/packages/NEST/2.4.4)

PR #2223, opened 2016-08-23
Backwards compatible solution for #2200

gmarz:
SourceFilter now has a Disable property which can be used to enable or
disable _source.  Instances that accepted Source(bool) now set this
property rather than ExcludeAll (which doesn't truly disable _source).

This is more of a workaround to maintain BWC in 2.x  In master, we will
use Union<bool, ISourceFilter> instead as proposed in #2200.


russcam:
Sad that it's a [`Union<bool, ISourceFilter>` on `IMultiGetRequest`](https://github.com/elastic/elasticsearch-net/blob/2.x/src/Nest/Document/Multiple/MultiGet/Request/MultiGetOperation.cs#L25) :sad:

russcam:
LGTM :+1:

PR #2233, opened 2016-09-01
Support for get task API

gmarz:
Adds support for GET /_tasks/{taskid}

Also renamed TasksList and TasksCancel for consistency.


src/Benchmarking/Tester.cs
```diff
@@ -137,7 +137,7 @@ protected IndexResults GenerateAndIndex(int numMessages, int bufferSize)
 					})
 				).ToArray();
 
-			Task.WaitAll(array);
+			System.Threading.Tasks.Task.WaitAll(array);
```
russcam: The bane of using a common noun :smile:

russcam:
LGTM :+1:

PR #2235, opened 2016-09-04
fix #2209 Several improvements in some edgecase querydsl scenarios

### LINKED ISSUE #2209: StackOverflowException when combining many bool queries
Repro here https://github.com/elastic/elasticsearch-net/commit/fe4a52ea075eb7a58eea06911996c280f1f52ede

Starts to happen after ~1000 iterations.  We should be able to support `max_clause_count`.

Mpdreamz:
We always flattened situations like:

`query1 && query2 && query3` into a single bool query with 3 must clauses.

But if the queries are `boolean` themselves that can't be merged we would
generate deeply nested pairs of two.

For instance we have 3 bool queries with only should clauses

`bq1 && bq2 && bq3`

ideally we'd generate:

```
bool:
  must:
     bool: should:,
     bool: should:,
     bool: should:
```

However due to the bitwise binary nature the way they were combined was:

Step1:

`bq1 && b2` - both sides only have should clauses merging them would be
wrong so we wrap to `result1`

```
bool:
  must:
     bool: should:,
     bool: should:
```

now we do `result1 && bq3` since result1 is a must we can not merge the
should from `bq3` on it and thus we need to wrap to `result2`

```
bool:
  must:
    bool:
      must:
        bool: should //bq1
        bool: should //bq2
    bool: should: //bq3
```

This is still semantically correct but now imagine you are doing this
in a loop combining thousands of these. Admittingly an edgecase it would
cause a nasty stackoverflow or in the best case query graphs that are unnecessary too deep
as is showcased by #2209

This is now fixed in a way that does not need constant traversal of to
see if a bool must or should clause consists of only bools.

Similarly the same happened if you'd || or a bunch of boolean queries only
containing musts.

A third fix is that we no longer modify the original query when stealing
clauses. This could cause nasty errors when reusing a static bool query
e.g

```csharp
var q = Query<Project>.Bool(b => b.Should(ATermQuery).Must(ATermQuery));
var c = new QueryContainer();
for(int i = 0;i<1000;i++) c &= q;
```

The `&` or `|` bitwise operators were not pure and could potentially steal clauses from
`q` and adding them to the new result. Causing problems when we want to
apply q again and again and again.

### DIFF
--- a/docs/query-dsl/bool-dsl/bool-dsl.asciidoc
+++ b/docs/query-dsl/bool-dsl/bool-dsl.asciidoc
@@ -364,3 +364,55 @@ Assert(
     c => AssertDoesNotJoinOntoLockedBool(c, "leftBool"));
 ----
 
+=== Perfomance considerations
+
+If you have a requirement of combining many many queries using the bool dsl please take the following into account.
+
+You *can* use bitwise assignments in a loop to combine many queries into a bigger bool.
+
+NOTE: This needs NEST 2.4.6 or higher. Prior versions did not do a good job flattening the result in the most optimal way and could cause a stackoverflow when doing ~2000 iterations In later version you can combine as many as you'd like but please keep mind of the following performance recommendations
+
+In this example we are creating a single bool query with a 1000 must clauses using the `&=` assign operator.
+
+[source,csharp]
+----
+var c = new QueryContainer();
+
+var q = new TermQuery { Field = "x", Value = "x" };
+
+c &= q;
+----
+
+As we can see while still fast its causes a lot of allocations to happen because on each iteration we need to re evaluate
+the mergability of our bool query.
+
+....
+|===
+|    Method|     Median|     StdDev|       Gen 0|  Gen 1|  Gen 2|  Bytes Allocated/Op
+|  Allocate|  1.8507 ms|  0.1878 ms|    1,793.00|  21.00|      -|        1.872.672,28
+|===
+....
+
+Since we already know the shape of our bool query in advance its much much faster to do this instead:
+
+[source,csharp]
+----
+QueryContainer q = new TermQuery { Field = "x", Value = "x" };
+
+var x = Enumerable.Range(0, 1000).Select(f => q).ToArray();
+
+var boolQuery = new BoolQuery
+{
+    Must = x
+};
+----
+
+ The drop both in performance and allocations is tremendous!
+
+....
+|===
+|    Method|      Median|     StdDev|   Gen 0|  Gen 1|  Gen 2|  Bytes Allocated/Op
+|  Allocate|  31.4610 us|  0.9495 us|  439.00|      -|      -|            7.912,95
+|===
+....
+
--- a/docs/query-dsl/bool-dsl/bool-dsl.asciidoc
+++ b/docs/query-dsl/bool-dsl/bool-dsl.asciidoc
@@ -370,8 +370,6 @@ If you have a requirement of combining many many queries using the bool dsl plea
 
 You *can* use bitwise assignments in a loop to combine many queries into a bigger bool.
 
-NOTE: This needs NEST 2.4.6 or higher. Prior versions did not do a good job flattening the result in the most optimal way and could cause a stackoverflow when doing ~2000 iterations In later version you can combine as many as you'd like but please keep mind of the following performance recommendations
-
 In this example we are creating a single bool query with a 1000 must clauses using the `&=` assign operator.
 
 [source,csharp]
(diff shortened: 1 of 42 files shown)


Mpdreamz:
**Machine info**

```ini
Host Process Environment Information:
BenchmarkDotNet.Core=v0.9.9.0
OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i7-4750HQ CPU 2.00GHz, ProcessorCount=8
Frequency=1948624 ticks, Resolution=513.1826 ns, Timer=TSC
CLR=MS.NET 4.0.30319.42000, Arch=32-bit RELEASE
GC=Concurrent Workstation
JitModules=clrjit-v4.6.2001.0

Type=BooleanAllocationBenchmark  Mode=Throughput  LaunchCount=3  
WarmupCount=5  TargetCount=10  Type=BooleanAllocationBenchmark  
Mode=Throughput  LaunchCount=3  WarmupCount=5  
TargetCount=10  
```


# Comparing `&=`:

```
var c = new QueryContainer();
QueryContainer q =new TermQuery { Field = "x", Value = "x" }; 
for (var i=0;i<1000;i++)
	c &= q;
```

### Latest released version (2.4.5) 

   Method |   Method |    Median |    StdDev |    Median |    StdDev |    Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
--------- |--------- |---------- |---------- |---------- |---------- |--------- |------ |------ |------------------- |
 Allocate | Allocate | 2.8882 ms | 0.4175 ms | 2.8882 ms | 0.4175 ms | 3,516.00 | 83.00 |     - |       3.308.101,43 |


### This branch:

   Method |   Method |    Median |    StdDev |    Median |    StdDev |    Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
--------- |--------- |---------- |---------- |---------- |---------- |--------- |------ |------ |------------------- |
 Allocate | Allocate | 1.6182 ms | 0.3165 ms | 1.6182 ms | 0.3165 ms | 1,681.00 | 49.00 |     - |       1.684.119,54 |



# Comparing `|=`:

```
var c = new QueryContainer();
QueryContainer q =new TermQuery { Field = "x", Value = "x" }; 
for (var i=0;i<1000;i++)
	c |= q;
```

### Latest released version (2.4.5) 

   Method |   Method |    Median |    StdDev |    Median |    StdDev |    Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
--------- |--------- |---------- |---------- |---------- |---------- |--------- |------ |------ |------------------- |
 Allocate | Allocate | 2.4455 ms | 0.4268 ms | 2.4455 ms | 0.4268 ms | 3,482.00 | 83.00 |     - |       3.824.267,35 |

### This branch:

   Method |   Method |    Median |    StdDev |    Median |    StdDev |    Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
--------- |--------- |---------- |---------- |---------- |---------- |--------- |------ |------ |------------------- |
 Allocate | Allocate | 1.6811 ms | 0.1959 ms | 1.6811 ms | 0.1959 ms | 1,832.00 | 16.00 |     - |       1.652.231,83 |

Mpdreamz:
This PR also adds some documentation around performance of the bool query dsl, if you know you need a single bool with a lot of must clauses doing this:

```
QueryContainer q = new TermQuery { Field = "x", Value = "x" };
var x = Enumerable.Range(0, 1000).Select(f => q).ToArray();
var boolQuery = new BoolQuery
{
	Must = x
};
```

Is going to peform muuuuuch better which is not that crazy considering we need to reevaluate the bools merge-ability on each bitwise operation.

   Method |   Method |     Median |    StdDev |     Median |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
--------- |--------- |----------- |---------- |----------- |---------- |------- |------ |------ |------------------- |
 Allocate | Allocate | 31.2520 us | 0.4767 us | 31.2520 us | 0.4767 us | 439.00 |     - |     - |           7.737,15 |

Mpdreamz:
This PR also fixes a small difference in behaviour in the initializer syntax between:

Our documented approach

```csharp
var search = new SearchRequest 
{
	Query = new BoolQuery {}
}
```

And undocumented but still valid:

```csharp
var search = new SearchRequest 
{
	Query = new QueryContainer(new BoolQuery {})
}
```

The implicit conversion was in charge of carrying some characteristics (IsStrict/IsVerbatim) from `QueryBase` over to `QueryContainer` which caused problems if you used the constructor directly.

docs/query-dsl/bool-dsl/bool-dsl.asciidoc
```diff
@@ -364,3 +364,55 @@ Assert(
...
 ----
 
+=== Perfomance considerations
+
+If you have a requirement of combining many many queries using the bool dsl please take the following into account.
+
+You *can* use bitwise assignments in a loop to combine many queries into a bigger bool.
+
+NOTE: This needs NEST 2.4.6 or higher. Prior versions did not do a good job flattening the result in the most optimal way and could cause a stackoverflow when doing ~2000 iterations In later version you can combine as many as you'd like but please keep mind of the following performance recommendations
+
```
Mpdreamz: This needs to be updated to specifically mention we did not do a good job combining many bools this way, other queries were flattened correctly.

gmarz:
LGTM :+1: Awesome work @Mpdreamz !

src/Nest/CommonAbstractions/Extensions/Extensions.cs
```diff
@@ -169,8 +188,8 @@ internal static void ForEach<T>(this IEnumerable<T> enumerable, Action<T> handle
 
 		internal static List<T> ToListOrNullIfEmpty<T>(this IEnumerable<T> enumerable)
 		{
-			var list = enumerable?.ToList();
-			return list.HasAny() ? list : null;
+			var list = enumerable.AsInstanceOrToListOrNull();
+			return list == null || list.Count > 0 ? list : null;
```
Mpdreamz: 👍 no

Mpdreamz: actually it is correct :) if `list == null || list.Count > 0` = `list != null && list.Count > 0` has the same effect here. Changing it to the latter for readability

PR #2249, opened 2016-09-13
fix #2119 allow custom per field similarity in the mapping other then…

### LINKED ISSUE #2119: Per field custom similarity in fluent PUT mapping
Nest version: 2.1.1

The current fluent syntax for creating a PUT mapping forces you to use a built in `SimilarityOption` (either `Default` or `BM25`) when specifying a per field similarity (https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Mapping/Types/CorePropertyDescriptorBase.cs), but I would like to use a custom similarity specified in the create index request instead. As far as I can tell this is currently only possible in Nest using `AllFieldDescriptor.Similarity`, but I don't want to apply the custom similarity to the `_all` field. According to the [Elastic docs](https://www.elastic.co/guide/en/elasticsearch/guide/current/changing-similarities.html) a per field custom similarity should be valid.

Mpdreamz:
Tad hacky binary bwc fix but this allows you to specify a per field custom similarity in a fields mapping. 

The put mapping and create settings already allowed you to create custom similarities but you had no way to apply them. Super silly.

### DIFF
--- a/src/Nest/Mapping/AttributeBased/ElasticsearchPropertyAttributeBase.cs
+++ b/src/Nest/Mapping/AttributeBased/ElasticsearchPropertyAttributeBase.cs
@@ -18,7 +18,11 @@ namespace Nest
 		bool? IProperty.Store { get; set; }
 		bool? IProperty.DocValues { get; set; }
 		IProperties IProperty.Fields { get; set; }
-		SimilarityOption? IProperty.Similarity { get; set; }
+
+#pragma warning disable CS0618 // Type or member is obsolete
+		SimilarityOption? IProperty.Similarity { get { return Self.CustomSimilarity?.ToEnum<SimilarityOption>(); } set { Self.CustomSimilarity = value.GetStringValue(); } }
+#pragma warning restore CS0618 // Type or member is obsolete
+		string IProperty.CustomSimilarity { get; set; }
 		Fields IProperty.CopyTo { get; set; }
 
 		public string Name { get; set; }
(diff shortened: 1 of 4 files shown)


src/Nest/Mapping/AttributeBased/ElasticsearchPropertyAttributeBase.cs
```diff
@@ -18,14 +18,23 @@ public abstract class ElasticsearchPropertyAttributeBase : Attribute, IProperty,
...
 
 		public string Name { get; set; }
 		public bool Ignore { get; set; }
 		public bool DocValues { get { return Self.DocValues.GetValueOrDefault(); } set { Self.DocValues = value; } }
 		public string IndexName { get { return Self.IndexName; } set { Self.IndexName = value; } }
 		public SimilarityOption Similarity { get { return Self.Similarity.GetValueOrDefault(); } set { Self.Similarity = value; } }
+
+#pragma warning disable CS0618 // Type or member is obsolete
+		[Obsolete("This is a temporary binary backwards compatible hack to make sure you can specify named similarities in 2.x, scheduled for removal in 5.0")]
+		public string SimilarityHack { get { return Self.SimilarityHack; } set { Self.SimilarityHack = value; } }
```
gmarz: Maybe `CustomSimilarity` is a better name?

Mpdreamz: Sounds too official in my mind, by naming it a hack explicitly we clearly signal our patch is a hack that will get removed as soon as we can.

Mpdreamz: @elastic/microsoft ready for another review

PR #2250, opened 2016-09-13
fix #2119 in master, make similarity on per field mapping a string

### LINKED ISSUE #2119: Per field custom similarity in fluent PUT mapping

Mpdreamz:
Forward port of #2249 minus the hacky bits


gmarz:
LGTM

PR #2252, opened 2016-09-13
fix #2125 bwc fix to support specifying any highlighter as string not…

### LINKED ISSUE #2125: Allow to use a custom Highlighter
In Nest 2.3.2 the `HighlightField` uses an enum `HighlighterType` as `Type`.

This prevents to use a custom highlighter instead of one of the 3 embedded within ElasticSearch.

This is a regression from Nest 1.7 where the same field was a String, therefore allowing to put the name of an highlighter provided by a plugin.

I think having a string as Type field and a few const string inside a static `HighlighterType` class would give the best of both world.

Mpdreamz:
… just the three buildin ones

### DIFF
--- a/src/Nest/Search/Search/Highlighting/HighlightField.cs
+++ b/src/Nest/Search/Search/Highlighting/HighlightField.cs
@@ -47,8 +47,11 @@ namespace Nest
 		[JsonProperty("boundary_chars")]
 		string BoundaryChars { get; set; }
 
-		[JsonProperty("type")]
 		HighlighterType? Type { get; set; }
+
+		[JsonProperty("type")]
+		[Obsolete("This is a temporary binary backwards compatible hack to make sure you can specify any custom highlightertype in 2.x, scheduled for removal in 5.0")]
+		string CustomType { get; set; }
 
 		[JsonProperty("force_source")]
 		bool? ForceSource { get; set; }
@@ -75,7 +78,12 @@ namespace Nest
 		public string TagsSchema { get; set; }
 		public bool? RequireFieldMatch { get; set; }
 		public string BoundaryChars { get; set; }
-		public HighlighterType? Type { get; set; }
+		public HighlighterType? Type
+		{
+			get { return this.CustomType.ToEnum<HighlighterType>(); }
+			set { this.CustomType = value.GetStringValue(); }
+		}
+		public string CustomType { get; set; }
 		public bool? ForceSource { get; set; }
 		public Fields MatchedFields { get; set; }
 		public QueryContainer HighlightQuery { get; set; }
@@ -110,7 +118,14 @@ namespace Nest
 
 		string IHighlightField.BoundaryChars { get; set; }
 
-		HighlighterType? IHighlightField.Type { get; set; }
+		HighlighterType? IHighlightField.Type
+		{
+#pragma warning disable CS0618 // Type or member is obsolete
+			get { return Self.CustomType.ToEnum<HighlighterType>(); }
+			set { Self.CustomType = value.GetStringValue(); }
+#pragma warning restore CS0618 // Type or member is obsolete
+		}
+		string IHighlightField.CustomType { get; set; }
 
 		bool? IHighlightField.ForceSource { get; set; }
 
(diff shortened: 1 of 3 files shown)


src/Nest/Search/Search/Highlighting/HighlightField.cs
```diff
@@ -47,9 +47,12 @@ public interface IHighlightField
 		[JsonProperty("boundary_chars")]
 		string BoundaryChars { get; set; }
 
-		[JsonProperty("type")]
 		HighlighterType? Type { get; set; }
 
+		[JsonProperty("type")]
+		[Obsolete("This is a temporary binary backwards compatible hack to make sure you can specify any custom highlightertype in 2.x, scheduled for removal in 5.0")]
+		string TypeHack { get; set; }
```
gmarz: Again, maybe name this `CustomType`?

Mpdreamz: Sounds too official in my mind, by naming it a hack explicitly we clearly signal our patch is a hack that will get removed as soon as we can.

russcam: I think the application of the `Obsolete` attribute with an appropriate message identifies it as a workaround.

gmarz: Agreed @russcam.  Additionally, I think `Type` and `CustomType` is a clearer distinction than `Type` and `TypeHack`, IMO.

Mpdreamz: A quorem has been established :)

Mpdreamz:
quorum*, i feel so bad right now.

src/Nest/Search/Search/Highlighting/HighlightField.cs
Mpdreamz: quorum*, i feel so bad right now

Mpdreamz: @gmarz @forloop ready for another review

src/Nest/Search/Search/Highlighting/HighlightField.cs
```diff
@@ -75,7 +78,12 @@ public class HighlightField : IHighlightField
 		public string TagsSchema { get; set; }
 		public bool? RequireFieldMatch { get; set; }
 		public string BoundaryChars { get; set; }
-		public HighlighterType? Type { get; set; }
+		public HighlighterType? Type
+		{
+			get { return this.CustomType.ToEnum<HighlighterType>(); }
```
gmarz: Missing null check `this.CustomType?.ToEnum()`

src/Nest/Search/Search/Highlighting/HighlightField.cs
```diff
@@ -110,7 +118,14 @@ public class HighlightFieldDescriptor<T> : DescriptorBase<HighlightFieldDescript
 
 		string IHighlightField.BoundaryChars { get; set; }
 
-		HighlighterType? IHighlightField.Type { get; set; }
+		HighlighterType? IHighlightField.Type
+		{
+#pragma warning disable CS0618 // Type or member is obsolete
+			get { return Self.CustomType.ToEnum<HighlighterType>(); }
```
gmarz: missing null check here as well

src/Nest/Search/Search/Highlighting/HighlightField.cs
Mpdreamz: ToEnum is an extension method and is safe to call on `null`

src/Nest/Search/Search/Highlighting/HighlightField.cs
Mpdreamz: ToEnum is an extension method and is safe to call on `null`

gmarz: ah, ok :+1: Threw me off because the previous PR had a null check, but I hadn't looked at the ToEnum() implementation, just assumed.

PR #2253, opened 2016-09-13
fix #2125 in master, allow to specify any highlighter not just the th…

### LINKED ISSUE #2125: Allow to use a custom Highlighter

Mpdreamz:
…ree build in ones


gmarz:
LGTM

PR #2259, opened 2016-09-13
fix #2178, can now add custom similarities using both fluent and ois …

### LINKED ISSUE #2178: Fix support for custom similarities
Currently, there's no way to specify a custom (plugged in) similarity via the fluent syntax.  See https://discuss.elastic.co/t/custom-similarities-in-elasticsearch-nest-2/56496/6

Also, looking at [SimilarityJsonConverter](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/IndexModules/Similarity/SimilarityJsonConverter.cs#L22), we have an ugly hack where we're resolving the similarity type via type name.  This would blow up when deserializing a custom similarity unless it was under the Nest namespace.

Mpdreamz:
…syntax in a matter that roundtrips (won't break get index settings)

this pr targets `master`

### DIFF
--- a/src/Nest/IndexModules/Similarity/CustomSimilarity.cs
+++ b/src/Nest/IndexModules/Similarity/CustomSimilarity.cs
@@ -1,0 +1,36 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Newtonsoft.Json;
+
+namespace Nest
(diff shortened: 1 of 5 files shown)


gmarz:
LGTM

PR #2262, opened 2016-09-15
ConfigureAwait(false) all async calls on BulkAllObservable

russcam:
Add missing `ConfigureAwait(false)` calls to async methods


Mpdreamz:
Needs to be ported to master.

PR #2263, opened 2016-09-15
Do not use IsoDateTimeConverter to handle DateTimeOffset

russcam:
`IsoDateTimeConverter` deserializes a `DateTimeOffset` represented according ISO8601 into a `DateTimeOffset` with a local offset. For example `"1999-01-01T01:01:01.001+05:00"` deserializes to a `DateTimeOffset` with a value of `01/01/1999 07:01:01 +11:00`. The +11 offset is my local offset (Australia). The default deserialization for a `DateTimeOffset` not using `IsoDateTimeConverter` deserializes into a `DateTimeOffset` with an offset that reflects the one in the ISO8601 string representation. 

This seems more appropriate behaviour.

Use `CultureInfo.InvariantCulture` in `IsoDateTimeConverter` when deserializing `DateTime`

### DIFF
--- a/src/Nest/CommonAbstractions/SerializationBehavior/ElasticContractResolver.cs
+++ b/src/Nest/CommonAbstractions/SerializationBehavior/ElasticContractResolver.cs
@@ -2,6 +2,7 @@
 using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Globalization;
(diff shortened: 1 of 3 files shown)


russcam:
Here's the related Json.NET issue: https://github.com/JamesNK/Newtonsoft.Json/issues/159

PR #2265, opened 2016-09-16
fix #2209 Several improvements in some edgecase querydsl scenarios (#…

### LINKED ISSUE #2209: StackOverflowException when combining many bool queries

Mpdreamz:
…2235)

* This commit brings about several improvements in some edgecase querydsl
scenarios

We always flattened situations like:

query1 && query2 && query3 into a single bool query with 3 must clauses.

But if the queries are booleans themselves that can't be merged we would
generate deeply nested pairs of two.

For instance we have 3 bool queries with only should clauses

bq1 && bq2 && bq3

ideally we'd generate:

bool:
  must:
     bool: should:,
     bool: should:,
     bool: should:

However due to the bitwise binary nature the way they were combined was:

Step1:

bq1 && b2 - both sides only have should clauses merging them would be
wrong so we wrap to `result1`

bool:
  must:
     bool: should:,
     bool: should:

now we do `result1 && bq2` since result1 is a must we can not merge the
should from `bq2` on it and thus we need to wrap to `result2`

bool:
  should:
    bool:
      must:
        bool: should:,
	bool: should:
    bool: should: //bq2

This is still sementically correct but now immagine you are doing this
in a loop combining thousands of these. Admittingly an edgecase it would
cause a nasty stackoverflow and query graphs that are way too deep
unnecessary as is showcased by #2209

This is now fixed in a way that does not need constant traversal of to
see if a bool must or should clause consists of only bools.

Similary the same happened if you'd || or a bunch of booleanqueries only
containing musts.

A third fix is that we no longer modify the original query when stealing
clauses. This could cause nasty errors when reusing a static bool query
e.g

```csharp
var q = Query<Project>.Bool(b => b.Should(ATermQuery).Must(ATermQuery));
var c = new QueryContainer();
for(int i = 0;i<1000;i++) c &= q;
```

The & or | bitwise operators were not pure and could potentially steal clauses from
`q` and adding them to the new result. Cuasing problems when we want to
apply q again and again and again.

* attempt to minimize allocations in the bool dsl by reusing lists and already instantiated bool queries

* documentation commit, hacked around asciidotnet not supporting tables just yet :) russcam/asciidotnet#2

* updated bool dsl performance section in the documentation

* inverted if statement for readabillity and removed some of the tests in the documentation output

Conflicts:
	docs/client-concepts/low-level/connecting.asciidoc
	src/Nest/Nest.csproj
	src/Nest/QueryDsl/Abstractions/Container/IQueryContainer.cs
	src/Nest/QueryDsl/Visitor/DslPrettyPrintVisitor.cs
	src/Tests/Framework/SerializationTests/SerializationTestBase.cs
	src/Tests/QueryDsl/QueryDslUsageTestsBase.cs
	src/Tests/Tests.csproj


docs/query-dsl/bool-dsl/operators/and-assign-many-manual-bools-usage.asciidoc
```diff
@@ -0,0 +1,160 @@
+:ref_current: https://www.elastic.co/guide/en/elasticsearch/reference/2.3
+
+:github: https://github.com/elastic/elasticsearch-net
+
+:nuget: https://www.nuget.org/packages
+
+////
+IMPORTANT NOTE
+==============
+This file has been generated from https://github.com/elastic/elasticsearch-net/tree/2.x/src/Tests/QueryDsl/BoolDsl/Operators/AndAssignManyManualBoolsUsageTests.cs.
```
russcam: Need to regen the docs to get the correct urls to the original 5.x files

russcam: LGTM - the docs need to be regen'ed to get the correct version and reference link

PR #2267, opened 2016-09-19
ClearCache FilterKeys should be a string[] array

### LINKED ISSUE #2229: Clearing a cache created in terms lookup by key
NEST version: 1.7.2
ES version: 1.7.2

In using terms lookup, I'm having it create a named cache for the terms.  Periodically, when I detect that the source of those terms has changed, I want to invalidate that cache.  I've discovered the endpoint and syntax for a raw request, for example:
POST posts,comments/_cache/clear?filter_keys=members_123
if the members list in the 123 document changes.

I cannot see how I can generate that request using NEST.   What I see in
`public IIndicesResponse ClearCache(List<string> indices, ClearCacheOptions options)`
is close, but it doesn't support `filter_keys` nor a way to pass those keys in.  Is there a way to accomplish this that I'm missing?

Thanks,
Tom Doman

russcam:
- Update the rest api spec to reflect the PR in https://github.com/elastic/elasticsearch/pull/20292/
- Rename FilterKeys string array property to FilterCacheKeys to maintain backwards binary compatibility.
- Mark all usages of the bool FilterKeys value as Obsolete, including the error flag.
- Bump ApiGenerator to point at 1.7 branch.
- Make FluentRequestParameters a partial class so that generated types can be extended.
- Add PatchMethod to IDescriptorOverrides to allow patching on a per method basis.
- Move Csharp property naming on Request types to ApiQueryParameters type so that property naming can be modified independently of the underlying query string parameter name.

Fixes #2229


russcam:
Added url tests for both fluent and object initializer syntax 👍

russcam:
@gmarz, would you mind reviewing please?

gmarz:
LGTM

Issue #2112, opened 2016-05-27
Backport reindex rethrottle to 2.x


russcam:
See https://github.com/elastic/elasticsearch-net/pull/2083

Mpdreamz:
Now Elasticsearch 2.4 is out we can backport it

russcam:
merged into 2.x in #2278

PR #2278, opened 2016-09-22
Support ReindexRethrottle

russcam:
- Update rest api specs to 2.4.0
- Patch documentation for `RequestsPerSecond` to indicate that `float.PositiveInfinity` should be used to set no throttle. Per https://github.com/elastic/elasticsearch/issues/20625, the spec indicates that it supports `0` for no throttle but Elasticsearch 2.4.0 does not allow this, so special case `float.PositiveInfinity` to `"unlimited"`. Similarly, on the response, convert `"unlimited"` to `float.PositiveInfinity` with a custom converter. 

A custom converter is used over `Union<string, float>` as it aligns the response with the request and also aligns with NEST 5.x where `"unlimited"` is removed.


russcam:
Closes https://github.com/elastic/elasticsearch-net/issues/2112

src/CodeGeneration/ApiGenerator/RestSpecification/Core/bulk.json
```diff
@@ -1,6 +1,6 @@
 {
   "bulk": {
-    "documentation": "http://www.elastic.co/guide/en/elasticsearch/reference/2.3/docs-bulk.html",
+    "documentation": "http://www.elastic.co/guide/en/elasticsearch/reference/master/docs-bulk.html",
```
Mpdreamz: why did this change from 2.3 to master?

russcam: I updated the rest api specs to point at `v.2.4.0` to pull in the spec for reindex rethrottle. [The urls have not been updated on that branch](https://github.com/elastic/elasticsearch/blob/v2.4.0/rest-api-spec/src/main/resources/rest-api-spec/api/bulk.json#L3).

PR #2279, opened 2016-09-22
Support UpdateByQuery fluent method taking only selector delegate

russcam:
This aligns with 5.x. Provide an overload to `UpdateByQuery()` that accepts only the selector, setting `Indices` and `Types` to `T`. Keep existing methods for BWC


src/Elasticsearch.Net/Domain/Enums.Generated.cs
```diff
@@ -1,117 +1,117 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Runtime.Serialization;
+
+///This file contains all the typed enums that the client rest api spec exposes.
+///This file is automatically generated from https://github.com/elasticsearch/elasticsearch-rest-api-spec
+///Generated of commit v2.3.5
+namespace Elasticsearch.Net
```
Mpdreamz: I wonder why we keep on changing linefeeds in these generated files

my `git config -l ` lists:

`core.autocrlf=input`

http://stackoverflow.com/questions/3206843/how-line-ending-conversions-work-with-git-core-autocrlf-between-different-operat

We should pick one and standardize on it.

russcam: My bad, looks like it's me 😞 

`core.autocrlf=true`. I'll change now to `input`

gmarz: Mine is also `core.autocrlf=true` which I thought was the recommended setting for Windows?

PR #2295, opened 2016-09-29
Updated xml docs to specify the correct method name

niemyjski:


gmarz:
Back ported to 2.x and 5.x

Issue #2308, opened 2016-10-07
Error deserializing "took" value on TermVector request.


maeserichar:
**NEST/Elasticsearch.Net version**: 2.4.6

**Elasticsearch version**: 2.3.3

**Description of the problem including expected versus actual behavior**: 
When asking for the TermVector, if the took field in the response is bigger than max value for an int, it gives a serialization exception.

I have noticed that depending of the response type, some has the Took field as a long and some others (like the received with the term vector request) as an integer. I would be glad to submit a PR unifying this types to long, but first I wanted to assure that this is not on purpose.

gmarz:
Closed via https://github.com/elastic/elasticsearch-net/pull/2324 and we have https://github.com/elastic/elasticsearch-net/pull/2325 as a potential fix for the next 2.x release.

PR #2317, opened 2016-10-11
Change type of took field in responses.

maeserichar:
This PR copes with the changes needed for #2308.

I have hunted down the rest of int "Took" fields, and modified some auxiliar classes that uses said field.

### DIFF
--- a/src/Benchmarking/PartitionExtension.cs
+++ b/src/Benchmarking/PartitionExtension.cs
@@ -35,13 +35,13 @@ namespace Benchmarking
 			}
 		}
 
-		public static decimal GetMedian(this IEnumerable<int> source)
+		public static decimal GetMedian(this IEnumerable<long> source)
 		{
 			// Create a copy of the input, and sort the copy
-			int[] temp = source.ToArray();
+			long[] temp = source.ToArray();
 			Array.Sort(temp);
 
-			int count = temp.Length;
+			long count = temp.Length;
 			if (count == 0)
 			{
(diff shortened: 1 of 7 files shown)


gmarz:
@maeserichar huge thanks for this PR.  Unfortunately this breaks backwards compatibility on the 2.x branch though.  Would you mind re-targeting this PR for master?  We'll have to come up with a modified approach for 2.x

maeserichar:
@gmarz What would be the process to retarget the PR (I'm kind of a newbie in git/github flow)?

For the modified approach for the 2.x branch, what can I do about that?

Regards,

gmarz:
> What would be the process to retarget the PR (I'm kind of a newbie in git/github flow)?

Since this is a relatively small change, the simplest way is probably to just pull down master, make the changes again, and open a new PR against master.

> For the modified approach for the 2.x branch, what can I do about that?

This is unfortunate, but we can introduce a new property that is of type long and move the the JsonProperty attribute over to it.  Then mark the original `Took` property as obsolete (with a verbose message) and change its getter to cast and return the new long property.  If the value exceeds int.MaxValue, then we can let it wrap around...but the obsolete warning message should make it clear that this is a possibility and why the new property should be used instead.  Thoughts?  Open to suggestions here.  /cc @elastic/microsoft

maeserichar:
@gmarz Ok, I have sent #2324 with the changes to the Master branch. I will try to submit a first version of the discussed changes over the 2.x branch.

Thanks!

PR #2321, opened 2016-10-18
Fix #2319 caching of constant Field expressions

### LINKED ISSUE #2319: 2.x Improperly caches field name when dynamic field + suffix combination used
NEST 2.4.4
Elasticsearch.Net.2.4.4

**Elasticsearch version**:
2.4.0

When functionally generating queries that includes a dynamic field's child multi-field, the first query condition is generated properly, but subsequent conditions will continue to use the original field name instead of the correct field name.  Only occurs when there both a constant and a non-constant in the Field() function's expression.

**Steps to reproduce**:
``` C#
using System;
using Nest;
using System.Text;

namespace Test
{

    public class Program
    {
        public static void Main()
        {
            var connectionSettings = new ConnectionSettings(new Uri("http://localhost:9200"));
            connectionSettings.DisableDirectStreaming();
            connectionSettings.PrettyJson();

            var elasticClient = new ElasticClient(connectionSettings);

            var expected = elasticClient.Search<TestDocument>(sd => sd
                .Query(qcd => qcd
                    .Bool(bqd => bqd
                        .Must(
                            qcd
                                .Term(tqd => tqd
                                    .Field(td => td.fields["foo"].Suffix("suffix"))
                                    .Value("a")),
                            qcd
                                .Term(tqd => tqd
                                    .Field(td => td.fields["bar"].Suffix("suffix"))
                                    .Value("b"))
                        )
                    )
                )
            );

            var actual = elasticClient.Search<TestDocument>(sd => sd
                .Query(qcd => qcd
                    .Bool(bqd => bqd
                        .Must(
                            CreateCondition("foo", "a"),
                            CreateCondition("bar", "b")
                        )
                    )
                )
            );

            Console.WriteLine("Expected: {0}", Encoding.UTF8.GetString(expected.ApiCall.RequestBodyInBytes));
            Console.WriteLine();
            Console.WriteLine("Actual:   {0}", Encoding.UTF8.GetString(actual.ApiCall.RequestBodyInBytes));
        }
        
        private static Func<QueryContainerDescriptor<TestDocument>, QueryContainer> CreateCondition(string fieldName, string value)
        {
            return qcd => qcd
                .Term(tqd => tqd
                    .Field(td => td.fields[fieldName].Suffix("suffix"))
                    .Value(value));
        }

        private class TestDocument
        {
            public IDictionary<string, string> fields { get; set; }
        }
    }
}
```

Outputs:
``` JSON
Expected: {
  "query": {
    "bool": {
      "must": [
        {
          "term": {
            "fields.foo.suffix": {
              "value": "a"
            }
          }
        },
        {
          "term": {
            "fields.bar.suffix": {
              "value": "b"
            }
          }
        }
      ]
    }
  }
}

Actual:   {
  "query": {
    "bool": {
      "must": [
        {
          "term": {
            "fields.foo.suffix": {
              "value": "a"
            }
          }
        },
        {
          "term": {
            "fields.foo.suffix": {
              "value": "b"
            }
          }
        }
      ]
    }
  }
}
```

Root cause is src/Nest/CommonAbstractions/Infer/Field/Field.cs#L48 setting the expression to cacheable when there is a constant component to the expression.  

Instead, CachableExpression should only be set when there are only constant expressions.  You could check that there are no MemberAccess nodes (variable references are captured as members), but that wouldn't handle cases such as ```tqd.Field(td => td.fields[SomeFunction()])```.

N/A

N/A

gmarz:
We only cache Field expressions that do not contain variables.  For
instance:

```csharp
  Field<Project>(p => p.Name)
  Field<Project>(p => p.Name.Suffix("foo"))
  Field<Project>(p => p.Metadata["foo"])
```
are all cachable since they only contain constant expressions.

whereas

```csharp
  Field<Project>(p => p.Name.Suffix(myVariable))
  Field<Project>(p => p.Metadata[myVariable])
```
are not cachable since we can't evaluate the value of `myVariable` to be
used as a cache key.

However, there was a bug which this commit fixes where we would cache a
variable expression that ended with a constant.  For instance:

```csharp
  Field<Project>(p => p.Metadata[myVariable].Suffix("foo"))
```
This expression was incorrectly cached because HasConstantExpressionVisitor
only concerned itself with the last method call: `Suffix("foo")`.

With this change, `HasConstantExpressionVisitor` has been renamed to
`HasVariableExpressionVisitor` to better align with what it actually
does.  And now accounts for every node in the expression, ensuring that
we only cache expressions that contain only constants.

### DIFF
--- a/src/Nest/CommonAbstractions/Infer/Field/FieldExpressionVisitor.cs
+++ b/src/Nest/CommonAbstractions/Infer/Field/FieldExpressionVisitor.cs
@@ -12,18 +12,24 @@ using System.Threading.Tasks;
 
 namespace Nest
 {
-	internal class HasConstantExpressionVisitor : ExpressionVisitor
+	internal class HasVariableExpressionVisitor : ExpressionVisitor
 	{
-		public bool Found { get; private set; }
+		private bool _found;
+		public bool Found
+		{
+			get { return _found; }
+			// This is only set to true once to prevent clobbering from subsequent node visits
+			private set { if (!_found) _found = value; }
+		}
 
-		public HasConstantExpressionVisitor(Expression e)
+		public HasVariableExpressionVisitor(Expression e)
 		{
 			this.Visit(e);
 		}
 
 		public override Expression Visit(Expression node)
 		{
-			if (!Found)
+			if (!this.Found)
 				return base.Visit(node);
 			return node;
 		}
@@ -35,7 +41,6 @@ namespace Nest
 				var lastArg = node.Arguments.Last();
 				var constantExpression = lastArg as ConstantExpression;
 				this.Found = constantExpression == null;
-				return node;
 			}
 			else if (node.Method.Name == "get_Item" && node.Arguments.Any())
 			{
@@ -50,15 +55,8 @@ namespace Nest
 				var lastArg = node.Arguments.Last();
 				var constantExpression = lastArg as ConstantExpression;
 				this.Found = constantExpression == null;
-				return node;
 			}
 			return base.VisitMethodCall(node);
-		}
-
-		protected override Expression VisitConstant(ConstantExpression node)
-		{
-			this.Found = true;
-			return node;
 		}
 	}
 
--- a/src/Nest/CommonAbstractions/Infer/Field/Field.cs
+++ b/src/Nest/CommonAbstractions/Infer/Field/Field.cs
@@ -38,7 +38,7 @@ namespace Nest
 				var comparisonValue = value.ComparisonValueFromExpression(out type);
 				_type = type;
 				SetComparisonValue(comparisonValue);
-				CacheableExpression = !new HasConstantExpressionVisitor(value).Found;
+				CacheableExpression = !new HasVariableExpressionVisitor(value).Found;
 			}
 		}
 
--- a/src/Nest/CommonAbstractions/Infer/PropertyName/PropertyName.cs
+++ b/src/Nest/CommonAbstractions/Infer/PropertyName/PropertyName.cs
@@ -35,7 +35,7 @@ namespace Nest
 			return new PropertyName
 			{
 				Expression = expression,
-				CacheableExpression = !new HasConstantExpressionVisitor(expression).Found,
+				CacheableExpression = !new HasVariableExpressionVisitor(expression).Found,
 				_comparisonValue = expression.ComparisonValueFromExpression(out type),
 				_type = type
 			};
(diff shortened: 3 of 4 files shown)


src/Nest/CommonAbstractions/Infer/Field/FieldExpressionVisitor.cs
```diff
@@ -14,7 +14,9 @@ namespace Nest
 {
 	internal class HasConstantExpressionVisitor : ExpressionVisitor
 	{
-		public bool Found { get; private set; }
+		public bool Found => this.FoundConstants.Any(b => b);
```
Mpdreamz: `.Count > 0` since its an `ICollection`

gmarz: We're checking for any `true` values within the collection, so `Count > 0` isn't enough.

Mpdreamz: You're right, i wonder if we can just make it a bool flag that once set is never unset? Would safe on allocs too.

gmarz: Yea, good call. No need to maintain a collection at all, as long as we still visit each method call then we can use just the `Found` flag.  I've changed this with my last commit.

Mpdreamz:
Lgtm :+1:

russcam:
LGTM 👍

brandondahler:
Shouldn't the expression visitor be renamed HasNonConstantExpressionVisitor?

src/Nest/CommonAbstractions/Infer/Field/FieldExpressionVisitor.cs
```diff
@@ -35,7 +41,7 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
 				var lastArg = node.Arguments.Last();
 				var constantExpression = lastArg as ConstantExpression;
 				this.Found = constantExpression == null;
-				return node;
```
brandondahler: This can be left in to prevent traversing inner nodes unnecessarily.

src/Nest/CommonAbstractions/Infer/Field/FieldExpressionVisitor.cs
```diff
@@ -50,16 +56,9 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
```
brandondahler: This can be left in to prevent traversing inner nodes unnecessarily.

src/Nest/CommonAbstractions/Infer/Field/FieldExpressionVisitor.cs
gmarz: I don't think so.  We need to actually visit all of the inner nodes to ensure there are only constants in the expression.

gmarz:
> Shouldn't the expression visitor be renamed HasNonConstantExpressionVisitor?

@brandondahler Agreed, it was confusing,  I've renamed it to `HasVariableExpressionVisitor`.

gmarz:
Ported to 5.x and master.

PR #2324, opened 2016-10-19
Change type of took field in responses

maeserichar:
As discussed on #2317, this is the first PR that modifies the type of the "Took" field in the remaining responses that are ints.


gmarz:
LGTM 👍 Thank you for this @maeserichar !

gmarz:
Ported to 5.x.

PR #2325, opened 2016-10-19
Adding a new field to deserialize took as long, and marking the old o…

maeserichar:
…ne as obsolete.

This is the second part of the changes discussed on #2317 (or at least a first try). 

Two things:
- I´m not very happy with the 'TookAsLong' name, but cannot find a better one, so suggestion are much appreciated.
- As the Took field is now marked as obsolete in those responses, there are some warnings in the tests. I have to use the new field, right?

/cc @gmarz

### DIFF
--- a/src/Nest/Document/Multiple/Bulk/BulkResponse.cs
+++ b/src/Nest/Document/Multiple/Bulk/BulkResponse.cs
@@ -2,12 +2,16 @@
 using System.Linq;
 using System.Text;
 using Newtonsoft.Json;
+using System;
 
 namespace Nest
 {
 	public interface IBulkResponse : IResponse
 	{
+		[Obsolete(@"Took field is an Int but the value in the response can exced the max value for Int.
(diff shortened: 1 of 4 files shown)


gmarz:
Thanks @maeserichar!  Discussed this with the team and we're deciding whether or not we should just fix this properly in 2.x as well (e.g. your changes in #2324) albeit being a breaking change.  Going to let this bake for a bit.

russcam:
We're going to pull this into 2.x; after several discussions, the `TookAsLong` seems to be the pragmatic way to maintain backwards binary compatibility here. I'm going to change the implementation slightly though to not overflow and instead return `int.MaxValue` if the value is greater than the value an int 32 can hold. This seems _truthier_ than overflowing

russcam:
pulled into 2.x in 2546ab73dd2539865efc49f96d60acf8531a075b

Thanks @maeserichar 👍

Issue #2322, opened 2016-10-19
Add support for half_float and scaled_float numeric datatypes


russcam:
https://www.elastic.co/guide/en/elasticsearch/reference/5.x/number.html

gmarz:
Closed via #2337

PR #2337, opened 2016-10-25
Refactor property and attribute mapping tests

gmarz:
Our property mapping tests weren't sufficient as they
only tested serilization of fluent syntax and attribute
types.

This commit adds new tests that test both the fluent and
initializer types, and run them as integration tests as well
so we can validate our properties against ES.

Also tackled #2322 and #2059 as part of this, as I did a full pass over
all of our datatype mappings.


src/Tests/Mapping/Types/TypeMappingTestBase.cs
```diff
@@ -1,41 +1,42 @@
...
 {
-	public abstract class TypeMappingTestBase<T>
-			where T : class
+	public abstract class TypeMappingTestBase
+		: ApiIntegrationAgainstNewIndexTestBase<WritableCluster, IPutMappingResponse, IPutMappingRequest, PutMappingDescriptor<Project>, PutMappingRequest<Project>>
 	{
-		protected abstract object ExpectJson { get; }
 
-		protected virtual object ExpectJsonFluentOnly { get; }
+		public TypeMappingTestBase(WritableCluster cluster, EndpointUsage usage) : base(cluster, usage) { }
```
russcam: nitpick: `protected`

russcam:
LGTM

gmarz:
Ported to master 00048122659b6afed967f284a71050b8c0a685ff

PR #2338, opened 2016-10-25
Fix AppendSuffix to work with different expressions and updated test

### LINKED ISSUE #2331: AppendSuffix does not work as expected
**NEST/Elasticsearch.Net version**: 2.4.6

When chaining multiple `AppendSuffix(string suffix)` calls, the suffixes are appended in reverse order.
For example `expression.AppendSuffix("raw").AppendSuffix("evendeeper")` will append "evendeeper.raw" instead of "raw.evendeeper".

There is also a closely related problem; if trying to append a suffix with `AppendSuffix(string suffix)` to expressions like `p => p.Metadata["hardcoded"]` an exception is thrown:

> System.ArgumentException was caught
>   _message=Die auf Typ "System.Collections.Generic.IDictionary'2[System.String,System.String]" deklarierte Methode "System.String get_Item(System.String)" kann nicht mit einer Instanz vom Typ "System.Object" aufgerufen werden.
> 
>    bei System.Linq.Expressions.Expression.ValidateCallInstanceType(Type instanceType, MethodInfo method)
>    bei System.Linq.Expressions.Expression.ValidateStaticOrInstanceMethod(Expression instance, MethodInfo method)
>    bei System.Linq.Expressions.Expression.Call(Expression instance, MethodInfo method, IEnumerable'1 arguments)
>    bei System.Linq.Expressions.InstanceMethodCallExpressionN.Rewrite(Expression instance, IList'1 args)
>    bei System.Linq.Expressions.ExpressionVisitor.VisitMethodCall(MethodCallExpression node)
>    bei System.Linq.Expressions.MethodCallExpression.Accept(ExpressionVisitor visitor)
>    bei System.Linq.Expressions.ExpressionVisitor.Visit(Expression node)
>    bei Nest.ExpressionExtensions.AppendSuffix[T](Expression'1 expression, String suffix)

**Possible solution**:
After some research I might have found the problem:
In [ExpressionExtensions](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/CommonAbstractions/Extensions/ExpressionExtensions.cs), the `SuffixExpressionVisitor` should override `Visit` instead of `VisitMember`. So the corrected class should look like this:

```
    /// <summary>
    /// Calls <see cref="SuffixExtensions.Suffix"/> on a member expression.
    /// </summary>
    private class SuffixExpressionVisitor : ExpressionVisitor
    {
        private readonly string suffix;

        public SuffixExpressionVisitor(string suffix)
        {
            this.suffix = suffix;
        }

        public override Expression Visit(Expression node)
        {
            return Expression.Call(
                typeof(SuffixExtensions),
                nameof(SuffixExtensions.Suffix),
                null,
                node,
                Expression.Constant(suffix));
        }

        protected override Expression VisitUnary(UnaryExpression node)
        {
            return node;
        }
    }
```

You might find these two tests useful, too:

```
        public void AppendingSuffixToExpressions() 
        {
            /** Here we have a list of expressions */
            var expressions = new List<Expression<Func<Project, object>>>
            {
                p => p.Name,
                p => p.Description,
                p => p.CuratedTags.First().Name,
                p => p.LeadDeveloper.FirstName,
                p => p.Metadata["hardcoded"]
            };

            /** and we want to append the suffix "raw" to each */
            var fieldExpressions =
                expressions.Select<Expression<Func<Project, object>>, Field>(e => e.AppendSuffix("raw")).ToList();


            Expect("name.raw").WhenSerializing(fieldExpressions[0]);
            Expect("description.raw").WhenSerializing(fieldExpressions[1]);
            Expect("curatedTags.name.raw").WhenSerializing(fieldExpressions[2]);
            Expect("leadDeveloper.firstName.raw").WhenSerializing(fieldExpressions[3]);
            Expect("metadata.hardcoded.raw").WhenSerializing(fieldExpressions[4]);
        }

        public void AppendingSuffixInDepthToExpressions() 
        {
            /** Here we have a list of expressions */
            var expressions = new List<Expression<Func<Project, object>>>
            {
                p => p.Name,
                p => p.Description,
                p => p.CuratedTags.First().Name,
                p => p.LeadDeveloper.FirstName
            };

            /** and we want to append the suffixes "raw" and "evendeeper" to each */
            var fieldExpressions =
                expressions.Select<Expression<Func<Project, object>>, Field>(e => e.AppendSuffix("raw").AppendSuffix("evendeeper")).ToList();


            Expect("name.raw.evendeeper").WhenSerializing(fieldExpressions[0]);
            Expect("description.raw.evendeeper").WhenSerializing(fieldExpressions[1]);
            Expect("curatedTags.name.raw.evendeeper").WhenSerializing(fieldExpressions[2]);
            Expect("leadDeveloper.firstName.raw.evendeeper").WhenSerializing(fieldExpressions[3]);
        }
```

Hope this helps!

SimonKoegler:
AppendSuffix should work with any expression instead of just member access expressions
Closes #2331

### DIFF
--- a/src/Nest/CommonAbstractions/Extensions/ExpressionExtensions.cs
+++ b/src/Nest/CommonAbstractions/Extensions/ExpressionExtensions.cs
@@ -31,7 +31,7 @@ namespace Nest
 				this.suffix = suffix;
 			}
 
-			protected override Expression VisitMember(MemberExpression node)
+			public override Expression Visit(Expression node)
 			{
 				return Expression.Call(
 					typeof(SuffixExtensions),
--- a/src/Tests/ClientConcepts/HighLevel/Inference/FieldInference.doc.cs
+++ b/src/Tests/ClientConcepts/HighLevel/Inference/FieldInference.doc.cs
@@ -284,7 +284,8 @@ namespace Tests.ClientConcepts.HighLevel.Inference
 				p => p.Name,
 				p => p.Description,
 				p => p.CuratedTags.First().Name,
-				p => p.LeadDeveloper.FirstName
+				p => p.LeadDeveloper.FirstName,
+				p => p.Metadata["hardcoded"]
 			};
 
 			/** and we want to append the suffix "raw" to each */
@@ -295,6 +296,17 @@ namespace Tests.ClientConcepts.HighLevel.Inference
 			Expect("description.raw").WhenSerializing(fieldExpressions[1]);
 			Expect("curatedTags.name.raw").WhenSerializing(fieldExpressions[2]);
 			Expect("leadDeveloper.firstName.raw").WhenSerializing(fieldExpressions[3]);
+			Expect("metadata.hardcoded.raw").WhenSerializing(fieldExpressions[4]);
+
+			/** or we might even want to chain multiple `.AppendSuffix()` calls */
+			var multiSuffixFieldExpressions =
+				expressions.Select<Expression<Func<Project, object>>, Field>(e => e.AppendSuffix("raw").AppendSuffix("evendeeper")).ToList();
+
+			Expect("name.raw.evendeeper").WhenSerializing(multiSuffixFieldExpressions[0]);
+			Expect("description.raw.evendeeper").WhenSerializing(multiSuffixFieldExpressions[1]);
+			Expect("curatedTags.name.raw.evendeeper").WhenSerializing(multiSuffixFieldExpressions[2]);
+			Expect("leadDeveloper.firstName.raw.evendeeper").WhenSerializing(multiSuffixFieldExpressions[3]);
+			Expect("metadata.hardcoded.raw.evendeeper").WhenSerializing(multiSuffixFieldExpressions[4]);
 		}
 
 		/**=== Attribute based naming


src/Nest/CommonAbstractions/Extensions/ExpressionExtensions.cs
```diff
@@ -31,7 +31,7 @@ public SuffixExpressionVisitor(string suffix)
 				this.suffix = suffix;
 			}
 
-			protected override Expression VisitMember(MemberExpression node)
+			public override Expression Visit(Expression node)
```
gmarz: Should be `protected`?

SimonKoegler: @gmarz, well I don't think it makes any difference. The public `Visit(Expression node)` method is called anyway and by default "dispatches the expression to one of the more specialized visit methods in this class". Thus the protected `VisitMember(MemberExpression node)` is called by the public `Visit(Expression node)` method. But since we want to append the suffix to the whole expression, not just to the specialized parts of it, I can't see any other way than directly override the public method. Or did I miss something?

gmarz: Nope, I overlooked it :).  Lgtm 👍

Mpdreamz:
Thank you @SimonKoegler, merged to 2x, 5x, and master 👍

PR #2349, opened 2016-10-27
Include ignore_above in string keyword mapping inference

### LINKED ISSUE #2329: Infer the mapping for strings as text types only in 5.x
in 5.x+, [we're inferring the mapping of a string property as a `text` type with a `keyword` sub-fields mapping](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Mapping/Visitor/PropertyWalker.cs#L89-L96). This feels
-  very overbearing as the default 
- may not play too nicely with large input strings to map them as `keyword`
- goes beyond what we do in 1.x and 2.x

Is there a reason that this is done? I feel like the default should be to map as a `text` type only, to align with similar behaviour in 1.x and 2.x.

russcam:
This aligns with the default inferred mapping for a string in Elasticsearch 5.0
Closes #2329

### DIFF
--- a/src/Nest/Mapping/Types/Core/String/StringAttribute.cs
+++ b/src/Nest/Mapping/Types/Core/String/StringAttribute.cs
@@ -32,6 +32,7 @@ namespace Nest
 		public TermVectorOption TermVector { get { return Self.TermVector.GetValueOrDefault(); } set { Self.TermVector = value; } }
 		public bool Norms { get { return Self.Norms.GetValueOrDefault(true); } set { Self.Norms = value; } }
 
+		[Obsolete("Only valid for indices created before Elasticsearch 5.0 and will be removed in the next major version.  For newly created indices, use Text or Keyword attribute instead.")]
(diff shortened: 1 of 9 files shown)


src/Nest/Mapping/Types/Core/String/StringAttribute.cs
```diff
@@ -32,6 +32,7 @@ public class StringAttribute : ElasticsearchDocValuesPropertyAttributeBase, IStr
 		public TermVectorOption TermVector { get { return Self.TermVector.GetValueOrDefault(); } set { Self.TermVector = value; } }
 		public bool Norms { get { return Self.Norms.GetValueOrDefault(true); } set { Self.Norms = value; } }
 
+		[Obsolete("Only valid for indices created before Elasticsearch 5.0 and will be removed in the next major version.  For newly created indices, use Text or Keyword attribute instead.")]
```
Mpdreamz: 👍

Mpdreamz:
picked to master e96f3d8 as well

PR #2351, opened 2016-10-28
Add SniffResponseTesting to expose SniffResponse.AddressRe for unit t…

### LINKED ISSUE #2350: SniffResponse AddressRe fails to match on IPv6 address
**NEST/Elasticsearch.Net version**: 5.0.0-rc1

**Elasticsearch version**: 5.0.0

The RegEx AddressRe does not match [::1]:9200 (because it's 'confused' by the additional colons)
I think it should, according to the examples given here: http://www.ietf.org/rfc/rfc2732.txt

**Steps to reproduce**:
This might be too specific to my environment (or I've messed it up somehow!) but I pass a single value of "http://awelburn.artesian.lan:9200" (my company's internal 'FQDN') into a SniffingConnectionPool, this is then passed into a new ElasticClient.

This produces two elements for NodeInfoHttp bound_address: [::1]:9200 and 10.1.2.59:9200

SniffResponse.ToNodes then calls ParseToUri(kv.Value.http?.bound_address.FirstOrDefault(), forceHttp)

it then tries to match [::1]:9200 and the Regex fails.

Whilst there is probably a way to force the IPv4 address as the first element, I think we should probably fix this RegEx to handle IPv6 as well.

I'm about to submit a PR to adjust the Rexex pattern.

awelburn:
Added SniffResponseTesting just to expose SniffResponse.AddressRe so we don't have to adjust the visibility of SniffResponse.

Also looked at adding InternalsVisibleTo for the Test project to Elasticsearch.Net, but the strong name / signing prevented this.
fixes #2350 

CLA signing is in progress

### DIFF
--- a/src/Elasticsearch.Net/Transport/Sniff/SniffResponse.cs
+++ b/src/Elasticsearch.Net/Transport/Sniff/SniffResponse.cs
@@ -5,12 +5,12 @@ using System.Text.RegularExpressions;
 
 namespace Elasticsearch.Net
 {
-	internal class SniffResponse
+	public class SniffResponse
 	{
-		internal static Regex AddressRe { get; } = new Regex(@"^((?<fqdn>[^/]+)/)?(?<ip>[^:]+|\[[\da-fA-F:\.]+\]):(?<port>\d+)$");
+		public static Regex AddressRegex { get; } = new Regex(@"^((?<fqdn>[^/]+)/)?(?<ip>[^:]+|\[[\da-fA-F:\.]+\]):(?<port>\d+)$");
 
 		public string cluster_name { get; set; }
-		public Dictionary<string, NodeInfo> nodes { get; set; }
+		internal Dictionary<string, NodeInfo> nodes { get; set; }
 
 		public IEnumerable<Node> ToNodes(bool forceHttp = false)
 		{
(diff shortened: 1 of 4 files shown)


src/Elasticsearch.Net/Transport/Sniff/SniffResponse.cs
```diff
@@ -7,39 +7,38 @@ namespace Elasticsearch.Net
 {
 	internal class SniffResponse
```
Mpdreamz: maybe make this class public with an internal constructor so that only `Elasticsearch.Net` can instantiate it?

That way `AddressRe` can be public and renamed to `AddressRegex`

Mpdreamz:
@awelburn thanks for submitting this, looks good other than my small suggestion.

Would you mind signing our CLA so that we are allowed to pull your bits in?

https://www.elastic.co/contributor-agreement

Mpdreamz:
@awelburn thanks for the updated commits! I'm eager to pull this in but I can't find you on our CLA list.

Did you sign our https://www.elastic.co/contributor-agreement ?

awelburn:
@Mpdreamz It's with our CTO for signing, will update when there's progress.

awelburn:
@Mpdreamz we signed the contract a few minutes ago

Mpdreamz:
Awesome, I can see it in our database. We'll pull this in before 5.0 GA hits the streets (this week or early next week).

gmarz:
Merged to 5.x and master.

PR #2366, opened 2016-11-03
Allow 404s for Get and Delete requests

### LINKED ISSUE #2306: 5.0.0-beta:  Delete Response (404) result is incorrect
I have a test where I'm deleting the a document that doesn't exist and checking the behaviors and I noticed that the result in the response is `not_found` but the returned model Result value is `Created`
screen shot 2016-10-03 at 2 49 55 pm

russcam:
Update Rest Api Specs to v5.0.0

Closes #2306

### DIFF
--- a/src/Tests/Reproduce/GithubIssue2306.cs
+++ b/src/Tests/Reproduce/GithubIssue2306.cs
@@ -1,0 +1,33 @@
+﻿using FluentAssertions;
+using Nest;
+using Tests.Framework;
+using Tests.Framework.Integration;
+using Tests.Framework.MockData;
+using Xunit;
+
+namespace Tests.Reproduce
+{
+	public class GithubIssue2306 : IClusterFixture<ReadOnlyCluster>
+	{
+		private readonly ReadOnlyCluster _cluster;
+
+		public GithubIssue2306(ReadOnlyCluster cluster)
+		{
+			_cluster = cluster;
+		}
+
+		[I]
+		public void DeleteNonExistentDocumentReturnsNotFound()
+		{
+			var client = _cluster.Client;
+			var response = client.Delete<Project>("non-existent-id");
+
+			response.ShouldBeValid();
+			response.Found.Should().BeFalse();
+			response.Result.Should().Be(Result.NotFound);
+			response.Index.Should().Be("project");
+			response.Type.Should().Be("project");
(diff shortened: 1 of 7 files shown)


src/Elasticsearch.Net/Domain/RequestParameters/RequestParameters.Generated.cs
```diff
@@ -1565,6 +1489,18 @@ public class DeleteByQueryRequestParameters : FluentRequestParameters<DeleteByQu
 		public DeleteByQueryRequestParameters Df(string df) => this.AddQueryString("df", df);
 		
 		
+		///<summary>Specify whether to return detailed information about score computation as part of a hit</summary>
+		public DeleteByQueryRequestParameters Explain(bool explain) => this.AddQueryString("explain", explain);
+		
+		
+		///<summary>A comma-separated list of stored fields to return as part of a hit</summary>
+		public DeleteByQueryRequestParameters StoredFields(params string[] stored_fields) => this.AddQueryString("stored_fields", stored_fields);
```
Mpdreamz: We should probably make sure everything with `*Fields` gets generated as fields that take Fields and Expressions. Like other fields do too. New issue though.

src/Tests/Reproduce/GithubIssue2306.cs
```diff
@@ -0,0 +1,30 @@
...
+
+		[I]
+		public void DeleteNonExistentDocumentReturnsNotFound()
+		{
+			var client = _cluster.Client;
+			var response = client.Delete<Project>("non-existent-id");
+			
+			response.ShouldBeValid();
+			response.Found.Should().BeFalse();
+			response.Result.Should().Be(Result.NotFound);
```
Mpdreamz: Assert .Index and the like as well

src/Elasticsearch.Net/Domain/RequestParameters/RequestParameters.Generated.cs
russcam: This is in Elasticsearch.Net though; `Fields` exists in NEST?

src/Tests/Reproduce/GithubIssue2306.cs
russcam: Done!

PR #2367, opened 2016-11-03
Allow 404s for Get and Delete requests

### LINKED ISSUE #2306: 5.0.0-beta:  Delete Response (404) result is incorrect

russcam:
Update Rest Api Specs to v2.4.1
Closes #2306


russcam:
Added integration tests for delete and get on non-existent documents.

rdehuyss:
Guys, we wanted to let you know that this breaks backwards compatability. We just lost a lot of time due to this enhancement... Can these kind of changes be kept for major upgrades?

Mpdreamz:
I'm sorry to hear @rdehuys we strive not to break bwc but I can see this being a big functional change. 

should we

1. Revert this and delist 2.4.7? Or would that be even worse. 
2. Keep this but with an obsolete warning on IsValid going forward. 

Neither is pretty but I'd hate for others to lose time here too. 

Cc @elastic/microsoft

rdehuyss:
We upgraded our part of the code, so all is fine now. Our test suite luckily failed so nothing went broken to production.

So, option 2 is fine for us!

PR #2372, opened 2016-11-04
fix #2368 by fixing the underlying problem that scalars should be easier

### LINKED ISSUE #2368: [5.0 Breaking] Incorrect numeric NumberType mappings are created.
In 1.x client the mappings were (mostly, with a bigger datatype) correctly created, now with the 5.x mappings everything I've come across seems to be mapped to a float.. It seems we are not setting the correct `NumberType`.

```cs
.Number(f => f.Name(e => e.Value).Alias(Alias.Value)) // value is of type decimal?
.Number(f => f.Name(e => e.Count).Alias(Alias.Count)) // count is of type int?
```

previous versions of the mapping would have created these as (mostly) correct types:
```json
{
               "value": {
                  "type": "double",
                  "include_in_all": false
               },
               "count": {
                  "type": "double",
                  "include_in_all": false
               },
}
```

However, they are currently being submitted as

```json
{
          "value": {
            "type": "float"
          },
          "count": {
            "type": "float"
          }
}
```

Looking at https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html the types should be 

```json
{
          "value": {
            "type": "double"
          },
          "count": {
            "type": "integer"
          }
}
```

The value here, I wonder should be [`double` or maybe a `scaled_float`](http://stackoverflow.com/questions/23017583/is-javas-bigdecimal-the-closest-data-type-corresponding-to-cs-decimal) ?

Mpdreamz:
... to map inside the fluent properties descriptor

```csharp
.Properties(ps => ps
 	.Scalar(p => p.Int, m => m)
 	.Scalar(p => p.IntNullable, m => m)
 	.Scalar(p => p.Float, m => m)
 	.Scalar(p => p.FloatNullable, m => m)
 	.Scalar(p => p.Char, m => m)
 	.Scalar(p => p.CharNullable, m => m)
 	.Scalar(p => p.Guid, m => m)
///snip 	
	.Scalar(p => p.String, m => m)
 );
```

### DIFF
--- a/src/Nest/Mapping/DynamicTemplate/SingleMapping.cs
+++ b/src/Nest/Mapping/DynamicTemplate/SingleMapping.cs
@@ -33,7 +33,7 @@ namespace Nest
 
 		public IProperty Binary(Func<BinaryPropertyDescriptor<T>, IBinaryProperty> selector) =>
 			selector?.Invoke(new BinaryPropertyDescriptor<T>());
-
+		/
 		public IProperty Attachment(Func<AttachmentPropertyDescriptor<T>, IAttachmentProperty> selector) =>
 			selector?.Invoke(new AttachmentPropertyDescriptor<T>());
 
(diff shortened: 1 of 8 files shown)


Mpdreamz:
@gmarz I quite like the separate `Scalar` route with overloads `.Number(n=>` would be amiguous until either its closed by `)` or you type `,`. 

The other reason i quite like `Scalar` is that in fluent mapping it doesn't flow too well you have to know what type you are going to target before specifying the name.

With `.Scalar(` you can simply map 90% of the properties in a much more exploratory friendly way.

russcam:
Check CLS Compliance for this

src/Nest/Mapping/Types/Properties-Scalar.cs
```diff
@@ -178,6 +181,7 @@ public partial class PropertiesDescriptor<T> : IsADictionaryDescriptorBase<Prope
 
 		public PropertiesDescriptor<T> Scalar(Expression<Func<T, string>> field, Func<TextPropertyDescriptor<T>, ITextProperty> selector) =>
 			SetProperty(selector?.InvokeOrDefault(new TextPropertyDescriptor<T>().Name(field)));
```
russcam: This should do the same thing as string inference on automapping i.e. add a keyword multi_field

russcam:
Add `IEnumerable` overloads

gmarz:
@forloop I added IEnumerable overloads with my last commit.  Mind giving this a final review?

src/Nest/Mapping/DynamicTemplate/SingleMapping.cs
```diff
@@ -61,5 +66,157 @@ public IProperty Nested<TChild>(Func<NestedPropertyDescriptor<T, TChild>, INeste
 
 		public IProperty Percolator(Func<PercolatorPropertyDescriptor<T>, IPercolatorProperty> selector) =>
 			selector?.Invoke(new PercolatorPropertyDescriptor<T>());
+
+#pragma warning disable CS3001 // Argument type is not CLS-compliant
+		public IProperty Scalar(Expression<Func<T, int>> field, Func<NumberPropertyDescriptor<T>, INumberProperty> selector) =>
+			selector?.InvokeOrDefault(new NumberPropertyDescriptor<T>().Name(field).Type(NumberType.Integer));
```
russcam: I think all of these should be
```
selector.InvokeOrDefault(...)
```
if `selector` is `null`, we want to return the default, but the null conditional on the `.InvokeOrDefault()` will return `null` if `selector` is `null`.

Mpdreamz: yep! good catch will fix these now.

Mpdreamz:
merged to `master` and `5.x`

PR #2373, opened 2016-11-07
Fix script representation for bulk updates

### LINKED ISSUE #2357: Should BulkUpdateOperation & IBulkUpdateOperation be changed to use IScript?
NEST/Elasticsearch.Net version: 5.0.0-rc1
Elasticsearch version: 5.0.0

This is related to #2343 / #2342

Reading the spec at https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html
I think the change is similar - removal of properties for Lang, Script, ScriptId, ScriptFile, Params and replaced with an IScript Script property.

the currently generated request looks similar to this:
POST /test/doc/_bulk
{"update": { "_id":1}}
{"upsert":{"id":1,"script":""} }

and gives an error of "script or doc is missing"

but changing the properties in BulkUpdateOperation produces a request similar to

POST /test/doc/_bulk
{"update":{"_id":1}}
{"upsert":{"id":1},"script":{"inline":""}}

which returns with "Cannot generate an empty script." - which is expected!

russcam:
Parameters related to an update script should be serialized as properties of an object assigned to a script property.
Closes #2357

### DIFF
--- a/src/Nest/Document/Multiple/Bulk/BulkOperation/BulkUpdate.cs
+++ b/src/Nest/Document/Multiple/Bulk/BulkOperation/BulkUpdate.cs
@@ -8,22 +8,52 @@ namespace Nest
 		where TDocument : class
 		where TPartialDocument : class
 	{
+		/// <summary>
+		/// Infers the id of the object to update from the provided <param name="object">object</param>.
+		/// See <see cref="Doc"/> to apply a partial object merge.
+		/// </summary>
 		TDocument InferFrom { get; set; }
 
+		/// <summary>
+		/// A document to upsert when the specified document to be updated is not found
+		/// </summary>
 		TDocument Upsert { get; set; }
-		
+
+		/// <summary>
+		/// The partial update document to be merged on to the existing object.
+		/// </summary>
 		TPartialDocument Doc { get; set; }
-		
+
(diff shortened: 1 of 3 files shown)


src/Nest/Document/Multiple/Bulk/BulkOperation/BulkUpdate.cs
```diff
@@ -152,19 +239,74 @@ public class BulkUpdateDescriptor<TDocument, TPartialDocument>
...
+		/// </summary>
 		public BulkUpdateDescriptor<TDocument, TPartialDocument> Params(Func<FluentDictionary<string, object>, FluentDictionary<string, object>> paramDictionary) =>
 			Assign(a => a.Params = paramDictionary(new FluentDictionary<string, object>()));
 
+		/// <summary>
+		/// How many times an update should be retried in the case of a version conflict.
+		/// </summary>
 		public BulkUpdateDescriptor<TDocument, TPartialDocument> RetriesOnConflict(int? retriesOnConflict) => Assign(a => a.RetriesOnConflict = retriesOnConflict);
+
+		private IScript CreateScriptFromProperties()
```
Mpdreamz: Pragmatic fix++;

PR #2374, opened 2016-11-07
Fix script representation for bulk updates

### LINKED ISSUE #2357: Should BulkUpdateOperation & IBulkUpdateOperation be changed to use IScript?

russcam:
Parameters related to an update script should be serialized as properties of an object assigned to a script property.

Change the name of the `InferFrom` on bulk update operation to `IdFrom` to align naming with fluent method name and indicate that it sets the id.

Closes #2357

### DIFF
--- a/src/Tests/Document/Multiple/Bulk/BulkApiTests.cs
+++ b/src/Tests/Document/Multiple/Bulk/BulkApiTests.cs
@@ -61,6 +61,13 @@ namespace Tests.Document.Multiple.Bulk
 			new Dictionary<string, object>{ { "delete", new { _type="project", _id = Project.Instance.Name + "1" } } },
 			new Dictionary<string, object>{ { "create", new { _type="project", _id = Project.Instance.Name + "2" } } },
 			Project.InstanceAnonymous,
+			new Dictionary<string, object>{ { "update", new { _type="project", _id = Project.Instance.Name + "2" } } },
+			new Dictionary<string, object>{ { "script", new
+			{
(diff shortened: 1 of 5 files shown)


src/Tests/Document/Multiple/Bulk/BulkApiTests.cs
```diff
@@ -121,6 +143,7 @@ protected override void ExpectResponse(IBulkResponse response)
 
 			var project2 = this.Client.Source<Project>(Project.Instance.Name + "2", p => p.Index(CallIsolatedValue));
 			project2.Description.Should().Be("Default");
+			project2.NumberOfCommits.Should().Be(30);
```
Mpdreamz: 👍

PR #2379, opened 2016-11-07
Return of the <inheritdoc />

Mpdreamz:
image


Mpdreamz:
Pulled into master and 5.x

PR #2381, opened 2016-11-09
fix #2286 normalize ApiCall and CallDetails and add better xmldocs to IResponse

### LINKED ISSUE #2286: 5.x CallDetails vs ApiCall
There are a lot of missing xml docs and it's hard to tell what the difference is between these two.

Mpdreamz:
If you get a hold of an `IResponse` you see both `ApiCall` and `CallDetails` which are in fact the same property this is confusing.

In `Elasticsearch.Net` response object can implement `IBodyWithApiCallDetails` which is a special interface the low level client knows how to handle and set the diagnostics `IApiCallDetails` on a `CallDetails` property.  

In the high level client we did not want to expose a setter on this so we hid the `CallDetails` from `IBodyWithApiCallDetails` and introduced `ApiCall`. 

We now more explicitly hide `IBodyWithApiCallDetails` and normalize the properties to the same name.

Also introduced xmldocs on these which were long overdue.

### DIFF
--- a/src/Elasticsearch.Net/Responses/IBodyWithApiCallDetails.cs
+++ b/src/Elasticsearch.Net/Responses/IBodyWithApiCallDetails.cs
@@ -1,7 +1,14 @@
 namespace Elasticsearch.Net
 {
+	/// <summary>
+	/// Implementing this interface on your response objects will cause the low level client to
+	/// automatically set <see cref="IApiCallDetails"/> diagnostic information on the object.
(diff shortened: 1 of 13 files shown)


gmarz:
This was indeed confusing.  LGTM 👍

russcam:
Much clearer 👍 LGTM

Mpdreamz:
pulled and merged to 5.x

Issue #2299, opened 2016-09-30
5.x API Feedback All enumerable results should be immutable and always return a result


niemyjski:
In 1.x all collections were empty collections regardless of if the api result was valid or invalid.  In 5.x this is very different and personally I don't like it as it makes me question what the api is going to do. Let me explain..

Previously this code along with bulk results would return an empty collection

``` cs
            var alias = await _client.GetAliasAsync(descriptor => descriptor.Name(_configuration.DailyLogEvents.Name));
            _logger.Trace(() => alias.GetRequest());
            Assert.False(alias.IsValid);
            Assert.Equal(0, alias.Indices.Count);
```

Now `alias.Indicies` throws a null reference exception, the same with bulk operations... This leads us to have to do null checking everywhere and if blocks around foreachs...

What we have been doing in our api (apps and libs) is simple, we return readonly collections everywhere and in the case where there are no results (and it's an enumerable result) we return an shared empty result instance like:

``` cs
private static IReadOnlyCollection<Aliases> EmptyAliases = new List<Aliases>(0).AsReadOnly();
```

This really cuts down on extra required code and keeps allocations to a minimal. 

**TLDR:** Always return an empty read only collection instead of null + make everything immutable.

niemyjski:
This just bit us again with aggregations:

```cs

        public override void Visit(GroupNode node, IQueryVisitorContext context) {
            base.Visit(node, context);

            var elasticContext = context as IElasticQueryVisitorContext;
            if (elasticContext == null)
                throw new ArgumentException("Context must be of type IElasticQueryVisitorContext", nameof(context));

            var container = GetParentContainer(node, context);
            foreach (var child in node.Children.OfType<IFieldQueryNode>()) {
                var aggregation = child.GetAggregation(() => child.GetDefaultAggregation(context));
                if (aggregation != null) {
                    container.Aggregations[((IAggregation)aggregation).Name] = (AggregationContainer)aggregation;
                }
            }

            if (node.Parent == null)
                node.SetAggregation(container);
        }
```

container.Aggregations was previously initialized and now it's not.

Mpdreamz:
This is now implemented as per #2383, we we're not doing this consistently thanks @niemyjski  🍺 ++;

PR #2383, opened 2016-11-09
Fix/response :heart:

Mpdreamz:
Give all the responses some ❤️ 

As per #2299 be more consistent responses with returning empty collections vs null on responses. We did this sporadically but not consistently.
Now that we dropped .NET 4.0 we can make the response IEnumerable and Dictionaries IReadOnly*

Made sure all the interfaces only define getters and not setters


Mpdreamz:
Can you explain why you wonder that @gmarz? Interested to hear your thoughts on this.

gmarz:
@Mpdreamz Robustness principle http://stackoverflow.com/questions/17170/when-to-use-ilist-and-when-to-use-list

> On the other hand, when returning an object out of a function, you want to give the user the richest possible set of operations without them having to cast around.

Previously the richest types would be `List<T>` and `Dictionary<T>`, and now with this PR `ReadOnlyCollection<T>` and `ReadOnlyDictionary<T>`.

Mpdreamz:
Not sure how I feel about returning the concrete types since they expose a lot of operations we dont want to expose e.g `searchResponse.Documents.Add()`. I generally agree with Postel's law but I feel the concrete types set wrong expectations. Also it ties us down. 

Need to sleep on this :)

Mpdreamz:
@gmarz @russcam  rebased against master and fixed the flakey tests. 

Still some integration test failures but they are unrelated to this branch.

russcam:
Will have a look at this today @Mpdreamz

russcam:
7 failing integration tests:

- Tests.XPack.Security.User.PutUser.PutUserRunAsApiTests.ReturnsExpectedIsValid [FAIL]
- Tests.XPack.Security.User.PutUser.PutUserRunAsApiTests.HandlesStatusCode [FAIL]
- Tests.XPack.Security.Role.PutRole.PutRoleRunAsApiTests.HandlesStatusCode [FAIL]
- Tests.XPack.Security.Role.PutRole.PutRoleRunAsApiTests.ReturnsExpectedIsValid [FAIL]

related to the param for shield run as #2394 

- Tests.Indices.StatusManagement.Upgrade.UpgradeApiTests.ReturnsExpectedIsValid [FAIL]
- Tests.Indices.StatusManagement.Upgrade.UpgradeApiTests.HandlesStatusCode [FAIL]

related to allow_no_indices #2397

- Tests.Indices.Monitoring.IndicesShardStores.IndicesShardStoresApiTests.AssertResponse [FAIL]

not sure why this one is failing right now.  `Expected collection not to be empty.`

russcam:
LGTM - would be good to have some Coding Standards unit tests to ensure that this is the case across all `IResponse` types

Mpdreamz:
The IndicesShardResponse was an easy fix, the property now needed an explicit [JsonProperty], squashing and merging this one in.

PR #2399, opened 2016-11-21
Remove IDictionary from IIsADictionary<TKey,TValue> definition

russcam:
This removes any ambiguity when enumerating the items in an instance of `IIsADictionary<TKey,TValue>`. 
 
Any implementation of `IIsADictionary<TKey, TValue>` must always provide implementations for the contract of `IDictionary<TKey, TValue>` so there is little value in having the `IDictionary` contract where boxing and unboxing will be incurred, except to distinguish types that implement `IDictionary` that are not `IIsADictionary`.

Removes the cost boxing of values in `VerbatimDictionaryKeysConverter` by having `TKey` and `TValue` generic type parameters.

### DIFF
--- a/src/Nest/Indices/MappingManagement/PutMapping/PutMappingRequest.cs
+++ b/src/Nest/Indices/MappingManagement/PutMapping/PutMappingRequest.cs
@@ -35,7 +35,7 @@ namespace Nest
 		/// <inheritdoc/>
 		public IIndexField IndexField { get; set; }
 		/// <inheritdoc/>
-		public FluentDictionary<string, object> Meta { get; set; }
+		public IDictionary<string, object> Meta { get; set; }
 		/// <inheritdoc/>
 		public bool? NumericDetection { get; set; }
 		/// <inheritdoc/>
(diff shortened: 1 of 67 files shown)


src/Nest/Indices/MappingManagement/PutMapping/PutMappingRequest.cs
```diff
@@ -35,7 +35,7 @@ public partial class PutMappingRequest
 		/// <inheritdoc/>
 		public IIndexField IndexField { get; set; }
 		/// <inheritdoc/>
-		public FluentDictionary<string, object> Meta { get; set; }
+		public IDictionary<string, object> Meta { get; set; }
```
Mpdreamz: ooemf we had a fluent dictionary as a property 😢

src/Nest/Mapping/Types/Properties.cs
```diff
@@ -142,13 +143,13 @@ internal static IProperties AutoMap<T>(this IProperties existingProperties, IPro
 		{
 			var properties = new Properties();
 			var autoProperties = new PropertyWalker(typeof(T), visitor, maxRecursion).GetProperties();
-			foreach (var autoProperty in (IEnumerable<KeyValuePair<PropertyName, IProperty>>)autoProperties)
+			foreach (var autoProperty in autoProperties)
```
Mpdreamz: 🙈

russcam:
ported to `5.x` in https://github.com/elastic/elasticsearch-net/commit/4dd15772415c64b87e7122d650788b84e16cdc61

PR #2402, opened 2016-11-21
fix #2220 allow more control in BulkAll 

### LINKED ISSUE #2220: Allow per document index operations on BulkAll
The [BulkAll helper introduced in `master`](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Document/Multiple/BulkAll/ElasticClient-BulkAll.cs) is really helpful for setting up bulk indexing. It could be even more helpful if per document modifications were allowed to each bulk index operation. For example, it's not possible to bulk index a collection of child documents with the helper and set the `_parent` for each index operation.

Mpdreamz:
how the buffer of `IList<T>` buffer is tranlated into a single bulk request, rather then always simply calling `IndexMany` on it.

This allows you to do use e.g the `IndexMany(buffer, perDocumentOpSelector)` overload or manually looping over the buffer and doing a `Create/Update/Index`

### DIFF
--- a/src/Nest/Document/Multiple/BulkAll/BulkAllObservable.cs
+++ b/src/Nest/Document/Multiple/BulkAll/BulkAllObservable.cs
@@ -108,12 +108,15 @@ namespace Nest
 			var r = this._partionedBulkRequest;
 			var response = await this._client.BulkAsync(s =>
 			{
-				s.IndexMany(buffer);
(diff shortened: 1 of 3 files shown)


russcam:
This should also be ported to `2.x`

russcam:
cherry-picked to 5.x in 22f5b4291a826be9b0c39e565ed31c778747acb1

PR #2405, opened 2016-11-22
Assert all collections on Response types are read only

russcam:
Following on from merge of https://github.com/elastic/elasticsearch-net/pull/2399 into `master`.

- Reflection unit test traverses IResponse properties to check their types.
- Add Fluent Assertions for IReadOnlyDictionary<TKey, TValue>. See https://github.com/dennisdoomen/FluentAssertions/issues/357
- Update Watcher APIs to use readonly collections
- Use Actions for actions on Watch type
- Enable readonly collection deserialization in SimpleJson

Should we also assert that `IReadOnlyCollection<T>` and `IReadOnlyDictionary<TKey, TValue>` properties on responses are not `null` as a general rule?


russcam:
opened PR for `5.x`: #2407

PR #2418, opened 2016-11-28
Faster enum -> string resolution

### LINKED ISSUE #2400: Generated Enum specific Resolve() methods for known enums
[`KnownEnums.Resolve(Enum e)`](https://github.com/elastic/elasticsearch-net/blob/master/src/Elasticsearch.Net/Domain/Enums.Generated.cs#L391) can be used to get a string value for an enum value. We should generate enum specific `Resolve()` extension methods to avoid the type check and casting.

russcam:
- Specific enum `GetStringValue` extension methods for known enums
- Caching of generic method of resolving string values for enums based on enum type.
- Renamed `KnownEnums.Resolve` to `GetStringValue` and made an extension method. Allows enum specific overloads to be introduced where necessary
- Added GetStringValue extension method for HttpMethod - this is a hot path, called on every request.
- Fixed casing issue in HttpConnection for setting ContentLength

Closes #2400


Mpdreamz:
ported to `master` and `2.x`

PR #2420, opened 2016-11-28
Fix RequestData hashcode and equality for .NET core HttpConnection

### LINKED ISSUE #2417: each call to ElasticClient.Index uses up a file descriptor

5.0.0-rc3 from nuget

**Elasticsearch version**:

docker image: elasticsearch:5.0.1

Every time I call .Index, or .IndexAsync on ElasticClient it uses up a file descriptor (opens a TCP connection and leaves it open). My process soon quits.

**Steps to reproduce**:

i'm running .NET Core 1.1 on linux (dotnet:latest docker image).

```
  var settings = new ConnectionSettings(new Uri($"http://{elasticHost}:9200"))
    .DefaultIndex("myindex");
  var ec = new Nest.ElasticClient(settings);

    ...

  var r = ec.Index(elasticSummary);
```

as above

```
lsof -a -p [pid] 
...
dotnet  22944 root  203u  sock    0,8      0t0 738704 protocol: TCP
...
etc.

```

nothing relevant is logged by elastic search (using default settings in docker image).

russcam:
RequestData's GetHashCode implementation uses instances of types that will return different values per instance (Headers, BasicAuthenticationCredentials). GetHashCode and Equals only need to reference properties that are unchanging per HttpClient instance, which are:

RequestTimeout
HttpCompression
ProxyAddress
ProxyUsername
ProxyPassword
DisableAutomaticProxyDetection

All other properties are set per request.

Fixes #2417

### DIFF
--- a/src/Elasticsearch.Net/Connection/HttpConnection-CoreFx.cs
+++ b/src/Elasticsearch.Net/Connection/HttpConnection-CoreFx.cs
@@ -31,30 +31,32 @@ namespace Elasticsearch.Net
 	public class HttpConnection : IConnection
 	{
 		private readonly object _lock = new object();
+
 		private readonly ConcurrentDictionary<int, HttpClient> _clients = new ConcurrentDictionary<int, HttpClient>();
 
 		private HttpClient GetClient(RequestData requestData)
 		{
-			var hashCode = requestData.GetHashCode();
+			var key = GetClientKey(requestData);
 			HttpClient client;
-			if (this._clients.TryGetValue(hashCode, out client)) return client;
-			lock (_lock)
+			if (!this._clients.TryGetValue(key, out client))
 			{
-				if (this._clients.TryGetValue(hashCode, out client)) return client;
-
-				var handler = CreateHttpClientHandler(requestData);
-
-				client = new HttpClient(handler, false)
+				lock (_lock)
 				{
-					Timeout = requestData.RequestTimeout
-				};
+					client = this._clients.GetOrAdd(key, h =>
+					{
+						var handler = CreateHttpClientHandler(requestData);
+						var httpClient = new HttpClient(handler, false)
+						{
+							Timeout = requestData.RequestTimeout
+						};
 
-				client.DefaultRequestHeaders.ExpectContinue = false;
-
-				this._clients.TryAdd(hashCode, client);
-				return client;
+						httpClient.DefaultRequestHeaders.ExpectContinue = false;
+						return httpClient;
+					});
+				}
 			}
 
+			return client;
 		}
 
 		public virtual ElasticsearchResponse<TReturn> Request<TReturn>(RequestData requestData) where TReturn : class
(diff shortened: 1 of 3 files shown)


mhowlett:
I can verify that when run on my mac (in the vs code debugger):

1. hashCode is now the same for each request (whereas before it was not).
2. the call to ElasticClient.Index is working correctly. 

and in the docker container on my linux machine:

1. `lsof -a -p` is no longer showing a new line after each .Index call.

Code LGTM, though I notice you implement GetHashCode on BasicAuthenticationCredentials but it's no longer used.

russcam:
I'll remove the Equality members on `BasicAuthenticationCredentials` 👍 
[Also looks like `GetOrAdd()` can execute more than once when run simultaneously on different threads](https://msdn.microsoft.com/en-us/library/ee378677(v=vs.110).aspx#Anchor_2). I'll fix that too

src/Elasticsearch.Net/Transport/Pipeline/RequestData.cs
```diff
@@ -113,23 +113,13 @@ private string CreatePathWithQueryStrings(string path, IConnectionConfigurationV
 			return path;
 		}
 
-
 		protected bool Equals(RequestData other) =>
```
Mpdreamz: thought bubble, move GetHashCode() out of `RequestData` and into CoreFx's `HttpConnection` its cache is already `ConcurrentDictionary<int, HttpClient>` so we do not have to rely on `GetHashCode()` of `RequestData()`. 

Then we can get rid of the equals too and rely on reference comparison. Eventhough RequestData is immutable comparing two against eachother as values is not ever needed.

src/Elasticsearch.Net/Connection/HttpConnection-CoreFx.cs
```diff
@@ -31,30 +31,32 @@ internal class WebProxy : IWebProxy
...
-			if (this._clients.TryGetValue(hashCode, out client)) return client;
-			lock (_lock)
+			if (!this._clients.TryGetValue(hashCode, out client))
 			{
-				if (this._clients.TryGetValue(hashCode, out client)) return client;
-
-				var handler = CreateHttpClientHandler(requestData);
-
-				client = new HttpClient(handler, false)
+				lock (_lock)
```
Mpdreamz: ++ on keeping the lock.

src/Elasticsearch.Net/Transport/Pipeline/RequestData.cs
russcam: I was thinking exactly the same thing 😄 The Equality and HashCode are only used for this. 

I'll update now

Mpdreamz:
@forloop I know you have a local commit with additional tests but I am going to pull this in to all active branches so that we can spin this on myget. Mind sending that commit as a new PR?

Mpdreamz:
ported to `master` and `2.x`

russcam:
No worries; will submit a separate PR 👍

PR #2422, opened 2016-11-28
Use ScriptProcessorDescriptor in the fluent API

russcam:


Mpdreamz:
ported to `master`

russcam:
Sadly not 😢 We have a placeholder for that test which I'll look to add now

PR #2426, opened 2016-11-29
Remove ResolveException and let exceptions either bubble out or

### LINKED ISSUE #1958: Consider deprecating UnexpectedElasticsearchClientException
Currently, all uhandled exceptions bubble up to [Transport](https://github.com/elastic/elasticsearch-net/blob/master/src/Elasticsearch.Net/Transport/Transport.cs#L90) and are wrapped in an `UnexpectedElasticsearchClientException` before being re-thrown.

It feels like a borderline anti-pattern, and I'm on the fence as to whether or not this is really necessary.  It forces us to create special exceptions (for instance, [ResolveException](https://github.com/elastic/elasticsearch-net/blob/master/src/Elasticsearch.Net/Exceptions/ResolveException.cs)) that can occur late in the pipeline, which need to be caught and re-thrown when all we want to do is simply allow an `ArgumentException` to bubble out.  Why not just let "unexpected" exceptions to bubble out?  It will be clear enough from the exception and stack trace what the issue is.

On the other hand, we'll lose things like the audit trail and response, but how often are they useful for these types of exceptions?  They're more relevant for troubleshooting what went wrong in the request pipeline (`PipelineException`).  Is this additional information worth the overhead?

gmarz:
be wrapped within UnexpectedElasticsearchClientException if they occur
late in the transport pipeline.

Closes #1958


Mpdreamz:
ported to `master`

PR #2428, opened 2016-11-30
Add specific TermVectors type for MultiTermVectors

russcam:
In NEST 2.x, MultiTermVectorsResponse reuses TermVectorsResponse as the type in the collection, but this exposes other properties inherited from IResponse. Introduce a specific type for this that does not inherit from IResponse.

### DIFF
--- a/src/Nest/CommonAbstractions/SerializationBehavior/GenericJsonConverters/ReadOnlyCollectionJsonConverter.cs
+++ b/src/Nest/CommonAbstractions/SerializationBehavior/GenericJsonConverters/ReadOnlyCollectionJsonConverter.cs
@@ -1,0 +1,51 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using Newtonsoft.Json;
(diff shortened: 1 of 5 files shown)


src/Nest/CommonAbstractions/SerializationBehavior/GenericJsonConverters/ReadOnlyCollectionJsonConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using Newtonsoft.Json;
+
+namespace Nest
+{
+	internal class ReadOnlyCollectionJsonConverter<TDocument, TInterface> : JsonConverter
+		where TDocument : TInterface
+		where TInterface : class
+	{
```
Mpdreamz: CanRead => true

russcam:
merged to master in a33f2cbfdefbb1a22253afa7e40c019eb2112a13

Issue #2234, opened 2016-09-03
Boolean Should Implicit Casting Of QueryContainerDescriptor


aluitink:
**NEST/Elasticsearch.Net version**:
2.4.1

**Elasticsearch version**: 
2.3.3

**Description of the problem including expected versus actual behavior**:
NEST Client Should method signature that takes (params QueryContainer[] queries) cannot handle an array of QueryContainerDescriptors.

I'm required to pass an array of Func < QueryContainerDescriptor < T> , QueryContainer >.

I'm guessing this is because the method signature down casts them to the base type. Wondering if there is another way around this.

**Steps to reproduce**:

```
List<QueryContainerDescriptor<Relationship>> mailboxQueryContainers = new List<QueryContainerDescriptor<Relationship>>();

            foreach (Guid mailboxId in mailboxIds)
            {
                var qcd = new QueryContainerDescriptor<Relationship>();
                qcd.Term(r => r
                        .Field(f => f.MailboxId)
                        .Value(mailboxId)
                    );

                mailboxQueryContainers.Add(qcd);
            }

            var s = new SearchDescriptor<Item>()
                .Query(q => q
                    .HasParent<Relationship>(hp => hp
                        .Query(qq => qq
                            .Bool(b => b
                                .Should(mailboxQueryContainers.ToArray())
                                .MinimumShouldMatch(1)
                            )
                        )
                    )
                );

            using (MemoryStream ms = new MemoryStream())
            {
                _client.Serializer.Serialize(s, ms);
                ms.Position = 0;
                var json = Encoding.UTF8.GetString(ms.ToArray());
            }
```

Is this expected behavior?

russcam:
`QueryContainerDescriptor<T>` inherits from `QueryContainer` so you can instantiate a collection of `QueryContainer` and add `QueryContainerDescriptor<T>` instances to the collection.

An important point to note in your example is that **calling `.Term()` on `QueryContainerDescriptor<T>` returns a new instance of `QueryContainerDescriptor<T>`**, so you either need to chain the `.Term()` call onto the instantiation of the instance or assign the result of calling `.Term()` on `qcd` to `qcd`. Here's an example with the former

```c#
void Main()
{
    var client = new ElasticClient();

    var mailboxIds = new[]
    {
        Guid.NewGuid(),
        Guid.NewGuid(),
        Guid.NewGuid(),
        Guid.NewGuid(),
        Guid.NewGuid(),
    };

    // collection of QueryContainer
[...]

PR #2430, opened 2016-11-30
Do not instantiate QueryContainerDescriptor<T> unneccessarily

russcam:
Query methods on `QueryContainerDescriptor<T>` wrap the `IQuery` in a new instantiated `QueryContainerDescriptor<T>` each time. The `QueryContainerDescriptor<T>` instance on which the method is called can be used when it does not already contain a query.

Align `.Raw()` query method call with the rest of the query methods.

Found this when looking at #2234


russcam:
Following conversation, this doesn't/shouldn't need to be thread-safe.

russcam:
- ported to 5.x in  caaaaba15ba8109fc92a3678bda42bb6b58c86c5
- ported to master in dc7c0c75162ccce5ff8d4227bc9c6010e23db448

PR #2434, opened 2016-11-30
fix #2164 you can now skip the create index step if you use our Reind…

### LINKED ISSUE #2164: Reindex() into existing index
Our Reindex() helpers does not allow you to reindex into an existing index:

https://github.com/elastic/elasticsearch-net/issues/2039#issuecomment-210611104

Quite a few folks have asked for this in the past.

Mpdreamz:
…ex() helper, also made sure we use a _doc sort on the scroll


Mpdreamz:
++ to name change added that in as well, thanks @gmarz.

PR #2439, opened 2016-11-30
Fix #2336 add support for specifying the key type for a terms aggrega…

### LINKED ISSUE #2336: The key of an aggregation bucket is always read as a string
We're currently running into a situation where we need to have the key of a `terms` aggregation (which is a double in Elasticsearch) available as a double in our code, but the value of `Key` in a `KeyedBucket` is always a string, so we now have to parse many doubles from strings to accomplish this. But when I look at the response from Elasticsearch, the key does come back as a number there, it's the following code in Nest that always reads the key as a string:

https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Aggregations/AggregateJsonConverter.cs#L538

``` csharp
var key = reader.ReadAsString();
reader.Read();
var property = reader.Value as string;
if (property == "from" || property == "to")
    return GetRangeBucket(reader, serializer, key);

var keyItem = new KeyedBucket {Key = key};
```

My proposal would be to add a `KeyAsObject` property to `KeyedBucket` that simply gets the `reader.Value` assigned to it. And to avoid an unnecessary string allocation, the `Key` property should delegate to `KeyAsObject` and only create a string when you use the `Key` property.

``` csharp
private string _key = null;

public string Key 
{
    get
    {
        if(_key == null && KeyAsObject != null)
        {
            _key = KeyAsObject.ToString();
        }
    }
    set
    {
        _key = value;
    }
}
```

Please let me know your thoughts on this!

gmarz:
…tion

rather than always deserializing to a string, by making KeyedBucket
generic.

One can now do `response.Aggs.Terms<int>("my_terms")`

The non-generic helper `.Terms("my_terms")` still remains for strings as it's
the most common use case, and also so we don't break anyone.

Also removed `HistogramBucket`, as it can now be simplified to
`KeyedBucket<double>`.


src/Nest/Aggregations/AggregationsHelper.cs
```diff
@@ -177,16 +158,43 @@ private TAggregate TryGet<TAggregate>(string key)
...
+			{
+				Buckets = GetKeyedBuckets<TKey>(bucket.Items).ToList(),
+				Meta = bucket.Meta,
+			};
+		}
+
+
+		private IEnumerable<KeyedBucket<TKey>> GetKeyedBuckets<TKey>(IEnumerable<IBucket> items)
+		{
+			var buckets = items.OfType<KeyedBucket<object>>().ToList();
```
russcam: I think this can be `.Cast<KeyedBucket<object>>()` here as we know it will be `KeyedBucket<object>`; this way, we skip the type check and just cast directly

gmarz: 👍

gmarz:
merged to 5.x and master

russcam:
👍

PR #2440, opened 2016-11-30
fix #2438 make sure each async part of HttpConnection is abortable

### LINKED ISSUE #2438: Elasticsearch.Net Doesn't Time Out Async Calls (HttpWebRequest Usage Bug)
**NEST/Elasticsearch.Net version**: 2.4.7

**Elasticsearch version**: All Versions
  Async requests don't respect RequestTimeout set in ConnectionSettings (including the 60 second default). My team recently ran into network issues, meaning these requests to the cluster weren't completing, and this bug in ES.Net exacerbated the issue by pooling up an ever growing number of connections until we couldn't handle any more.

I located the bug here:
https://github.com/elastic/elasticsearch-net/blob/1de5538efda53dac24b024ce18c1b0c5012973ce/src/Elasticsearch.Net/Connection/HttpConnection.cs#L183

As you can see on MSDN, [HttpWebRequest.Timeout is NOT respected on async requests](https://msdn.microsoft.com/en-us/library/system.net.httpwebrequest.timeout(v=vs.110).aspx). Not sure the best solution, but one is to use HttpClient, which doesn't have this issue.

**Steps to reproduce**:
 1. Network issues or slow cluster response that causes a request to take longer than RequestTimeout in ConnectionSettings.
 2. The request does not time out.

Mpdreamz:
Everything old is new again, some lessons have to be learned over and over 😢 

In NEST `0.x` and `1.x` the[ HttpConnection async bits is implemented using a marvalous APM callback soup implementation](https://github.com/elastic/elasticsearch-net/blob/1.x/src/Elasticsearch.Net/Connection/HttpConnection.cs#L332) which was hard to read but battle hardened. 

With `2.x` we finally took the plunge to a TAP and async/await based implementation which is a joy to read. However in doing so we did not port all the timeouts registered with the threadpool to make the unabortable bits such as `GetRequestStreamAsync` and `GetResponseAsync`.

This PR brings back some APM wrapped to a Task with `Task.Factory.FromAsync` and `ThreadPool.RegisterWaitForSingleObject` is back in action to guard these calls. 

To test this works start an `5.x` elasticsearch server and use the AutoResponder in fiddler and set up the following rules:

image

Then using 

```csharp
{
	Task.Run(async () =>
	{
		var connectionSettings = new ConnectionSettings(new Uri("http://ipv4.fiddler:9200"))
			.RequestTimeout(TimeSpan.FromSeconds(3));
		var client = new ElasticClient(connectionSettings);

		//var response = client.RootNodeInfo();
		var response = await client.RootNodeInfoAsync();
		Console.WriteLine(response.DebugInformation);
		Console.ReadKey();
	}).Wait();
}
```

First using the latest `rc3` you'll see the synchronous code path adheres the timeout but the async code path does not. 

If you then add manual references to a build from this branch you'll see both behaving fine. 

If you uncheck the delay rule in fiddler you'll see both request succeeding again with no timeouts.

### DIFF
--- a/src/Elasticsearch.Net/Connection/HttpConnection.cs
+++ b/src/Elasticsearch.Net/Connection/HttpConnection.cs
@@ -155,6 +155,16 @@ namespace Elasticsearch.Net
 			return builder.ToResponse();
 		}
 
+
+		private static void RegisterApmTaskTimeout(IAsyncResult result, WebRequest request, RequestData requestData) =>
+			ThreadPool.RegisterWaitForSingleObject(result.AsyncWaitHandle, TimeoutCallback, request, requestData.RequestTimeout, true);
+
+		private static void TimeoutCallback(object state, bool timedOut)
+		{
+			if (!timedOut) return;
+			(state as WebRequest)?.Abort();
+		}
+
 		public virtual async Task<ElasticsearchResponse<TReturn>> RequestAsync<TReturn>(RequestData requestData, CancellationToken cancellationToken) where TReturn : class
 		{
 			var builder = new ResponseBuilder<TReturn>(requestData, cancellationToken);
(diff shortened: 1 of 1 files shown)


russcam:
LGTM 👍 will merge into 5.x and port to 2.x and master now

russcam:
- ported to 2.x in d377118618730516716717fa3d40facff5eb0cf7
- ported to master in e3d961bd43bb6e6f4d8d40fdd57c2b8797831de5

PR #2455, opened 2016-12-06
Make milliseconds nullable on Time

russcam:
Milliseconds do not apply when the Time is Month or Year.
Use constants for the milliseconds in each interval

### DIFF
--- a/src/Nest/CommonOptions/TimeUnit/Time.cs
+++ b/src/Nest/CommonOptions/TimeUnit/Time.cs
@@ -9,21 +9,20 @@ namespace Nest
 	[JsonConverter(typeof(TimeJsonConverter))]
 	public class Time : IComparable<Time>, IEquatable<Time>
 	{
-		private static readonly Regex _expressionRegex = new Regex(@"^(?<factor>[-+]?\d+(?:\.\d+)?)(?<interval>(?:y|M|w|d|h|m|s|ms))?$", RegexOptions.Compiled | RegexOptions.ExplicitCapture);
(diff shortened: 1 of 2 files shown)


src/Nest/CommonOptions/TimeUnit/Time.cs
```diff
@@ -9,21 +9,20 @@ namespace Nest
...
 	public class Time : IComparable<Time>, IEquatable<Time>
 	{
-		private static readonly Regex _expressionRegex = new Regex(@"^(?<factor>[-+]?\d+(?:\.\d+)?)(?<interval>(?:y|M|w|d|h|m|s|ms))?$", RegexOptions.Compiled | RegexOptions.ExplicitCapture);
+		private const double MillisecondsInAYearApproximate = MillisecondsInADay * 365;
+		private const double MillisecondsInAMonthApproximate = MillisecondsInADay * 30;
+		private const double MillisecondsInAWeek = MillisecondsInADay * 7;
+		private const double MillisecondsInADay = MillisecondsInAnHour * 24;
+		private const double MillisecondsInAnHour = MillisecondsInAMinute * 60;
+		private const double MillisecondsInAMinute = MillisecondsInASecond * 60;
+		private const double MillisecondsInASecond = 1000;
```
Mpdreamz: ++ we should be careful though with making consts out of things that could potentially change in between versions:
http://www.stum.de/2009/01/14/const-strings-a-very-convenient-way-to-shoot-yourself-in-the-foot/

Not the case here though

russcam: Agreed!

PR #2459, opened 2016-12-06
Add type to set auto expand replicas

russcam:


src/Nest/IndexModules/IndexSettings/Settings/AutoExpandReplicas.cs
```diff
@@ -0,0 +1,142 @@
...
+			get { return _maxReplicas; }
+			set
+			{
+				if (value == null && _minReplicas == null) _enabled = false;
+				else _enabled = true;
+				_maxReplicas = value;
+			}
+		}
+
+		public static AutoExpandReplicas Disabled() => new AutoExpandReplicas();
```
Mpdreamz: `AutoExpandReplicas` looks immutable to me so this could be a static property `{ get; } = new AutoExpandReplicas()` so that single instance could be reused.

russcam: ++ I'll make it immutable and make the disabled instance static 👍

PR #2462, opened 2016-12-06
ReindexAll() improvements

Mpdreamz:
`ReindexAll()` is now composed over the `IObservable`s from `ScrollAll()`
and `BulkAll()` taking advantage of both's build in concurrency.

Since the rate at which the scrolls produces far exceeds the rate at which we can consume them through bulks  `ReindexAll()` also implements a producerconsumer rate limiter. 
This is controlled by a `backPressureFactor` which controls the max amplification factor of running scrolls with the safe guard of:

`searchSize * maxConcurrency * backPressureFactor >= bulkSize`

Otherwise not enough scrolls would be spawned to feed a single bulk
request. Obviously if the scroll observable calls on complete it will
flush out any pending scrolls and feed it to `BulkAll()`.


Mpdreamz:
>remote: Resolving deltas: 100% (10/10), completed with 10 local objects.
>remote: Unexpected system error after push was received.
>remote: These changes may not be reflected on github.com!
>remote: Your unique error code: df69348cdee013b3cc32ec98aa672431

This PR seems to be missing my force push :( never seen this before

Mpdreamz:
OK that message can happen when github fails to update its caches on serverside commit hook. If you are looking at a diff on github.com that has limitter misspelled you are not reviewing the right commit. Please review locally. (commit list lists the correct sha numbers on here though).

src/Nest/CommonAbstractions/Infer/Inferrer.cs
```diff
@@ -31,6 +34,7 @@ public Inferrer(IConnectionSettingsValues connectionSettings)
 			this.CreateMultiHitDelegates = new ConcurrentDictionary<Type, Action<MultiGetHitJsonConverter.MultiHitTuple, JsonSerializer, ICollection<IMultiGetHit<object>>>>();
 			this.CreateSearchResponseDelegates = new ConcurrentDictionary<Type, Action<MultiSearchResponseJsonConverter.SearchHitTuple, JsonSerializer, IDictionary<string, object>>>();
 		}
+		public string Resolve(IUrlParameter urlParameter) => urlParameter.GetString(this._connectionSettings);
```
russcam: 👍

src/Nest/CommonAbstractions/Reactive/CoordinatedRequestObserverBase.cs
```diff
@@ -1,8 +1,15 @@
 ﻿using System;
 
-
-namespace Nest
+namespace Nest.CommonAbstractions.Reactive
```
russcam: namespace should be `Nest`

src/Nest/CommonAbstractions/Reactive/GetEnumerator.cs
```diff
@@ -0,0 +1,110 @@
+using System;
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Nest.CommonAbstractions.Reactive
```
russcam: namespace should be `Nest`

src/Nest/CommonAbstractions/Reactive/PartitionHelper.cs
```diff
@@ -0,0 +1,40 @@
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Nest.CommonAbstractions.Reactive
```
russcam: namespace should be `Nest`

src/Nest/CommonAbstractions/Reactive/ProducerConsumerBackPressure.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Nest.CommonAbstractions.Reactive
```
russcam: namespace should be `Nest`

src/Nest/CommonAbstractions/Reactive/ProducerConsumerBackPressure.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+	/// <summary>
+	/// Simple back pressure implementation that makes sure the minimum max concurrency between producer and consumer
+	/// is not amplified by the greedier of the by more then the backPressureFactor
```
russcam: "...greedier of the two..."

src/Nest/Document/Multiple/BulkAll/BulkAllRequest.cs
```diff
@@ -175,8 +185,19 @@ public BulkAllDescriptor(IEnumerable<T> documents)
...
+		public BulkAllDescriptor<T> BufferToBulk(Action<BulkDescriptor, IList<T>> modifier) => Assign(p => p.BufferToBulk = modifier);
+
+		/// <summary>
+		/// Simple back pressure implementation that makes sure the minimum max concurrency between producer and consumer
+		/// is not amplified by the greedier of the two by more then a given back pressure factor
+		/// When set each scroll request will additionally wait on <see cref="ProducerConsumerBackPressure.WaitAsync"/> as well as
+		/// <see cref="MaxDegreeOfParallelism"/> if set. Not that the consumer has to call <see cref="ProducerConsumerBackPressure.Release"/>
+		/// on the same instance every time it is done.
+		/// <param name="maxConcurrency">The minimum maximum concurrency which would be the bottleneck of the producer consumer pipeline</param>
+		/// <param name="backPressureFactor">The maximum amplification back pressure of the greedier part of the producer consumer pipeline</param>
```
russcam: I think we should point out what the default backPressureFactor is if unspecified.

Mpdreamz:
ported to `5.x` but was to eager and did not fix the xml doc warning and the test will do so now on both branches

PR #2468, opened 2016-12-07
fix #2448 favor publish_address in 5.x

### LINKED ISSUE #2448: Sniffing uses bound_address instead of publish_address
**Elasticsearch.Net version**: 5.0.0-rc4
**Elasticsearch version**: 5.0.1

I have Elasticsearch server bound on both loopback interface and LAN interface:
`network.host: ["_local_", "_site_"]`

Here is part of `/_nodes` response:
```
"http": {
    "bound_address": [
        "[::1]:9200",
        "127.0.0.1:9200",
        "10.9.0.129:9200"
    ],
    "publish_address": "10.9.0.129:9200",
    "max_content_length_in_bytes": 104857600
}
```

`SniffResponse` class uses the first value of `bound_address` array. As a result, when I use SniffingConnectionPool, it is initialized with server [::1]:9200, and all subsequent requests go to local machine. If I use only `_site_` binding, the client connects to the server normally.

Mpdreamz:
Now that 5.x no longer returns http_address we need to give precedence
to `publish_address` over `bound_address`

### DIFF
--- a/src/Elasticsearch.Net/Transport/Sniff/SniffResponse.cs
+++ b/src/Elasticsearch.Net/Transport/Sniff/SniffResponse.cs
@@ -5,35 +5,12 @@ using System.Text.RegularExpressions;
 
 namespace Elasticsearch.Net
 {
-	public class SniffResponse
+	public static class SniffParser
 	{
-		//internal ctor  - so that only Elasticsearch.Net can instantiate it
-		internal SniffResponse() { }
-
 		public static Regex AddressRegex { get; } = new Regex(@"^((?<fqdn>[^/]+)/)?(?<ip>[^:]+|\[[\da-fA-F:\.]+\]):(?<port>\d+)$");
-
-		public string cluster_name { get; set; }
-
-		public Dictionary<string, NodeInfo> nodes { get; set; }
-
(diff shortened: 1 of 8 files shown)


src/Elasticsearch.Net/Transport/Sniff/SniffResponse.cs
```diff
@@ -5,13 +5,29 @@
 
 namespace Elasticsearch.Net
 {
-	public class SniffResponse
+	public static class SniffParser
```
gmarz: 👍

Mpdreamz:
In the `2.x` branch we look at [http_address](https://github.com/elastic/elasticsearch-net/blob/2.x/src/Elasticsearch.Net/Transport/Sniff/SniffResponse.cs#L20)


which will yield `"http_address" : "elastic.co/35.160.254.14:9200",`

and an http section of:

```json
"http" : {
        "bound_address" : [ "127.0.0.1:9200", "[::1]:9200" ],
        "publish_address" : "elastic.co/35.160.254.14:9200",
        "max_content_length_in_bytes" : 104857600
}
```

Which is why this does not need to be ported to 2.x as we do the right thing there already

Mpdreamz:
ported to `5.x`

PR #2469, opened 2016-12-07
Quick sweap over all the TODO's that were quick to

Mpdreamz:
verify

also its sweep yeah i know 😢

### DIFF
--- a/src/Tests/Cluster/TaskManagement/TasksCancel/TasksCancelApiTests.cs
+++ b/src/Tests/Cluster/TaskManagement/TasksCancel/TasksCancelApiTests.cs
@@ -41,6 +41,7 @@ namespace Tests.Cluster.TaskManagement.TasksCancel
 				);
 
 				var taskId = reindex.Task;
+
 				//TODO change this to GetTasks when it's implemented
 				var taskInfo = client.ListTasks(new ListTasksRequest());
 				taskInfo.IsValid.Should().BeTrue();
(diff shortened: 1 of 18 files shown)


src/Tests/Framework/ManagedElasticsearch/Process/ElasticsearchNode.cs
```diff
@@ -239,9 +239,7 @@ private void ValidateLicense(IElasticClient client, XplatManualResetEvent handle
 
 			var exceptionMessageStart = "Server has license plugin installed, ";
 #if DOTNETCORE
-			//TODO Why is this here hardcoded ?
-			var licensePath = @"C:\license.json";
-			var licenseFile = File.Exists(licensePath) ? licensePath : string.Empty;
+			var licenseFile = Environment.GetEnvironmentVariable("ES_LICENSE_FILE");
```
russcam: ++

src/Tests/Cluster/TaskManagement/TasksCancel/TasksCancelApiTests.cs
```diff
@@ -41,6 +41,7 @@ protected override void IntegrationSetup(IElasticClient client, CallUniqueValues
 				);
 
 				var taskId = reindex.Task;
+
 				//TODO change this to GetTasks when it's implemented
```
gmarz: `GetTask` is implemented now, so we should be able to do `GetTask(taskId)`.

Issue #2248, opened 2016-09-13
The great documentation update


Mpdreamz:
We made great strides in 2.x having our documentation generated from actual code using roslyn to visit certain tests. 

This issue serves as an aggregate for remaining documentation issues so we can track it in one issue and make one big last swoop to get it into its final shape. 
- [x] Our docs are flattened on elastic.co site. We closely mimic elasticsearch's documentation tree but our generator loses much of this information. We need to be sure the documentation maintains the same tree in the generated asciidoc output. 
- [ ] We do not generate documentation for the API tests yet. The idea is to visit all  `Usage` and `Crud` tests and based on a consistent hash pick one of the 4 ways (sync/async and inititalizer/fluent) (was (#2239). 
- [ ] similarly all our analysis tests should end up in the documentation (#1991)
- [ ] For various reasons we would like to embedd tables (BenchMarkDotNet output for instance) our AsciiDoc parser does not handle this yet (https://github.com/russcam/asciidocnet/issues/2)
- [x] Revisit the flow of our Core docs (#2069)
- [x] Documentation: Clarify how a Connection Pool works and that its not responsible for the actual TCP connection pooling (#2260) 
- [x] Write a dedicated section on how to see what the client is doing (fiddler, logging, tracing, debuginformation, audits) (#864 #865)
- [ ] Add Compatibility Matrix from README to docs #2243
- [ ] Document Search Template #2729 
- [ ] Provide better context in `.doc.cs` cases see #3183 
- [ ] Document Suggesters (#2928)
- [ ] Improve Parent/Child documentation (#3542)

russcam:
Here's the rough Table of Contents for the new documentation

# NEST Documentation

- Introduction

- Breaking Changes
	- Elasticsearch.Net breaking changes
	- NEST breaking changes

- Elasticsearch.Net - Low level client
	- Getting started
		- Installing
		- Connecting
		- Calling an API
		- Providing a request body

- NEST - High level client
	- Getting started
		- Installing
		- Connecting
		- Indexing
		- Searching
		- Aggregations
	- Connecting
		- Configuration Options
		- Connection pools
		- Configuring SSL
		- Modifying the default connection
	- Serialization
		- Changing serializers
		- Modifying the default serializer
	- Mapping
		- Automapping
			- Inferred / Mapping from POCOs
			- Attributes
			- Fluent
			- Visitor pattern
		- Multi-fields
	- Analysis
		- Writing analyzers
		- Testing Analyzers
[...]

PR #2473, opened 2016-12-08
Limit the number of TCP connections on .NET core

russcam:
Give the same limit as in desktop clr. Might make sense to expose this as a configurable property in the future.
Remove the async action in the Parallel.For


russcam:
@Mpdreamz @gmarz I've added a `ConnectionLimit` value to `ConnectionConfiguration` with a default value of 80. Also added comments for all options; thinking that we can include these on a documentation page as part of #2248

russcam:
Rebasing and squashing

PR #2475, opened 2016-12-09
Add Create API

### LINKED ISSUE #2454: Add support for CreateRequest
https://www.elastic.co/guide/en/elasticsearch/reference/master/docs-index_.html#operation-type

currently unmapped

russcam:
Tidy up generated comments

Closes #2454


src/Nest/Document/Single/Create/CreateRequestJsonConverter.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using System.Linq;
+using System.Reflection;
+using Newtonsoft.Json;
+
+namespace Nest
+{
+	internal class CreateRequestJsonConverter : JsonConverter
```
gmarz: This is essentially identical to `IndexRequestJsonConverter`.  Can we make `IndexRequestJsonConverter` generic instead?

gmarz:
Minor comment regarding the JsonConverter, but otherwise LGTM.  We should probably back port this to 2.x as well.

src/Nest/Document/Single/Create/CreateRequestJsonConverter.cs
russcam: ++ will do 👍

PR #2477, opened 2016-12-09
found one remaining instance where an readonly* initialized property …

Mpdreamz:
…only has a getter, while this works and can also deserialize using black magic reflection an internal setter is nicer


Mpdreamz:
already ported to `5.x` by @russcam

PR #2480, opened 2016-12-11
Reindex now allows the target type to be different

Mpdreamz:
User has to supply an explicit map between TSource and TTarget, the old
overloads where TSource == TTarget still exists.

The simplified helpers for reindex contained a bug, out integration
tests only reindexed 2 documents but if you had more the backpressure
sanity checks would kick in complaining, these now run with sane defaults.

### DIFF
--- a/src/Nest/Document/Multiple/Reindex/ElasticClient-Reindex.cs
+++ b/src/Nest/Document/Multiple/Reindex/ElasticClient-Reindex.cs
@@ -9,10 +9,46 @@ namespace Nest
 		/// Helper method that allows you to reindex from one index into another using ScrollAll and BulkAll.
 		/// </summary>
 		/// <returns>An IObservable&lt;IReindexResponse&lt;T&gt;$gt; you can subscribe to to listen to the progress of the reindex process</returns>
-		IObservable<IBulkAllResponse> Reindex<T>(
-			Func<ReindexDescriptor<T>, IReindexRequest<T>> selector,
+		IObservable<IBulkAllResponse> Reindex<TSource,TTarget>(
+			Func<TSource, TTarget> mapper,
+			Func<ReindexDescriptor<TSource,TTarget>, IReindexRequest<TSource,TTarget>> selector,
 			CancellationToken cancellationToken = default(CancellationToken)
-		) where T : class;
+		)
+			where TSource : class
+			where TTarget : class;
+
+		/// <summary>
+		/// Helper method that allows you to reindex from one index into another using ScrollAll and BulkAll.
+		/// </summary>
+		/// <returns>An IObservable&lt;IReindexResponse&lt;T&gt;$gt; you can subscribe to to listen to the progress of the reindex process</returns>
+		IObservable<IBulkAllResponse> Reindex<TSource>(
+			Func<ReindexDescriptor<TSource, TSource>, IReindexRequest<TSource, TSource>> selector,
+			CancellationToken cancellationToken = default(CancellationToken)
+		)
+			where TSource : class;
+
+		/// <summary>
+		/// Helper method that allows you to reindex from one index into another using ScrollAll and BulkAll.
+		/// </summary>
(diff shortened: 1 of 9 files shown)


src/Nest/Document/Multiple/Reindex/ElasticClient-Reindex.cs
```diff
@@ -21,59 +57,126 @@ public partial interface IElasticClient
...
 			CancellationToken cancellationToken = default(CancellationToken)
-		) where T : class;
+		)
+			where TSource : class
+			where TTarget : class;
 
 		/// <summary>
-		/// Helper method that allows you to reindex from one index into another using ScrollAll and BulkAll.
+		/// Simplified form for reindex which will cover 80% of its usecases. Allows you to index all documents of type T from <paramref name="fromIndex" /> to <paramref name="toIndex" />
+		/// optionally limitting the documents found in <paramref name="fromIndex" /> by using <paramref name="selector"/>.
```
russcam: minor nitpick: spelling of `limiting`

src/Nest/Document/Multiple/Reindex/ElasticClient-Reindex.cs
```diff
@@ -21,59 +57,126 @@ public partial interface IElasticClient
...
+		)
+			where TTarget : class
+			where TSource : class
 		{
 			if (request.ScrollAll == null)
 				throw new ArgumentException(
 					"Reindex must have ScrollAll property set so we know how to get the source of the reindex operation");
 			if (request.BulkAll == null)
 				throw new ArgumentException(
 					"Reindex must have BulkAll property set so we know how to get the target of the reindex operation");
```
russcam: minor grammar nitpick: I think we should remove the subject pronouns in exception messages, for example
instead of:

> Reindex must have BulkAll property set so we know how to get the target of the reindex operation

can be

> BulkAll property must set in order to get the target of a Reindex operation

src/Nest/Document/Multiple/Reindex/ReindexObservable.cs
```diff
@@ -151,7 +169,7 @@ private IScrollAllResponse<T> ObserveScrollAllResponse(IScrollAllResponse<T> res
 
 		private int CreateIndex(string toIndex, IScrollAllRequest scrollAll)
 		{
-			var fromIndices = this._reindexRequest.ScrollAll?.Search?.Index ?? Indices<T>();
+			var fromIndices = this._reindexRequest.ScrollAll?.Search?.Index ?? Indices<TSource>();
 
 			if (string.IsNullOrEmpty(toIndex))
 				throw new Exception($"Can not resolve the target index name to reindex to make sure the bulk all operation describes one");
```
russcam: should we throw a more specific exception here?

src/Nest/Document/Multiple/Reindex/ReindexRequest.cs
```diff
@@ -77,49 +88,65 @@ public class ReindexRequest<T> : IReindexRequest<T> where T : class
 		/// <inheritdoc/>
 		/// <param name="scrollSource">The scroll operation yielding the source documents for the reindex operation</param>
 		/// <param name="bulkAllTarget">A factory that instantiates the bulk all operation over the lazy stream of search result hits</param>
-		public ReindexRequest(IScrollAllRequest scrollSource, Func<IEnumerable<IHit<T>>, IBulkAllRequest<IHit<T>>> bulkAllTarget)
+		public ReindexRequest(IScrollAllRequest scrollSource, Func<TSource, TTarget> map, Func<IEnumerable<IHitMetadata<TTarget>>, IBulkAllRequest<IHitMetadata<TTarget>>> bulkAllTarget)
 		{
 			scrollSource.ThrowIfNull(nameof(scrollSource), "without a scroll all request we don't know where to read from!");
```
russcam: I think we should remove the subject pronouns in the exception messages here as well

Mpdreamz:
ported to `5.x`

PR #2482, opened 2016-12-12
Use SemanticVersioning nuget package

russcam:
Supports .NET Core
Remove SemVer class as no longer required


Mpdreamz:
ported to `master` and `2.x`

PR #2495, opened 2016-12-14
Add important admonition about using System.Decimal POCO property types

### LINKED ISSUE #2463: 5.0 RC4 - Scalar mappings are storing Decimal as double
I noticed an issue where you can't round trip  `Decimal.MaxValue` as the mapping type is double. This occurs due to an overflow (`Value was either too large or too small for a Decimal.`). It might be good to have unit tests for all the c# numeric types and do max value of each type and then call AutoMap / Scalar and then attempt to round trip them. As precision will be lost or gained and it should be handled :).

The following snippet shows off what's happening in the simplest of terms (http://stackoverflow.com/questions/21337241/convert-from-decimal-to-single-and-back-why-does-it-throw-an-overflow-exception).
```cs
double dbl = (double)Decimal.MaxValue;
decimal dec = (decimal)dbl;
```

I think this has to due with two things that should be noted. The types supported by elastic don't always match up to .NET Types and serialization.

russcam:
Bump AsciiDoctNet to support additional asciidoc syntax
Closes #2463

### DIFF
--- a/docs/aggregations/writing-aggregations.asciidoc
+++ b/docs/aggregations/writing-aggregations.asciidoc
@@ -144,7 +144,6 @@ return s => s
     );
 ----
 <1> a list of aggregation functions to apply
-
 <2> Using LINQ's `Aggregate()` function to accumulate/apply all of the aggregation functions
 
 Combining multipe `AggregationDescriptor`'s is also possible using the bitwise `&` operator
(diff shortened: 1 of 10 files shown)


Mpdreamz:
ported to `master` and `5.x`

PR #2506, opened 2016-12-21
Mark type level analyzers as obsolete and add a convenience method fo…

gmarz:
…r setting an index-level default

They've been removed since 2.0 :\

I've also added a `Default()` method for conveniently setting the index-level default analyzer rather than having to do `.Keyword("default")` which might be a little more intuitive.


Mpdreamz:
Ported to `2.x` and `master`

PR #2510, opened 2016-12-22
Prevent multiple date time allocations

niemyjski:
1. DateTime.SpecifyKind will return a new DateTime instance.
2. Creating new DateTime instance for Epoch,
3. Adding Milliseconds creates a new DateTime Instance

Now only one instance is created (apart from the one time static hit)


Mpdreamz:
ported to `2.x` and `5.x`

PR #2514, opened 2016-12-23
Fix #2513 add TermVector option to Text attribute

### LINKED ISSUE #2513: 5.X Is TermVector intentionally not exposed on Nest.TextAttribute?
Seems like it's missing the public property to allow us to set it:

```c#
[Text(TermVector = TermVectorOption.WithPositionsOffsets)] 
```

If it just needs adding, I'll submit a PR to add a public property that wraps the explicit interface property - just like the others do!

gmarz:


Mpdreamz:
ported to `master`

PR #2515, opened 2016-12-23
Add dedicated match_phrase and match_phrase_prefix queries

### LINKED ISSUE #2509: MatchPhrasePrefix is deprecated
**NEST/Elasticsearch.Net version**: 5.0

**Elasticsearch version**: 5.1.1

Using Elasticsearch 5.1.1 with the MatchPhrasePrefix operator, I get deprecation warnings

`Deprecated field [type] used, replaced by [match_phrase and match_phrase_prefix query]`

**Steps to reproduce**:
 1. Create a query using the MatchPhrasePrefix operator

Examining the generated query, it looks like this 
`     "match": {
                            "navn": {
                              "type": "phrase_prefix",
                              "query": "Rasta"                        }                          }`
The new and improved syntax should be
`  "match_phrase_prefix": {    "navn": {                         
                              "query": "Rasta"
                            }
                          }`

gmarz:
And removed deprecated type field and irrelevant properties from the match query.

@Mpdreamz @russcam Should we back port this to 5.x?

Closes #2509


src/Tests/QueryDsl/Compound/FullText/MatchPhrasePrefix/MatchPhrasePrefixUsageTests.cs
```diff
@@ -3,33 +3,24 @@
...
 				{
 					_name = "named_query",
 					boost = 1.1,
 					query = "hello worl",
 					analyzer = "standard",
-					fuzzy_rewrite = "constant_score_boolean",
-					fuzziness = "AUTO",
-					fuzzy_transpositions = true,
-					cutoff_frequency = 0.001,
 					prefix_length = 2,
```
Mpdreamz: https://github.com/elastic/elasticsearch/blob/6d987a9b699c40376667fbd47ac8dda767ba1289/core/src/main/java/org/elasticsearch/index/query/MatchPhrasePrefixQueryBuilder.java#L203

Is this still a valid option?

gmarz: Nope, good catch.  It was also missing the slop option.  Fixed with my last commit.

gmarz:
Merged to master.  Agreed, I don't think there's a good way to back port this unfortunately.

PR #2521, opened 2016-12-28
Added ToString to Nest.CausedBy for proper string representation.

kirilyuro:
Nest.CausedBy didn't have a ToString override, and hence had the default "[Nest.CausedBy]" string representation.
This also caused non-informative string representations of BulkError: 
Type: "type" Reason: "reason" CausedBy: "[Nest.CausedBy]".

Now, string representation of Nest.CausedBy will contain the full cause chain (i.e., including the inner causes, in a way similar to a stack trace).

Should also be ported to 2.x & 5.x.

### DIFF
--- a/src/Nest/CommonOptions/Failures/BulkError.cs
+++ b/src/Nest/CommonOptions/Failures/BulkError.cs
@@ -20,6 +20,13 @@ namespace Nest
 		[JsonProperty("caused_by")]
 		public CausedBy CausedBy { get; internal set; }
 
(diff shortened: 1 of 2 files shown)


Mpdreamz:
Hey @kirilyuro I'm not seeing the additions to `CausedBy` in this commit can you double check?

Thanks for the PR 👍

kirilyuro:
The addition is the `ToString` override in `src/Nest/CommonOptions/Failures/ShardFailure.cs` (which contains the definition of `Nest.CausedBy`).

Mpdreamz:
Ported to `2.x` and `5.x` thanks again @kirilyuro

Issue #2395, opened 2016-11-20
Use FAKE's DotNetCli helper


russcam:
[Use the DotNetCli helper that is now part of FAKE](http://fsharp.github.io/FAKE/apidocs/fake-dotnetcli.html)

Mpdreamz:
👍 I'd wait though until the tooling itself RTM's in which case we can take the plunge fully and jump on the `new new csproj` files with VS2017 only support and a single `sln` file.

Mpdreamz:
Implemented as per #2523

PR #2523, opened 2016-12-30
Improvement/build

Mpdreamz:
- Normalize module vs type usages in fsx files
- unit methods now all in Method() = form
- cleaned up build.bat and its skiptests leniency, now only accepted as
  the first argument
- Improved usage notice in build.bat
- Building now checks if `dotnet.exe` resolves to using the proper SDK
- Moved to DotNetCli FAKE helper (#2395)
- Separated Projects and Tooling out of Paths.fsx
- Now have a truly quick compile target
  `build quick [test_filter]` does an incremental build and runs the
  unit tests from the bin output folder. it also skips paket.
- moved global.json to the root of repos
- unit tests now parallizes over all tests ignoring all clusters
  rather then piggy backing on xunit which we do in 5.x (our own routine
  is 10+seconds faster) or on the integration routine (which we
  currently do in master)


Mpdreamz:
@gmarz do you also get this after callinig `build` first? Had to upgrade to a later version of FAKE and `test` skips restore.

gmarz:
I did try calling `build` initially and was hitting the same error.  Just did a `git clean -fxd` and ran `build` again and now I'm hitting

```
build.bat: target="build" version= esversions= skiptests=0 apiKey="<empty>" feed="elasticsearch-net" escluster= testfilter=
The system cannot find the path specified.
```

gmarz:
Looks like [this line](https://github.com/elastic/elasticsearch-net/blob/improvement/build/build.bat#L40) is incorrectly setting SKIPPAKET

gmarz:
Sorry, it's not incorrectly setting SKIPPAKET.  It's just that we never run paket restore for the first time.  I'm not sure if SKIPPAKET is really useful, since bootstrapping is always required whether we're running "quick" or not.

Mpdreamz:
Updated this PR with some clean checkout and mixing regular with quick builds on the command line resiliency changes.

gmarz:
LGTM!

Mpdreamz:
ported to `2.x` and `master`

PR #2530, opened 2017-01-10
Index JObject and collection of JObjects

russcam:
Elasticsearch.Net does not have a dependency on Json.NET, so use the fully qualified type name to check the type is not `JObject`.

### DIFF
--- a/src/Elasticsearch.Net/Connection/RequestHandlers/RequestHandlerBase.cs
+++ b/src/Elasticsearch.Net/Connection/RequestHandlers/RequestHandlerBase.cs
@@ -64,7 +64,11 @@ namespace Elasticsearch.Net.Connection.RequestHandlers
 				return (string.Join("\n", ss) + "\n").Utf8Bytes();
 			}
 
-			if (typeof(IEnumerable<object>).IsAssignableFrom(dataType))
+			// JObject should not be treated as an IEnumerable<object> as doing so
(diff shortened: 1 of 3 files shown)


src/Elasticsearch.Net/Connection/RequestHandlers/RequestHandlerBase.cs
```diff
@@ -64,7 +64,8 @@ protected byte[] PostData(object data)
 				return (string.Join("\n", ss) + "\n").Utf8Bytes();
 			}
 
-			if (typeof(IEnumerable<object>).IsAssignableFrom(dataType))
+			if (typeof(IEnumerable<object>).IsAssignableFrom(dataType) && 
+				dataType.FullName != "Newtonsoft.Json.Linq.JObject")
```
gmarz: LGTM.  Maybe just add a comment to clarify why we have this check.

gmarz:
Also, do we need to port this to 2.x/5.x?

russcam:
2.x and 5.x handle `JObject` correctly; the refactoring of the pipeline for 2.x means that `JObject` ends up being handled correctly. We have tests in place for `JObject` in 2.x +: https://github.com/elastic/elasticsearch-net/blob/2.x/src/Tests/Document/Single/Index/IndexApiTests.cs#L132

PR #2532, opened 2017-01-11
Support GeometryCollection geo_shape types and queries

### LINKED ISSUE #2387: Add support for GeometryCollection
https://www.elastic.co/guide/en/elasticsearch/reference/5.x/geo-shape.html#_ulink_url_http_geojson_org_geojson_spec_html_geometrycollection_geometry_collection_ulink

russcam:
Tidy up xml comments on queries
Fix warnings

Closes #2387

### DIFF
--- a/src/Nest/QueryDsl/Geo/GeoCoordinateJsonConverter.cs
+++ b/src/Nest/QueryDsl/Geo/GeoCoordinateJsonConverter.cs
@@ -8,7 +8,7 @@ namespace Nest
 		public override bool CanConvert(Type objectType) => true;
 		public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
 		{
-			var p = value as GeoCoordinate;
+			var p = (GeoCoordinate)value;
 			if (p == null)
 			{
 				writer.WriteNull();
@@ -27,6 +27,5 @@ namespace Nest
 			if (doubles.Length != 2) return null;
 			return new GeoCoordinate(doubles[1], doubles[0]);
 		}
-
 	}
 }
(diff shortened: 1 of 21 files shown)


src/Nest/QueryDsl/Geo/GeoCoordinateJsonConverter.cs
```diff
@@ -8,7 +8,7 @@ internal class GeoCoordinateJsonConverter : JsonConverter
 		public override bool CanConvert(Type objectType) => true;
 		public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
 		{
-			var p = value as GeoCoordinate;
+			var p = (GeoCoordinate)value;
```
gmarz: Will this blow up if `value` isn't a `GeoCoordinate`?  Not sure if the null check makes sense anymore with the direct cast.

gmarz:
Minor comment, but otherwise LGTM.

src/Nest/QueryDsl/Geo/GeoCoordinateJsonConverter.cs
russcam: It would. yes 😄 

My thinking for converting to a direct cast however is that the converter is internal so we have control over which types it is applied to, so we should only be applying it to `GeoCoordinate`. If it's applied to another type other than `GeoCoordinate`, I think it would be better to have the invalid cast exception (to fix) rather than swallowing. In addition, the direct cast will slightly faster than `as`, which is effectively an `is` check and then cast.

gmarz: ++ but I don't think the null check makes sense anymore?

russcam: The null check should still be there because it could be passed a null `GeoCoordinate` still

gmarz: ah, ok 👍

PR #2537, opened 2017-01-13
Limit NEST 5.x compatibility to Elasticsearch.Net 5.x versions

russcam:


Mpdreamz:
LGTM is there anyway we can make the build smarter here as well so we do not need to bump on each major version?

russcam:
I have a fix for this

russcam:
I've updated the commit to pass the next major version and current year to nuget pack.

build/scripts/Releasing.fsx
```diff
@@ -23,7 +30,33 @@ module Release =
             let nuspec = (sprintf @"build\%s.nuspec" name)
             let nugetOutFile =  Paths.Output(sprintf "%s.%s.nupkg" name Versioning.FileVersion)
 
-            Tooling.Nuget.Exec ["pack"; nuspec; "-version"; Versioning.FileVersion; "-outputdirectory"; Paths.BuildOutput; ] |> ignore
+            let nextMajorVersion =
```
Mpdreamz: 👍 very nice!

Mpdreamz:
ported to `master`

PR #2538, opened 2017-01-13
Add debugging mode to connection settings

### LINKED ISSUE #2535: Shower thought: Debug mode?
Often when we're helping someone troubleshoot an issue we ask them to set `DisableDirectStreaming()` and also `PrettyJson()`.  This can sometimes be verbose and not very intuitive.

Thoughts on adding something more intuitive like `.DebugMode()` or `.EnableDebugging()` which encapsulates both of these, and any potential future features we add that aid in debugging (like https://github.com/elastic/elasticsearch-net/issues/1762 for instance)?

gmarz:
This commit adds the ability to set .EnableDebugMode() on
ConnectionSettings which turns direct streaming and pretty JSON on and
writes DebugInformation to the output window.

It also accepts an optional Action<IApiCallDetails> to perform when the
request is completed, but preserves the original OnRequestCompleted action
that might be set on ConnectionSettings.

Closes #2535

### DIFF
--- a/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
+++ b/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Specialized;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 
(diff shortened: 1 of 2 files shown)


src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
```diff
@@ -363,6 +363,25 @@ public T Proxy(Uri proxyAdress, string username, string password)
...
+		{
+			this._disableDirectStreaming = true;
+			this._prettyJson = true;
+
+			var originalCompletedRequestHandler = this._completedRequestHandler;
+			this._completedRequestHandler = d =>
+			{
+				originalCompletedRequestHandler?.Invoke(d);
+				onRequestCompleted?.Invoke(d);
+				System.Diagnostics.Debug.WriteLine(d.DebugInformation);
```
Mpdreamz: I think we need to let the user do this in their delegate themselves

gmarz: Write to the debug output, or preserve the original request handler?  Or both? :)

Mpdreamz: Write to debug output :)

gmarz: I sort of like that we write to it automatically as part of "debug mode", but I see the argument against it.  I'll remove it and update the PR 👍.

gmarz: Or we can make the default handler write the debug info, and let the user supplied one override it?

Mpdreamz: that works for me  👍

gmarz: done with my last commit

Mpdreamz:
ported to `2.x` and `5.x`

PR #2542, opened 2017-01-17
Generate common query parameters from the REST spec

### LINKED ISSUE #2541: Generate global query parameters from the spec
Now that https://github.com/elastic/elasticsearch/pull/22569 is merged, we no longer have to manually maintain global query string parameters and can remove [GlobalQueryStringParameters](https://github.com/elastic/elasticsearch-net/blob/master/src/CodeGeneration/ApiGenerator/Overrides/Global/GlobalQueryParameters.cs).  We should now generate them directly from the spec.

gmarz:
Closes #2541


Mpdreamz:
This looks good to me but mind splitting up the code changes to the generator with the generator output? 

Will make it easier not to include the breaking changes from `master` to `5.x`

gmarz:
@Mpdreamz good call, I've updated the PR.  I'll port to 5.x once ES 5.3 drops (since the _common change didn't make it into 5.2).

gmarz:
Merged to master and 5.x.  I did not update the spec on 5.x just yet though.

PR #2543, opened 2017-01-18
Count requests without a query should default to GET fix #2540

### LINKED ISSUE #2540: Count with no body sends POST, but could send GET
I use NEST 2.4.6 to get count:
```csharp
var count = this.elasticClient.Count<T>(x => x.Index("myindex").Type<T>()).Count;
```
It sends POST with
```json
{}
```

I believe it could send GET in this case. This may be fruitful if my credentials prevent POST, but I still need some overall count for application side optimizations/estimates and I do not want to go low level and POST adds slightly more performance overhead above GET and I want to copy/paste NEST debug output my browser. I believe it is possible to infer that it may be GET, not POST.

Mpdreamz:


Mpdreamz:
ported to `master` and `5.x` in those branches we were always sending an explicit `match_all` due to:

https://github.com/elastic/elasticsearch/issues/19422

But the fix for that went into 5.0 GA.

Issue #2493, opened 2016-12-13
5.0: Add constants for known date formatters


niemyjski:
I think it would be helpful to provide some known constant values for things like date formatters: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html#built-in-date-formats

What are your thoughts on this? Also, should a default one be sent in elastic 5 (dateOptionalTime was sent in 1.x).

russcam:
Known date formats seems like a good idea 👍

gmarz:
+1 on adding a `DateFormat` enum.  This means adding an overload everywhere we take a string though.

Mpdreamz:
We now have constants on `DateFormat.*` thanks for the suggestion @niemyjski 👍

PR #2544, opened 2017-01-18
added DateFormat constants with xmldocs

Mpdreamz:
fix for #2493

### DIFF
--- a/src/Nest/CommonOptions/DateFormat/DateFormat.cs
+++ b/src/Nest/CommonOptions/DateFormat/DateFormat.cs
@@ -1,0 +1,162 @@
+﻿namespace Nest.CommonOptions.DateFormat
+{
+	// ReSharper disable InconsistentNaming
+	//these const reflect their output on purpose
+	public static class DateFormat
+	{
+		///<summary>A formatter for the number of milliseconds since the epoch. Note, that this timestamp is subject to the limits of a Java Long.MIN_VALUE and Long.MAX_VALUE.</summary>
+		public const string epoch_millis = "epoch_mills";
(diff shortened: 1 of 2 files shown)


src/Nest/CommonOptions/DateFormat/DateFormat.cs
```diff
@@ -0,0 +1,162 @@
+﻿namespace Nest.CommonOptions.DateFormat
```
russcam: namespace should be `"Nest"`

src/Nest/CommonOptions/DateFormat/DateFormat.cs
```diff
@@ -0,0 +1,162 @@
...
+{
+	// ReSharper disable InconsistentNaming
+	//these const reflect their output on purpose
+	public static class DateFormat
+	{
+		///<summary>A formatter for the number of milliseconds since the epoch. Note, that this timestamp is subject to the limits of a Java Long.MIN_VALUE and Long.MAX_VALUE.</summary>
+		public const string epoch_millis = "epoch_mills";
```
russcam: Strong opinions about using the original names v.s. idiomatic .NET Pascal casing?

Mpdreamz: kinda :) `EpochMillis(seconds)` not that bad but turning `strict_t_time` into `StrictTTIme` :/ their is litttle guessing or ambiguity in keeping the casing which is a big plus over dogmatic casing, What do you  guys reckon?

gmarz: I'm cool with using the original names.

Mpdreamz:
backported to `5.x` and `2.x`

PR #2551, opened 2017-01-24
improve #2531, throw ArgumentException when troublesome agg keys are …

Mpdreamz:
…used

PR #2554, opened 2017-01-25
add meta as reserved aggregation key name

Mpdreamz:
Already added this in `2.x` as part of the backport of #2551


Mpdreamz:
ported to `5.x`

PR #2558, opened 2017-01-27
fix #1725 add support for ICU OOTB

### LINKED ISSUE #1725: IcuCollation Token Filter
After installing the analysis icu plugin I tried to add a icu collation token filter in the index settings via NEST. Unfortunately there is no way I see to set custom filter. It is possible to add this class?
Something like this should work:

```
public class IcuCollationTokenFilter : TokenFilterBase
{
    public IcuCollationTokenFilter()
        : base("icu_collation")
    {
    }

    [JsonProperty("language")]

    [JsonProperty("country")]
    public string Country { get; set; }
}
```

Mpdreamz:
We now finally support `analysis-icu` plugin OOTB


src/Nest/Analysis/Plugins/Icu/IcuCollationTokenFilter.cs
```diff
@@ -0,0 +1,149 @@
...
+	/// <summary>
+	/// Collations are used for sorting documents in a language-specific word order. The icu_collation token filter is available to all indices and defaults to using the DUCET collation, which is a best-effort attempt at language-neutral sorting.
+	/// Part of the `analysis-icu` plugin: https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-icu.html
+	/// </summary>
+	public interface IIcuCollationTokenFilter : ITokenFilter
+	{
+		[JsonProperty("language")]
+		string Language { get; set; }
+
+		string Country { get; set; }
```
gmarz: Missing JsonProperty?

src/Nest/Analysis/Plugins/Icu/Collation/IcuCollationAlternate.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+	/// <summary>
+	/// Sets the alternate handling for strength quaternary to be either shifted or non-ignorable.
+	/// Which boils down to ignoring punctuation and whitespace.
+	/// </summary>
+	[JsonConverter(typeof(StringEnumConverter))]
+	public enum IcuCollationAlternate
+	{
+		[EnumMember(Value="shifted")] Shifted,
+		[EnumMember(Value="non-ignorable")] NonIgnorable
```
russcam: hyphen instead of underscore?

src/Nest/Analysis/Plugins/Icu/IcuCollationTokenFilter.cs
Mpdreamz: Good catch!

src/Nest/Analysis/Plugins/Icu/Collation/IcuCollationAlternate.cs
Mpdreamz: The ICU analysis plugin has some inconsistencies like this:

https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-icu-collation.html

https://github.com/elastic/elasticsearch/blob/master/plugins/analysis-icu/src/main/java/org/elasticsearch/index/analysis/IcuCollationTokenFilterFactory.java#L127

I opened a couple issues on elasticsearch for this but not yet for this option will do so now.

Mpdreamz:
Updated with PR feedback and rebase against master

Mpdreamz:
ported to `2.x` and `5.x`

PR #2560, opened 2017-01-27
fix #2556 when writing out reflected properties from AutoMap() check …

### LINKED ISSUE #2556: InferMappingFor.Ignore for inherited property
**NEST/Elasticsearch.Net version**: 5.0.1 (and also in 2.4)

I'd like to ignore an inherited property without using attributes. `InferMappingFor` is the other option then, however I'm only able to ignore declared properties. 

**Steps to reproduce**:

The snippet below makes a mapping with a single inheritedProp-field, instead of no fields, as I expected.

```c#
void Main()
{
	new ElasticClient(new ConnectionSettings(new Uri("http://localhost:9200"))
		.InferMappingFor<Foo>(i => i
			.Ignore(f => f.InheritedProp)
			.Ignore(f => f.DeclaredProp))
	).CreateIndex("fooindex", c => c.Settings(s => s.NumberOfShards(1).NumberOfReplicas(0))
		.Mappings(ms => ms.Map<Foo>(m => m.AutoMap()))
	);
}

public class FooBase
{
	public int InheritedProp { get; set; }
}

public class Foo : FooBase
{
	public int DeclaredProp { get; set; }
}
```

I also unsuccessfully tried `InferMappingFor<FooBase>`.

Mpdreamz:
…against ConnectionSettings mappings to rename or ignore them in a late bound fashion

### DIFF
--- a/src/Nest/CommonAbstractions/Extensions/TypeExtensions.cs
+++ b/src/Nest/CommonAbstractions/Extensions/TypeExtensions.cs
@@ -24,6 +24,9 @@ namespace Nest
 		private static readonly ConcurrentDictionary<Type, IList<JsonProperty>> CachedTypeProperties =
 			new ConcurrentDictionary<Type, IList<JsonProperty>>();
 
+		private static readonly ConcurrentDictionary<Type, IList<PropertyInfo>> CachedTypePropertyInfos =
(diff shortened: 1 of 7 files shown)


Mpdreamz:
ported to `5.x` and `2.x`

PR #2562, opened 2017-01-30
as per elastic/elasticsearch#20305 deprecate Suggest() in favor over …

Mpdreamz:
…suggest through Search()


Mpdreamz:
ported to `5.x`

PR #2572, opened 2017-02-05
Feature/range mappings

Mpdreamz:
This adds support for the `*_range` data types introduced in Elasticsearch 5.2


src/Nest/Mapping/Types/FieldType.cs
```diff
@@ -69,5 +61,35 @@ public enum FieldType
...
+		[EnumMember(Value = "integer_range")]
+		IntegerRange,
+		[EnumMember(Value = "float_range")]
+		FloatRange,
+		[EnumMember(Value = "long_range")]
+		LongRange,
+		[EnumMember(Value = "double_range")]
+		DoubleRange,
+		[EnumMember(Value = "date_range")]
+		DateRange,
```
russcam: This also needs to handle custom field types e.g. [Rosette's `name` data type](https://www.rosette.com/elastic/)

Mpdreamz: The converter did not understand custom IProperty when they were string either. We can ship with an abstract `PluginPropertyBase` that opens it back up.

russcam: Sounds like a plan 👍

Mpdreamz: Amended PR with `PluginPropertyBase` and test

Mpdreamz: @forloop thumbs up counts as PR approval? :)

russcam: @Mpdreamz yes 😄👍

Mpdreamz:
ported to `master`

PR #2577, opened 2017-02-05
add ignore_missing on user_agent geo_ip and attachment processors as …

Mpdreamz:
…per elastic/elasticsearch#22273


Mpdreamz:
ported to `master`

PR #2578, opened 2017-02-05
add support for the kv processor

Mpdreamz:


Mpdreamz:
I was on the fence with `KVProcessor` since that would also mean having a `KV()` on `ProcessorsDescriptor` which looks a bit iffy :)

lmk if you object we can rename post merge.

Mpdreamz:
ported to `master`

russcam:
> I was on the fence with KVProcessor since that would also mean having a KV() on ProcessorsDescriptor which looks a bit iffy :)

yeah, `.KV()` is a bit funky. I don't have strong opinions about it to be honest 😄 @gmarz ?

gmarz:
I'm good with KeyValue 👍

PR #2580, opened 2017-02-05
Return Warning HTTP headers as deprecations on response

Mpdreamz:
still pending elastic/elasticsearch#22986


Mpdreamz:
ported to `master`

PR #2585, opened 2017-02-13
Fix #2561 make Missing option on terms aggregation generic

### LINKED ISSUE #2561: .Missing() for Term-aggregation of int-field
**NEST/Elasticsearch.Net version**: 5.0.1

When specifying `.Missing()` for Term-aggregation of integer-fields elastic reponds with a confusing error. 
It appears elastic 5 is stricter on types for missing-values in aggregations. `"missing":-1` works, but not `"missing":"-1"`.

**Steps to reproduce**:

```c
void Main()
{
	var client = new ElasticClient(new Uri("http://localhost:9200"));
	
	client.CreateIndex("fooindex", c => c.Settings(s => s.NumberOfShards(1).NumberOfReplicas(0))
		.Mappings(ms => ms.Map<Foo>(m => m.AutoMap()))
	);
	
	var resp = client.Search<Foo>(s => s
		.Index("fooindex")
		.Aggregations(a => a
			.Terms("a", t => t.Field(b => b.SomeInt)
				.Missing("-1") // need overload that takes int?
			)
		)
	);
	
	// writes: Field [someInt] of type [integer] does not support custom time zones
	Console.WriteLine(resp.ServerError.Error.RootCause.First().Reason); 
}

public class Foo
{
	public int Id { get; set; }
	public int SomeInt { get; set; }
}
```

gmarz:


src/Nest/Aggregations/Bucket/Terms/TermsAggregation.cs
```diff
@@ -40,7 +40,7 @@ public interface ITermsAggregation : IBucketAggregation
 		TermsAggregationCollectMode? CollectMode { get; set; }
 
 		[JsonProperty("missing")]
-		string Missing { get; set; }
+		object Missing { get; set; }
```
Mpdreamz: Why did you not opt for `TFieldType`, not crazy about the boxing here.

gmarz: Me either, but we'd have to make `AggregationContainer<TFieldType>` as well which opens up a whole can of worms, and I'm not sure we really should.

Mpdreamz:
@gmarz added a commit with a way to prevent boxing, including an integration/unit tests for the missing property sending `-1`.

src/Nest/Aggregations/AggregationContainer.cs
```diff
@@ -500,7 +500,11 @@ public class AggregationContainerDescriptor<T> : DescriptorBase<AggregationConta
 			_SetInnerAggregation(name, selector, (a, d) => a.Sum = d);
 
 		public AggregationContainerDescriptor<T> Terms(string name,
-			Func<TermsAggregationDescriptor<T>, ITermsAggregation> selector) =>
+			Func<TermsAggregationDescriptor<T, string>, ITermsAggregation<string>> selector) =>
+			_SetInnerAggregation(name, selector, (a, d) => a.Terms = d);
+
+		public AggregationContainerDescriptor<T> Terms<TMissingType>(string name,
+			Func<TermsAggregationDescriptor<T, TMissingType>, ITermsAggregation<TMissingType>> selector) =>
```
gmarz: I would rename `TMissingType` => `TFieldType`

gmarz:
LGTM 👍 Just a minor nitpick

gmarz:
Much better BTW :)

Mpdreamz:
rebased and amended @gmarz 👍

Issue #2582, opened 2017-02-07
ServerError needs to add some null checking around FillValues


niemyjski:
**Please note I have break on all exceptions on :)**

I'm running latest stable client against 5.1.2 (haven't updated it yet) Code in question: https://github.com/elastic/elasticsearch-net/blob/master/src/Elasticsearch.Net/Responses/ServerError.cs#L157

I'm trying to track down how to reproduce but I started writing a test on a live stream and I got this exception:
```
System.NullReferenceException occurred
 HResult=-2147467261
 Message=Object reference not set to an instance of an object.
 Source=Elasticsearch.Net
 StackTrace:
      at Elasticsearch.Net.RootCauseExtensions.FillValues(IRootCause rootCause, IDictionary`2 dict) in C:\Users\russ\source\elasticsearch-net-5.x\src\Elasticsearch.Net\Responses\ServerError.cs:line 159
 InnerException:
```

I just checked fiddler and it occurs right after a `HEAD` request.
```
HEAD http://127.0.0.1:9200/test-events-v1-%2A HTTP/1.1
Accept: application/json
Content-Type: application/json
Host: 127.0.0.1:9200
```
With the response

```
HTTP/1.1 404 Not Found
content-type: text/plain; charset=UTF-8
content-length: 0
```

This is the callstack when the exception is thrown

```
 	Elasticsearch.Net.dll!Elasticsearch.Net.RootCauseExtensions.FillValues(Elasticsearch.Net.IRootCause rootCause = {Elasticsearch.Net.Error}, System.Collections.Generic.IDictionary<string, object> dict = null) Line 159	C#
 	Elasticsearch.Net.dll!Elasticsearch.Net.Error.Create(System.Collections.Generic.IDictionary<string, object> dict, Elasticsearch.Net.IJsonSerializerStrategy strategy = null) Line 94	C#
 	Elasticsearch.Net.dll!Elasticsearch.Net.ServerError.Create(System.Collections.Generic.IDictionary<string, object> dict, Elasticsearch.Net.IJsonSerializerStrategy strategy) Line 52	C#
 	Elasticsearch.Net.dll!Elasticsearch.Net.SimpleJson.DeserializeObject<Elasticsearch.Net.ServerError>(string json, Elasticsearch.Net.IJsonSerializerStrategy jsonSerializerStrategy) Line 565	C#
[...]

PR #2586, opened 2017-02-14
More resilient error parsing 

Mpdreamz:
Relates to: https://github.com/elastic/elasticsearch-net/issues/2582

Could not quite reproduce but the point in the code @niemyjski pointed out could use a null reference check so added it. 

This also makes sure 

```
{ 
   "error" : "some error"
   "status" : 404
}

Is properly deserialized into a `ServerError`, we were incorrectly assuming this is always an object.


gmarz:
Ported to 2.x, 5.x, and master

Issue #1322, opened 2015-04-23
SniffingConnectionPool handling of node types


ppf2:
From discussion with @gmarz, the current implementation of SniffingConnectionPool does not take into account the ES node type.  In general, we do not recommend sending search and indexing requests to dedicated master nodes.  It will be nice for the sniffing functionality to identify nodes but filter out the master only nodes to prevent requests from being sent to the master only nodes.

Mpdreamz:
++

Grave omission from the current implementation thats been on the backlog for too long.

We should skip data nodes, in addition to this in 2.0 the sniffing part will be injectable and responsibility of `IConnectionPool` itself so its easier to override default behavior here.

Mpdreamz:
bumping this because we need a discussion with the other clients to come up with general rules here

Mpdreamz:
Closing this because we recommended using dedicated clients for different subset of calls if needed. E.g a ConnectionPool that only accepts nodes with the ingest role for ingest bulk indexes. 

Subclassing SniffingConnectionPool and overriding Reseed is already supported to bake in custom filter rules.

Mpdreamz:
@elastic/microsoft closed this but if you feel otherwise lets reopen to discuss

PR #2619, opened 2017-02-16
NodePredicate support on ConnectionSettings

Mpdreamz:
You can now set `NodePredicate(Func<Node, bool> filter)` on `ConnectionSettings`. 

This allows you to control which nodes are considered targets in the `RequestPipeline` for *normal* Elasticsearch API calls. Sniffing in the `RequestPipeline` does not take this predicate into account allowing you to seed your `SniffingConnectionPool` with master nodes and have it sniff on startup afterwhich none of the API calls will touch the master nodes.

The default predicate for any `IConnectionPool` that supports reseeding of nodes (`SniffingConnectionPool`) is to filter out `master only` nodes. 

For those that don't `SingleNode-, Static-, StickyConnectionPool et all) the default `NodePredicate` returns true for all nodes, assuming the list of nodes you pass are correct from the get go.

Our `Node` type returns a bunch of interesting information so you could use it to sniff all the warm or cold nodes in the cluster and only talk to those for instance. 

See the > [tests](https://github.com/elastic/elasticsearch-net/blob/2a4ac8b214df2f109074e983473a4f29fde1c80c/src/Tests/ClientConcepts/ConnectionPooling/Sniffing/RoleDetection.doc.cs#L118) <  for more examples. 

We also protect against a predicate filtering out all the nodes by throwing a new `UnexpectedElasticsearchClientException()` where its`DebugInformation` fully explaina what just happened.

cc @bleskes @gmoskovicz and @ppf2 (whom i too easily dismissed on #1322, while sub classing SniffingConnectionPool and overriding `Reseed` was **A** solution its far from pretty).

### DIFF
--- a/src/Elasticsearch.Net/Auditing/AuditEvent.cs
+++ b/src/Elasticsearch.Net/Auditing/AuditEvent.cs
@@ -18,6 +18,7 @@ namespace Elasticsearch.Net
 
 		MaxTimeoutReached,
 		MaxRetriesReached,
-		BadRequest
+		BadRequest,
+		NoNodesAttempted,
 	}
 }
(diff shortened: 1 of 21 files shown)


gmoskovicz:
@Mpdreamz that is amazing! Thanks for this addition!

bleskes:
Thanks Martijn

gmarz:
Merged to 5.x and ported to master

PR #2625, opened 2017-02-17
Stop tracking tests.yaml and add a default configuration

gmarz:
`tests.yaml` is constantly checked in every time it's changed locally
which is annoying when porting commits between branches (for instance,
master branch ends up testing againts `5.3.0-SNAPSHOT` rather than `latest`
and vice versa).

This file is really user-specific and should not be tracked -- the only
reason why we do currently is because there isn't a way to check in a
file and ignore changes to it (except `git update-index
--assume-unchanged`, 
but that would be a step required by everyone who pulls down the repo).

This PR proposes a fix by adding `tests.default.yaml` which is just a
default / template configuration for the containing branch.  This file
is tracked and only changes to the default settings should be made in
it.

To change the test runner behavior, a copy of this file has to be made
and named `tests.yaml`, which is no longer tracked.

The test runner first looks for the local `tests.yaml`, and if it's not found
falls back to the default configuration.


Mpdreamz:
Aww yeah! I was thinking of doing this exact same thing today 👍  👍

src/Tests/Framework/TestClient.cs
```diff
@@ -39,17 +39,20 @@ private static ITestConfiguration LoadConfiguration()
 			// If running the classic .NET solution, tests run from bin/{config} directory,
 			// but when running DNX solution, tests run from the test project root
 			var yamlConfigurationPath = directoryInfo.Name == "Tests" &&
-										directoryInfo.Parent != null &&
-										directoryInfo.Parent.Name == "src"
-				? "tests.yaml"
-				: @"../../../tests.yaml";
+			                            directoryInfo.Parent != null &&
```
Mpdreamz: spaces

src/Tests/Framework/TestClient.cs
```diff
@@ -97,28 +101,33 @@ private static IClrTypeMapping<Project> ProjectMapping(ClrTypeMappingDescriptor<
...
-			var defaultSettings = DefaultSettings(new ConnectionSettings(createPool(CreateUri(port)), CreateConnection(forceInMemory: forceInMemory), serializerFactory));
+			var defaultSettings = DefaultSettings(new ConnectionSettings(createPool(CreateUri(port)),
+				CreateConnection(forceInMemory: forceInMemory), serializerFactory));
 #pragma warning restore CS0618 // Type or member is obsolete
 			var settings = modifySettings != null ? modifySettings(defaultSettings) : defaultSettings;
 			return settings;
 		}
 
-		public static IElasticClient GetInMemoryClient(Func<ConnectionSettings, ConnectionSettings> modifySettings = null, int port = 9200) =>
+		public static IElasticClient GetInMemoryClient(Func<ConnectionSettings, ConnectionSettings> modifySettings = null,
```
Mpdreamz: Keep as one line? we should settle on a line length, general code formatting etcetera.

gmarz: Agreed -- I accidentally ctrl+k+d'ed everything and formatted the whole file.  Sorry for the noise.

gmarz: @Mpdreamz fixed

Mpdreamz: merge merge merge

gmarz:
This is now in master, 5.x and 2.x

PR #2628, opened 2017-02-20
Refactor Managed Elasticsearch implementation 

Mpdreamz:
Our integration tests are self managing, acquiring and installing elasticsearch and its plugins and making sure that the clusters we start are in the state we need them to be.

The code to get them in to that state was scattered around different classes with different responsibilities,
The code that actually started and waited for the node to be **really** started (e.g elasticsearch printing `started`) grew too much in complexity,

- Simplified startup code, we had too many waithandles and callbacks.
- Refactored tasks out of ElasticsearchNode and FileSystem to dedicated
  classes
- Introduced a dedicated task runner that breaks up the following stage
  s in multiple tasks:
  - Installation of elasticsearch (previously in ElasticsearchNode)
  - PreStart tasks (previously in NodeFileSystem)
  - Post start validation (previously in ElasticsearchNode)

- Tasks runner now also keeps track of tasks it ran in
  %APPDATA%/NEST/{version}, so that we can introduce more tasks without
  deleting this folder. Tasks now also not necessarily in charge of
  idempotency. Allthough writing idempotent tasks should still be the
  goal


Mpdreamz:
ported to `master`

PR #2632, opened 2017-02-22
AutoMap() is now aware of overridden properties in derived types

### LINKED ISSUE #2566: .Automap() unexpected behaviour with overridden properties
**NEST/Elasticsearch.Net version**: 5.0.1

NEST Automap implementation returns unexpected mappings when scanning types with overridden properties.
If I have the following class hierarchy:

```csharp
public class TestClass
{
	public Guid Property1 { get; set; }
	public Guid Property2 { get; set; }
}

public class TestOverriddenClass : TestClass
{
	public new int Property1 { get; set; }
}
```

what I get creating this descriptor:

```csharp
var descriptor = new CreateIndexDescriptor("myindex")
	.Mappings(ms => ms
		.Map<TestOverriddenClass>(m => m
			.AutoMap()
		)
	);
```

is this json:

```json
{
  "mappings": {
    "testoverriddenclass": {
      "properties": {
        "property1": {
          "type": "keyword"
        },
        "property2": {
          "type": "keyword"
        }
      }
    }
  }
}
```

I know the example classes above are not strictly POCOs, and this is quite a fringe case, but I wanted to know if you plan to support this scenario, because it's quite simple to do with .NET reflection.

gmarz:
Previously, `AutoMap()` would not take into account properties that
have been overridden in derived classes:

```csharp
public class A
{
    public int Foo { get; set; }
}

public class B : A
{
    public new string Foo { get; set; }
}
```

So in this instance, mapping class `B` would result in `Foo` being mapped
as an `int`, even though it has been overridden as a `string` using the `new`
keyword.

With this commit, `AutoMap()` is now aware of overridden properties and
in this example, will map `Foo` as a `string`.

Closes #2566

### DIFF
--- a/src/Nest/CommonAbstractions/Extensions/TypeExtensions.cs
+++ b/src/Nest/CommonAbstractions/Extensions/TypeExtensions.cs
@@ -41,8 +41,7 @@ namespace Nest
 
 		internal static object CreateGenericInstance(this Type t, Type[] closeOver, params object[] args)
 		{
-			var argKey = closeOver.Aggregate(new StringBuilder(), (sb, gt) => sb.Append("--" + gt.FullName),
-				sb => sb.ToString());
+			var argKey = closeOver.Aggregate(new StringBuilder(), (sb, gt) => sb.Append("--" + gt.FullName), sb => sb.ToString());
 			var key = t.FullName + argKey;
 			Type closedType;
 			if (!CachedGenericClosedTypes.TryGetValue(key, out closedType))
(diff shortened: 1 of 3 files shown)


src/Nest/CommonAbstractions/Extensions/TypeExtensions.cs
```diff
@@ -139,21 +143,53 @@ internal static IList<PropertyInfo> AllPropertiesCached(this Type t)
...
+		}
 
+		/// <summary>
+		/// Determines if a property is overriding an inherited property of its base class
+		/// </summary>
+		private static bool IsHidingMember(PropertyInfo propertyInfo)
+		{
+	#if DOTNETCORE
+			var baseType = propertyInfo.DeclaringType?.GetTypeInfo()?.BaseType;
+    #else
```
Mpdreamz: space before `#else`

src/Nest/CommonAbstractions/Extensions/TypeExtensions.cs
```diff
@@ -40,7 +41,8 @@ internal static object CreateGenericInstance(this Type t, Type closeOver, params
 
 		internal static object CreateGenericInstance(this Type t, Type[] closeOver, params object[] args)
 		{
-			var argKey = closeOver.Aggregate(new StringBuilder(), (sb, gt) => sb.Append("--" + gt.FullName), sb => sb.ToString());
+			var argKey = closeOver.Aggregate(new StringBuilder(), (sb, gt) => sb.Append("--" + gt.FullName),
```
gmarz: Argh, not use to Riders auto-formatting yet.  Will fix 👍

gmarz:
Merged to master and 5.x, ported to 2.x

Issue #2633, opened 2017-02-22
5.2.0 Integration Tests Failing


natebunton:
**NEST/Elasticsearch.Net version**:
5.2.0
**Elasticsearch version**:
5.2.1
**Description of the problem including expected versus actual behavior**:
The integration tests are failing with the following errors...I expect the integration tests to succeed.  When I run fiddler, I see a lot of web request succeeding and failing.

**Steps to reproduce**:
 1.  Create a branch from tag 5.2.0
 2.  In the root of the repo, run build.bat Integrate 5.2.1

**Provide `ConnectionSettings` (if relevant)**:
N/A
**Provide `DebugInformation` (if relevant)**:

Running build failed.
Error:
System.Exception: Calling packages/build/xunit.runner.console/tools/xunit.console.exe resulted in unexpected exitCode 1
   at FSI_0005.Tooling.Tooling.execProcess@76.Invoke(String message)
   at FSI_0005.Testing.Tests.testDesktopClr()
   at FSI_0005.Testing.Tests.RunIntegrationTests()
   at FSI_0005.Targets.clo@51-18.Invoke(Unit arg00@)
   at Fake.TargetHelper.runSingleTarget(TargetTemplate`1 target) in C:\code\fake\src\app\FakeLib\TargetHelper.fs:line 492

Here is the full build console output:  
[build console output.txt](https://github.com/elastic/elasticsearch-net/files/794791/build.console.output.txt)

**Describe the feature**:

gmarz:
Ah, yea the `BulkAllApiTests` are flaky.  We've hit this before locally and in [CI](http://elastic-dotnet.westeurope.cloudapp.azure.com/project.html?projectId=Nest5x) (which is failing at the moment for a different reason, which looks to be due to another flaky test).  If you run the tests again does it pass or does it fail consistently?

Rest assured, we always make sure the integration tests are all green before releasing, and did so against 5.2.1 -- so I'm fairly confident it's a bug in the test itself.

natebunton:
It is consistently failing.  Any ideas or work arounds to get the integration tests to pass?

gmarz:
Sorry, I just looked at the output again and it's also failing the x-pack tests because you don't have a valid license file:

```
FATAL called: Server has license plugin installed,  but the license was not found!
```

[...]

PR #2635, opened 2017-02-24
Tests: Multiple cluster and exclusive filtering support

gmarz:
This PR makes it possible to filter by multiple clusters when running
integration tests and also exclude specific clusters, where previously
cluster filtering was inclusive only.

Examples:

Run readonly tests only (unchanged):

```build integrate 5.2.1 readonly```

Run readonly and writable tests only:

```buld integrate 5.2.1 "readonly,writable"```

Note the quotation marks.  These are required when passing multiple
cluster names due to how bat arguments are processed (commas are treated
like spaces).

Run all tests except xpack:

```build integrate 5.2.1 -xpack```

Run all tests except xpack and intrusiveoperation:

```build integrate 5.2.1 "-xpack,-intrusiveoperation"```

NOTE: both including and excluding clusters is redundant, but we do not throw
an exception.  For instance:

```build integrate 5.2.1 "readonly,-xpack"```

is equivalent to just

```build integrate 5.2.1 readonly```

Relates #2633

### DIFF
--- a/build.bat
+++ b/build.bat
@@ -72,7 +72,7 @@ IF /I "%TARGET%"=="forever" (
 )
 IF /I "%TARGET%"=="integrate" (
 	IF NOT [%2]==[] (set ESVERSIONS="%2")
-	IF NOT [%3]==[] (set NEST_INTEGRATION_CLUSTER="%3")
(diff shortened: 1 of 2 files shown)


gmarz:
Merged to 2.x 5.x and master

PR #2636, opened 2017-02-24
Add IDocStorageBulkDescriptor interface to BulkDescriptor

natebunton:
Add reference to CSG.StatHub.DocStorageCore and mark the BulkDescriptor with the IDocStorageBulkDescriptor interface


Mpdreamz:
@natebunton out of curiosity why are you building a private version without .NET core and fixed namespaces. Is there anywhere you need to be able to hook into you currently can't?

natebunton:
@Mpdreamz this pull request was a mistake.  I intended on doing a pull request into our private forked repo.  We are working through some upgrades and have some rigid dependency management which is forcing us to temporarily go with a fixed namespace so we can incrementally upgrade our versions.  We don't have any dotnet core needs at the time and removing them was the path of least resistance at the time.  Once we complete the upgrade we will no longer need the fixed namespaces.  Sorry for the confusion.

PR #2643, opened 2017-03-01
Fix for covariant search date parsing

gmarz:
This commit fixes an issue where indexing dates as strings and pulling
them back out using covariant search would result in the format being
different due to the default DateParseHandling serialization settings.

ConcereteTypeConverter now temporarily sets DateParseHandling to None
when deserializing covariant results.

/cc @gmoskovicz

### DIFF
--- a/src/Nest/CommonAbstractions/SerializationBehavior/StatefulDeserialization/ConcreteTypeConverter.cs
+++ b/src/Nest/CommonAbstractions/SerializationBehavior/StatefulDeserialization/ConcreteTypeConverter.cs
@@ -95,6 +95,12 @@ namespace Nest
 		private static JObject CreateIntermediateJObject(JsonReader reader)
 		{
 			var original = reader.DateParseHandling;
+			// Temporarily turn off DateTime parsing since we deserialize
+			// to a dynamic object and the date could be stored as a string
+			// in the users POCO and we need to preserve its format. This is
+			// side-effect free since we read the reader to completion using
+			// JObject.Load before handing off a new one to the deserialized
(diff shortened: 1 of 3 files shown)


src/Nest/CommonAbstractions/SerializationBehavior/StatefulDeserialization/ConcreteTypeConverter.cs
```diff
@@ -94,7 +94,10 @@ internal static class ConcreteTypeConverter
 
 		private static JObject CreateIntermediateJObject(JsonReader reader)
 		{
+			var original = reader.DateParseHandling;
+			reader.DateParseHandling = DateParseHandling.None;
```
Mpdreamz: Add a comment why we feel this is not side effects.

gmoskovicz: Great fix. ++ on adding a comment.

Quick question,  where do we use the StatefulDeserialization class? Only for Covariant Search type? Just trying to understand whether this can affect other functionality.

Mpdreamz: CovariantSearch needs ConcreteTypeConverter, in order to create an isolated serializer that has access to state we call the Stateful deserialization factory on `ISerializerFactory` and we append a new instance of ConcreteTypeConverter. Search/MultiGet/MultiSearch/SearchTemplate all possibly call this (but only if needed)

Issue #2644, opened 2017-03-02
Memory leak in BulkAll


sergii-sakharov:
NEST 5.2+:

Elasticsearch: Any version:

In case data producer is faster than (e.g. indexing) consumer in BulkAll, all of the data sent is not garbage collected until the very end of the BulkAll operation

**Steps to reproduce**:
 1. Create a large lazyEnumerable
 2. Index data through BulkAll
 3. Memory usage is constatnly rising as opposed to stabilizing

PR #2646, opened 2017-03-02
Modified ForEachAsync to release items as processed

sergii-sakharov:
#2644

Memory leak is caused by WhenAll that expects all of the tasks in the
bulk operations to complete to further invoke done() delegate in
ForEachAsync. Referencing all of the tasks in WhenAll causes them to
prevent GC from collecting processed items.

### DIFF
--- a/src/Nest/CommonAbstractions/Extensions/Extensions.cs
+++ b/src/Nest/CommonAbstractions/Extensions/Extensions.cs
@@ -206,11 +206,11 @@ namespace Nest
 
 		internal static IEnumerable<T> EmptyIfNull<T>(this IEnumerable<T> xs) => xs ?? new T[0];
 
-		internal static Task ForEachAsync<TSource, TResult>(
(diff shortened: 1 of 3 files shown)


karmi:
Hi @sergii-sakharov, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/2646.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

sergii-sakharov:
Heh, will have to redo whole thing with proper email - other email exists already associated with different account. Sorry for the mess - first time experience forking/pull requesting

karmi:
Hi @sergii-sakharov, no need to close the PR, just a) use a different e-mail in your Git config, use `commit --amend --reset-author` and force push, or b) add the e-mails into your Github account.

sergii-sakharov:
Thanks - will keep that in mind for future (too late for this one)

PR #2647, opened 2017-03-02
Make it easier to work with server and client certificates in the client

Mpdreamz:
### Server certificates

Rather then making folks register their ServerCertificateValidation callback globally on the static `ServicePointManager` or subclass `HttpConnection` to set it on the request/itself
we now expose it on `ConnectionSettings` this callback is fired for each unique endpoint (node) until it returns true after which its cached for the duration of that servicepoint.

We also ship with handy baked in validations on `CertificateValidations`:

* `CertificateValidations.AllowAll` simply returns true
* `CertificateValidations.DenyAll` simply returns false

If your client application however has access to the public CA certificate locally Elasticsearch.NET/NEST ships with handy helpers that assert that the certificate that the server presented was one that came from our local CA certificate. If you use x-pack's `certgen` tool to [generate SSL certificates](https://www.elastic.co/guide/en/x-pack/current/ssl-tls.html) the generated node certificate does not include the CA in the certificate chain. This to cut back on SSL handshake size. In those case you can use `CertificateValidations.AuthorityIsRoot` and pass it your local copy of the CA public key to assert that the certificate the server presented was generated off that.

If you go for a vendor generated SSL certificate its common practice for them to include the CA and any intermediary CA's in the certificate chain in those case use `CertificateValidations.AuthorityPartOfChain` which validates that the local CA certificate is part of that chain and was used to generate the servers key.

### Client certificates

`ConnectionSettings` now also accepts `ClientCertificates` as a collection or `ClientCertificate` as a single certificate to be used as the user authentication for ALL requests.
`RequestConfiguration` accepts the same but will be the client certificate for that single request only.

The client certificate should be a certificate that has the public and private key available (`pfx` or `p12`) however x-pack `certgen` generates two separate `cer` and `key` files.

For .NET 4.5/4.6 we ship with a helper that creates a proper self contained certificate from these two files `ClientCertificate.LoadWithPrivateKey` but because we can no longer update a
certificates `PublicKey` algorithm in .NET core the setter is no longer there. Its typically recommended to generate a single `pfx` or `p12` file using `openssl` or `pvk2pfx` since those can simply be passed to `X509Certificate`'s constructor directly

### Test runner changes

`ClusterBase` can signal the port it wants to run under, since our test validation clusters need to be isolated each one runs on a different port so that we do not run into cached callbacks. 

It can also now indicate the cluster expects to be running under SSL and include its own installation tasks. 

These changes need to be hand picked and ported back to `2.x` where it makes sense.


Mpdreamz:
ported to `master`

Mpdreamz:
ported to `2.x` but the integration tests are not ported because the setup would be to laborious there without `certgen`. cc @forloop @gmarz

Issue #2634, opened 2017-02-24
BulkAll vs. Exceptions


rosrosros:
Using NEST 5.2 with this index code;
```

 var waitHandle2 = new CountdownEvent(1);

            var bulkAll = elasticClient.BulkAll(docs, b => b
                .MaxDegreeOfParallelism(8)
                .BackOffTime(TimeSpan.FromSeconds(10))
                .BackOffRetries(2)
                .Size(1000)
                .RefreshOnCompleted() 
                .Index(indexName)
            );

            bulkAll.Subscribe(new BulkAllObserver(
                onNext: (b) => { Console.Write("."); },
                onError: (e) => { throw e; },
                onCompleted: () => waitHandle2.Signal()
            ));

            waitHandle2.Wait();

            return;
```

If an exception is thrown the `onError` is hit, but the code never reaches the `return` statement 

Please advise

rosrosros:
Ended up using this solution:


var waitHandle2 = new ManualResetEvent(false);

	var docs = new[] {new Product(),new Product()};
	var bulkAll = client.BulkAll(docs, b => b
		.MaxDegreeOfParallelism(8)
		.BackOffTime(TimeSpan.FromSeconds(1))
		.BackOffRetries(1)
		.Size(1000)
		.RefreshOnCompleted()
		.Index(indexName)
	);

	//we set up an observer
	Exception ex = null;
	var bulkObserver = new BulkAllObserver(
		onError: (e) =>
		{
			ex = e;
			waitHandle2.Set();
			throw e;
		} ,
		onCompleted: () => waitHandle2.Set(),
		onNext: (b) =>  Console.Write("on next")
	);
	
	bulkAll.Subscribe(bulkObserver);
	waitHandle2.WaitOne();
	
	if (ex != null )
		throw ex;

Mpdreamz:
In the first example noone signals `waitHandle2` in RX only `onError` or `onCompleted` are called at most once.

If you include a reference to the full RX framework you'll gain handy operators like [Finally() and others](http://www.introtorx.com/Content/v1.0.10621.0/11_AdvancedErrorHandling.html) to deal with this in a more composable way.

It should not be necessary to rethrow in the onError but reviewing our current code i do think i spotted a bug where we need to throw here:

https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Document/Multiple/BulkAll/BulkAllObservable.cs#L89-L94

and rely on this guy to call OnError once: 

[...]

PR #2648, opened 2017-03-02
Modified ForEachAsync to release items as processed

sergii-sakharov:
#2644

Memory leak is caused by WhenAll that expects all of the tasks in the
bulk operations to complete to further invoke done() delegate in
ForEachAsync. Referencing all of the tasks in WhenAll causes them to
prevent GC from collecting processed items.

### DIFF
--- a/src/Nest/CommonAbstractions/Extensions/Extensions.cs
+++ b/src/Nest/CommonAbstractions/Extensions/Extensions.cs
@@ -206,11 +206,11 @@ namespace Nest
 
 		internal static IEnumerable<T> EmptyIfNull<T>(this IEnumerable<T> xs) => xs ?? new T[0];
 
-		internal static Task ForEachAsync<TSource, TResult>(
+		internal static async Task ForEachAsync<TSource, TResult>(
 			this IEnumerable<TSource> lazyList,
 			Func<TSource, long, Task<TResult>> taskSelector,
 			Action<TSource, TResult> resultProcessor,
-			Action<Task> done,
+			Action<Exception> done,
 			int maxDegreeOfParallelism,
 			SemaphoreSlim additionalRateLimitter = null
 		)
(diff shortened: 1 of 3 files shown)


Mpdreamz:
ty @sergii-sakharov this is the better solution rather then assuming `Task,WhenAll` wont OOM. 👍 

Im merging this locally because not awaiting that new async method has some implications for Exceptions.

Also adding `.Wait()` extension methods that will behave and rethrow exceptions as expected (relates #2634 )

sergii-sakharov:
I'm trying to write a test for ForEachAsync at the moment but it is internal, and as far as I can tell there is no InternalsVisibleTo settings. Is it ok to elevate it to public or should I try to test the issue via BulkAll invocation as a full-blown integration test?

Mpdreamz:
I think this warrants an `InternalsVisisbleTo` as much as I tried to avoid having to set it 👍

sergii-sakharov:
Just to confirm - to enable InternalsVisibleTo I had to sign Tests assembly - used the same key as Nest assebly (build/keys/..). I hope that does not affect any of the builds...

sergii-sakharov:
Unfortunately it turned out to be far from straightforward - Tests assembly has a dependency on nuget package Semantic Versioning (https://github.com/adamreeve/semver.net) which is not strongly named for some reason. I am not sure how straightforward would be to sign the dll after nuget restore - E.g. sn.exe is bundled with tooling but it works only with already signed/delay signed assemblies. Tool that is needed for signing unsigned one is AssemblyLinker (al.exe) which as far as I can tell might be platform specific. There is also an option of using something like https://www.nuget.org/packages/Nivot.StrongNaming/
Right now I am at a loss - will try to think of other ways tomorrow.

Additionaly - on original topic - I had an impression that the root cause of having to cut-off async-await chain propagating to the api level is IObserver/IObservable interfaces. They don't seem to be designed with async/await in mind. Which made me curious what is the reason for using them?

Mpdreamz:
One other solution would be be to link the extension into the Test project but that is really only a viable solution in the new csproj world and we're not ready to switch to that before Rider supports it.

Closing this in favor of https://github.com/elastic/elasticsearch-net/pull/2651 where I added some commits to this PR specifically around exception handling. Would love a quick review @sergii-sakharov !

PR #2651, opened 2017-03-03
Continuation of #2648 

Mpdreamz:
@sergii-sakharov rightfully pointed out in #2648 the Task.WhenAll is not the most
resource friendly solution and replaced it with bounded list we await
WhenAny is released and then continue.

That change had impact though on the way we relied on Task.WhenAll to be
blocking and being able to catch exceptions in Subscribe. This is now no
longer possible because we return early from calling an async method.
This is fine in our case because the returned Observable allows you to
track and dispose that async process.

This also introduces Wait(TimeSpan, Action onNext) extension methods for
period and have the exceptions propagated out of the subscribe as
expected. Fixes https://github.com/elastic/elasticsearch-net/issues/2634

### DIFF
--- a/src/Nest/CommonAbstractions/Extensions/Extensions.cs
+++ b/src/Nest/CommonAbstractions/Extensions/Extensions.cs
@@ -206,11 +206,11 @@ namespace Nest
 
 		internal static IEnumerable<T> EmptyIfNull<T>(this IEnumerable<T> xs) => xs ?? new T[0];
 
-		internal static Task ForEachAsync<TSource, TResult>(
+		internal static async Task ForEachAsync<TSource, TResult>(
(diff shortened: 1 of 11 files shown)


sergii-sakharov:
I think I am close to finishing my test for the released references (btw is there any way of running integration tests from VS using Resharper or NCrunch?). After that I will run this branch against data original version OOMed against.

sergii-sakharov:
I just finished running tests - all seems good

Mpdreamz:
ported to `master` backport to `2.x` ended up being a new PR since only in 5.x do we compose Reindex over `ScrollAll` and `BulkAll` so we only needed to update `BulkAll` there.

PR #2654, opened 2017-03-05
Added test for bulkall memory usage

sergii-sakharov:
Even though `MaxDegreeOfParallelism` is set to 1 there are multiple (3 or 4) items GC rooted in heap at all times for various reasons. E.g. as variables in enumerators in `PartitionHelper<>`, in response object, in tasks list in `ForEachAsync<,>()`. Seems acceptable but can cause higher than expected/necessary memory usage.


Mpdreamz:
LGTM 👍 ty for adding this @sergii-sakharov

Mpdreamz:
ported to `master`

PR #2665, opened 2017-03-11
fix 2662 missing scripted upsert on bulk update request in 2.x and up clients

Mpdreamz:


src/Nest/Document/Multiple/Bulk/BulkOperation/BulkUpdateBody.cs
```diff
@@ -18,5 +18,8 @@ internal class BulkUpdateBody<TDocument, TPartialUpdate>
 
 		[JsonProperty("script")]
 		internal IScript _Script { get; set; }
+
+		[JsonProperty("scripted_upsert")]
+		public bool? _ScriptedUpsert { get; set; }
```
Mpdreamz: should be internal

Mpdreamz: fixed in 72f058f

Mpdreamz:
ported to `5.x` and `master`

PR #2700, opened 2017-03-31
Correctly handle terms include/exclude

russcam:
for Significant Terms aggregation
- Deprecate dictionary that was incorrectly used to filter terms. Use TermsIncludeExclude type to represent the available options.

(port of commit a309d38d8e4b5165dc7a00a23e99a118c61b01c7)


Mpdreamz:
LGTM 👍

PR #2711, opened 2017-04-10
Feature/new new csproj

Mpdreamz:
This PR moves us over to the new new csproj files, visual studio 2017 and c# 7.0.

Updated the FAKE build in places to handle this switch. Refactored versioning a bit, still works the same though.

Updated the new new DocGenerator to be able to load new new csproj files as well. This was fixed by merely referencing `2.0.0` of `Microsoft.Csharp.*` and making sure `Microsoft.Build.*` was referenced too.

Sadly we can not use `Paket` for the time being we are running into an issue that the Paket integration with dotnet core does not take `targetframework` into account so theres no way to currently prevent certain packages to be references for full framework TFM's. Confirmed this is a bug/missing feature with @enricosada at #fsharpx. Otherwise Paket integrates with the dotnet CLI beautifully.

The `scripts` project sadly won't load either, dotnet core fsproj files are simply not supported by 2017 😢 

For now use Ionide/vim/emacs to edit our build scripts. 

On a positive note: this totally gets of us out of the hodgepodge of csproj/project.json we have in place currently!


Mpdreamz:
One thing left to update is the fileversion should not be the FixedVersion but the actual version.  This is from `2.x` currently 

```
Assembly: 2.0.0.0 File: 2.5.4 Product: 2.5.4 => C:\Projects\elastic\net-2\build\output\_packages\tmp\lib\netstandard1.3\Nest.dll
````

Where as on this branch `File:` also yields the Fixed Assembly version

enricosada:
> the Paket integration with dotnet core does not take targetframework into account

@Mpdreamz logged as https://github.com/fsprojects/Paket/issues/2233

Mpdreamz:
Ok this branch (and once pulled) is all setup on travis again and on appveyor.

Travis builds and tests dotnet core only on linux. Appveyor tests both desktop and dotnet core on windows.

I think this branch is in good shape and a pragmatic approach to moving us over.

Will open a new issue describing where we are blocked to move completely to dotnet cli and have a true singular cross platform build without hacks.

Mpdreamz:
Used admin rights to merge this in myself so that we do not keep this pending for too long and can start work on Elasticsearch `5.4` parity. cc @russcam

Issue #2522, opened 2016-12-29
Drop generic type constraint Query<T> to be T : class, and allow value types


dzmitry-lahoda:
As of now https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/QueryDsl/Query.cs is next
```
public static class Query<T> where T : class
```

Could you drop `class`? I.e.:
```
public static class Query<T> 
```

 So I could write for `stuct`s (i.e. value types)? Any reasons it is impossible?

Use case:
During usage of ElasticSearch as NoSQL storage for audit we do client side aggregation as elastic does not support summaries (sum,count) during usage of `top_hits` aggregation. 
Given different tactics, one of these like take first of all and take last for that day, leads to store millions entities in memory so it could be good to have these to be `struct`s.
Another issues that trying to build LINQ provider for our case is impossible as LINQ provider methods has no `class` constraint.

If `class` consideration was for inheritance and proxy generation in future, do not do it.

Follow VITA ORM or other delegation pattern coding/code generation - use interfaces.

Why not build LINQ to low level client? I like typing and query validation done by NEST, otherwise why I use NEST at all for our usage case.

russcam:
Thanks for opening this @asd-and-Rizzo. I'll need to look into whether this is possible, based on (off the top of my head):

1. the propagation of the generic parameter constraint from other client components i.e. can the constraint be dropped for all
2. breaking binary backwards compatibility

asd-and-Rizzo:
It would not break binary backward compatibility, I think. I did no tests, but if it would break than any drop of constraint in library code would break users - this will contradict all other good rules of CLR about backward compatibility. I think user will work fine with their constraint remained even if new assembly is used with no constraint.

asd-and-Rizzo:
I have done experiments here https://gitlab.com/dzmitry-lahoda/clr/blob/master/enterprise/compatiblity/generic_class_constraint_drop_user/Program.cs

It is safe to drop constraint for both callers of code and for thus who inherited.

asd-and-Rizzo:
I have made first try to drop constraint in all places and failed. I think try to go from most used/root classes and go to childs/users. It is possible to drop constraint or root/most used, while childs/users still have constraint. Go from bottom to top. What are these most important classes?

asd-and-Rizzo:
I am unable to do GroupBy -> Aggregations, https://msdn.microsoft.com/en-us/library/bb534492.aspx , as IGrouping is not class.

[...]

PR #2712, opened 2017-04-11
drop class constrain in promise for linq support

dzmitry-lahoda:
#2522 

Started with first place. If you fine checkin one by one, please let me know. Will sign agreement then. My research have shown that it is fine one by one, as all childs do drug constraint with them, so changes on parent has ZERO effect on existing code.


russcam:
@dzmitry-lahoda I'm going to close this PR whilst discussion is still taking place on #2522; the changes required to drop the class constraint are much more pervasive than the changes in this PR.

PR #2722, opened 2017-04-13
moved to dotnet-xunit instead of self hosted runner, teamcity appveyo…

Mpdreamz:
…r reporters and out files should be back now


Mpdreamz:
ported to `2.x` and `5.x`

Issue #2719, opened 2017-04-12
ReindexOnServerAsync error still returns true for IsValid


ejsmith:
**NEST/Elasticsearch.Net version**: 5.3.0

**Elasticsearch version**: 5.3.0

**Description of the problem including expected versus actual behavior**:

When calling ReindexOnServerAsync the IsValid on response still returns true when the server returned a 500 http status code.

**Steps to reproduce**:
 1. Call ReindexOnServerAsync with an invalid painless script.

Mpdreamz:
Thanks for reporting this @ejsmith! 

I opened https://github.com/elastic/elasticsearch-net/pull/2735 to fix this and explain why this mistake creeped into the client there as well.

Mpdreamz:
closed by https://github.com/elastic/elasticsearch-net/pull/2735

PR #2735, opened 2017-04-26
Fix #2725 and #2719

### LINKED ISSUE #2725: DeleteByQueryResponse.IsValid is true when request failed
**NEST/Elasticsearch.Net version**: 5.3.1

**Elasticsearch version**: 5.2.2

`DeleteByQueryResponse.IsValid` is `true` when the request fails. For example, if you create the ElasticClient with URLs to in-existent nodes, `DeleteByQueryResponse.OriginalException` will be something like "Maximum number of retries reached." or "Unable to connect to remote server". However, `DeleteByQueryResponse.IsValid` is `true`. In this situation, I expect `DeleteByQueryResponse.IsValid` to be `false`

**Steps to reproduce**:
 1. Create an ElasticClient with URLs to nonexistent nodes.
 2. Perform a delete by query request.

Expect: `DeleteByQueryResponse.IsValid = false`
Actual: `DeleteByQueryResponse.IsValid = true`

Mpdreamz:
Fixes #2725 and #2719

Do a double read for error for response that are always valid since they return valid object shapes on non 200-300 such as reindex and deletebyquery.

Relates too https://github.com/elastic/elasticsearch/issues/17548

In normal cases we shortcircuit reading `error` in the low level client on invalid http status codes. Some API's still return a response on bad status codes as well (such as Reindex, DeleteByQuery). If these API's **actually** return an `error` we  failed to pick up on these causing `.IsValid` to always be true here.

### DIFF
--- a/src/Elasticsearch.Net/Transport/Pipeline/ResponseBuilder.cs
+++ b/src/Elasticsearch.Net/Transport/Pipeline/ResponseBuilder.cs
@@ -66,7 +66,7 @@ namespace Elasticsearch.Net
 		private void SetBody(ElasticsearchResponse<TReturn> response, Stream stream)
 		{
 			byte[] bytes = null;
-			if (NeedsToEagerReadStream())
+			if (NeedsToEagerReadStream(response))
 			{
 				var inMemoryStream = this._requestData.MemoryStreamFactory.Create();
 				stream.CopyTo(inMemoryStream, BufferSize);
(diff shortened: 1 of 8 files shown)


src/Elasticsearch.Net/Transport/Pipeline/ResponseBuilder.cs
```diff
@@ -108,23 +104,37 @@ private void SetBody(ElasticsearchResponse<TReturn> response, Stream stream)
...
-				if (response.Success)
+				if (response.Success || response.AllowAllStatusCodes)
 				{
 					if (!SetSpecialTypes(stream, response, bytes))
 					{
 						if (this._requestData.CustomConverter != null) response.Body = this._requestData.CustomConverter(response, stream) as TReturn;
 						else response.Body = await this._requestData.ConnectionSettings.Serializer.DeserializeAsync<TReturn>(stream, this._cancellationToken).ConfigureAwait(false);
 					}
+					if (response.AllowAllStatusCodes)
+						await ReadServerErrorAsync(response, new MemoryStream(bytes), bytes);
```
russcam: I think the `new MemoryStream(bytes)` should be disposed here too

src/Elasticsearch.Net/Transport/Pipeline/ResponseBuilder.cs
```diff
@@ -76,29 +76,25 @@ private void SetBody(ElasticsearchResponse<TReturn> response, Stream stream)
...
-				if (response.Success)
+				if (response.Success || response.AllowAllStatusCodes)
 				{
 					if (!SetSpecialTypes(stream, response, bytes))
 					{
 						if (this._requestData.CustomConverter != null) response.Body = this._requestData.CustomConverter(response, stream) as TReturn;
 						else response.Body = this._requestData.ConnectionSettings.Serializer.Deserialize<TReturn>(stream);
 					}
+					if (response.AllowAllStatusCodes) 
+						ReadServerError(response, new MemoryStream(bytes), bytes);
```
russcam: I think the `new MemoryStream(bytes)` should be disposed here too

Mpdreamz: `MemoryStream` does not hold managed resources and will get gc'd just like the `byte[]` it wraps.

That said it will be disposed down the line like network streams would be too when they are deserialized (see ServerError.Create(Async).

russcam: good call 👍

Mpdreamz:
backported to `2.x` and forwardported to `master`

PR #2746, opened 2017-05-02
add boundary scanner and max fragment length support to unified highl…

Mpdreamz:
…igther as per elastic/elasticsearch#23431


src/Nest/Search/Search/Highlighting/Highlight.cs
```diff
@@ -115,5 +148,13 @@ public class HighlightDescriptor<T> : DescriptorBase<HighlightDescriptor<T> ,IHi
 		public HighlightDescriptor<T> BoundaryCharacters(string boundaryCharacters) => Assign(a => a.BoundaryChars = boundaryCharacters);
 
 		public HighlightDescriptor<T> BoundaryMaxSize(int boundaryMaxSize) => Assign(a => a.BoundaryMaxSize = boundaryMaxSize);
```
russcam: I realise this is not part of the changes here, but it looks like `boundary_max_size` should be `boundary_max_scan`:

https://github.com/elastic/elasticsearch/blob/4863a9cfdf2ca5a2fdb503e5dab9372a0462cd20/core/src/main/java/org/elasticsearch/search/fetch/subphase/highlight/AbstractHighlighterBuilder.java#L64-L64

I think it makes sense to update as part of this PR

Mpdreamz: https://github.com/elastic/elasticsearch/issues/1614 we're in good company of making this mistake :)

Mpdreamz: added _scan and obsoleted _size

Mpdreamz:
Good catch @russcam added support for `fragmenter` as well.

russcam:
++ I think `fragmenter` should also be added to the highlighter tests

Mpdreamz:
@russcam added test and also copied the properties to `IHighlightField` as well.

Mpdreamz:
ported to `master`

Issue #2657, opened 2017-03-07
Sniffing connection pool with attribute awareness


srgclr:
Hi,

We typically setup multiple VMs each with one instance of our .NET API server, one data node and one coordinating node. Then we replicate this on another server on a separate rack.

The existing sniffing connection pool is interesting but instead of round robin, we would like to always use a preferred node (coordinator) and if this fails switch to the next best and so on.

The failover node will be the one that shares the most attributes with the preferred one.

For example say we configure the nodes with attributes like '_server_', '_rack_id_' and '_room_'.

If the preferred node dies, it will select one on the same host, then one on the same server, then rack, then room and so on.

We will implement something like this but wanted to check if this feature would be desired by the community and if that's the case we will fork the project and do it here rather that in our project.

Any thoughts?

Thanks

Mpdreamz:
Hi @srgclr 

We have two things that you could use right now to do this:

[StickyConnectionPool](https://github.com/elastic/elasticsearch-net/pull/1831)

And:

[NodeSelector](https://github.com/elastic/elasticsearch-net/pull/2619) on ConnectionSettings. Look at [this test](https://github.com/elastic/elasticsearch-net/blob/2a4ac8b214df2f109074e983473a4f29fde1c80c/src/Tests/ClientConcepts/ConnectionPooling/Sniffing/RoleDetection.doc.cs#L173) for an example taking node attributes into account.

However the `StickyConnectionPool` does not support reseeding and `NodeSelector` and `Sniff` only runs on `IConnectionPools` that [supports reseeding](https://github.com/elastic/elasticsearch-net/blob/master/src/Elasticsearch.Net/ConnectionPool/IConnectionPool.cs#L24).

So we'd need a `StickySniffingConnectionPool`, happy to take a PR for that! We're currently at elasticon so limitted keyboard time is available. If not I can start looking at adding that in in two weeks.

Mpdreamz:
I guess the `stickyness` decider needs to be pluggable too so `CreateView()` returns node in the order you want.

srgclr:
Thanks @Mpdreamz,

We can start soonish but can't promise it will be ready in less than two weeks.

Will keep in touch in case you get to it first.

Cheers

srgclr:
Hi @Mpdreamz,

We've started working on this. Should have something to review soon.

Thanks

srgclr:
We're happy with #2789 
Thanks

Mpdreamz:
[...]

PR #2749, opened 2017-05-02
Add SniffingSortedStickyConnectionPool

CodeFlavoured:
Added a new connection pool type to cover scenario raised in issue #2657 by combining the capabilities of the StickyConnectionPool and SniffingConnectionPools, and allowing for the sorting of the Node collection to be defined externally.


Mpdreamz:
Thank you @CodeFlavoured ! 👍 Will review in the morning.

src/Elasticsearch.Net/ConnectionPool/SniffingSortedStickyConnectionPool.cs
```diff
@@ -0,0 +1,145 @@
...
+
+		/// <inheritdoc/>
+		public DateTime LastUpdate { get; protected set; }
+
+		public SniffingSortedStickyConnectionPool(IEnumerable<Uri> uris, Func<IEnumerable<Node>, IEnumerable<Node>> sortMethod, IDateTimeProvider dateTimeProvider = null)
+			: this(uris.Select(uri => new Node(uri)), sortMethod, dateTimeProvider)
+		{
+		}
+
+		public SniffingSortedStickyConnectionPool(IEnumerable<Node> nodes, Func<IEnumerable<Node>, IEnumerable<Node>> sortMethod, IDateTimeProvider dateTimeProvider = null)
```
Mpdreamz: `sortMethod` should have the same signature as `.OrderBy(Descending)()` from LINQ.

src/Elasticsearch.Net/ConnectionPool/SniffingSortedStickyConnectionPool.cs
```diff
@@ -0,0 +1,145 @@
...
+		/// Update the node list and sort it
+		/// </summary>
+		public void Reseed(IEnumerable<Node> nodes)
+		{
+			if (!nodes.HasAny()) return;
+
+			try
+			{
+				this._readerWriter.EnterWriteLock();
+				List<Node> sortedNodes = _sortMethod(nodes).ToList();
```
Mpdreamz: _sortMethod can be null? The constructor should guard against this if this should never be null. I think its better if we call `.OrderBy()` on the nodes since the current signature also allows filtering of the node which we already support through `NodePredicate()` on `ConnectionSettings()`.

src/Elasticsearch.Net/ConnectionPool/SniffingSortedStickyConnectionPool.cs
```diff
@@ -0,0 +1,145 @@
...
+			this._sortMethod = sortMethod;
+		}
+
+		protected int GlobalCursor = -1;
+
+		/// <summary>
+		/// Creates a view of all the live nodes preserving their order,
+		/// if there are no live nodes yields a different dead node to try once
+		/// </summary>
+		public IEnumerable<Node> CreateView(Action<AuditEvent, Node> audit = null)
```
Mpdreamz: Is it possible to extend this class from sniffing connection pool (and in turn static connection pool) and move the common bits of this routine to protected methods?

CodeFlavoured:
Hi @Mpdreamz, thanks for the review! I'll get on to those changes as soon as I get a chance today, should have an update ready either today or Monday.

karmi:
Hi @CodeFlavoured, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/2749.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

CodeFlavoured:
Hi @karmi, I've signed again with the email I used in the git commit, apologies for the confusion.

Both of my email addresses are added to my github profile already, but are private. Is there anything I should change to avoid this in the future?

src/Elasticsearch.Net/ConnectionPool/SniffingSortedStickyConnectionPool.cs
CodeFlavoured: How about passing in an `IComparer<Node>` on the constructor instead of the `Func<IEnumerable<Node>, IEnumerable<Node>>`  to control node ordering? That way Node set can't be filtered here, and it'd allow the desired _server_ > _rack_id_ > _room_ logic in issue #2657 to be implemented.

If I were to take this approach would a including an `IComparer<Node>` implementation which provides that functionality in this PR be acceptable? It would take a `Dictionary<string, string>` in the constructor and compare the values in that to those in `Node.Settings` in order to favour nodes with the greatest number of matching attributes.

Mpdreamz: @CodeFlavoured I'd prefer a simple `Func` that can be handed to .OrderBy to keep the footprint lower. e.g choosing `float` for `TKey` so you can give each node a `weight` to sort on.

CodeFlavoured: Ah ok, sounds good. I'll get right on it. Cheers!

src/Elasticsearch.Net/ConnectionPool/SniffingSortedStickyConnectionPool.cs
```diff
@@ -5,74 +5,31 @@
 
 namespace Elasticsearch.Net
 {
-	public class SniffingSortedStickyConnectionPool : IConnectionPool
+	public class SniffingSortedStickyConnectionPool : SniffingConnectionPool
```
Mpdreamz: 👍 I like this :) Maybe also refactor the common bits of `CreateView` to protected virtual members of Sniffing/Static connection pool?

CodeFlavoured: I've just pushed a change through doing that, and also making `StickyConnectionPool` inherit from `StaticConnectionPool`, since there didn't seem to be any reason to keep them separate any more that I could see. Hope that's ok!

Mpdreamz: Awesome :+1: ty for taking on more then initially bargained for!

Reviewing first thing in the morning

srgclr:
Hi @Mpdreamz, I appreciate that this isn't a priority but could you give us an indication of when this could be merged?
Many thanks.

Mpdreamz:
@srgclr this week I reckon, I did not want to include this in the previous 2.x and 5.x releases but will do in the next round which im hoping to put out by wednesday this week!

Mpdreamz:
@CodeFlavoured @srgclr thanks so much for this PR LGTM 👍 

Closing this in favor of https://github.com/elastic/elasticsearch-net/pull/2789 

- Cherry-picked relevant commits from this PR.
- made sorting a property of StickyConnectionPool
- Added tests and documentation
- Sorting now also applied to Nodes passed in the constructor not just when a sniffing reseed happens.

Should make it in the next release!

PR #2757, opened 2017-05-08
Simplify build

Mpdreamz:
Made the bash,dos script dumber by introducing a pretty horrible command line parser in F# :)

This also replaces `inc` target with `test` and targets can be suffixed with `-all` in which case commands will target all TFM's. the default is `build-all` 

e.g:

```
> build test-all geohashcell
```

Runs unit tests for all TFM's `dotnet-xunit` supports filtering the tests we with those that contain `geohashcell`

```
> build test geohashcell
```

Only runs against `netstandard1.3`


Mpdreamz:
ported to `master` and `2.x`

PR #2761, opened 2017-05-09
Fix/muliple scripted metric aggs

Mpdreamz:
Manual merge of # 2756 to include the inlined test

> If search requests multiple aggregations and there present scripted metric aggregation that returns Map (dictionary), some aggregations absent in Aggregations after json deserializaion
This fixes requires perfrom reader.Read() in any case before this line:
return new ScriptedMetricAggregate { _Value = scriptedMetric };
So reader state stays correct after current aggregate processing

ty @konbur !


Mpdreamz:
ported to `2.x` and `master`

PR #2768, opened 2017-05-22
Added troubleshooting section, corrected heading markdown tags

christofur:
For those new to compiling with .NET Core, a gotcha is the limitation of `global.json` having to specify an exact SDK version. See: https://docs.microsoft.com/en-us/dotnet/articles/core/tools/global-json

This PR adds a section to _Troubleshooting_, reminding anyone building from source, that they must have the exact SDK version installed. I have also cleaned up some markdown heading tags which were not displaying properly.  

(Elasticsearch CLA signed)

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -1,4 +1,4 @@
-#Contributing
+# Contributing
 
 Contributing back to `Elasticsearch.Net` and `NEST` is very much appreciated. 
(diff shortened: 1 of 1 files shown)


Mpdreamz:
:heart: doc PR's, thank you @christofur !

PR #2772, opened 2017-05-29
WIP PR for fixing Issue2767

SeanFarrow:
This PR extracts the required code in the tests/framework folder to allow anyone to run ElasticSearch nodes without relying on the test project. It currently targets .net standard 1.4, but this can be changed.
I'm anticipating adding this to the v6 milestone, due to the fact that I think certain API's will changed, notably the API's around the NodeFileSystem and how we access things through .ne standard.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "1.0.1"
+    "version": "1.0.3"
   },
   "version": "5.4.0"
 }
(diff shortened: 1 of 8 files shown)


Mpdreamz:
We have moved work for this here: https://github.com/Mpdreamz/elasticsearch-net-abstraction

And will resume this once `6.0` and `5.6` have been released. 

TY for getting the ball rolling @SeanFarrow 👍

SeanFarrow:
Hi,

Has this work re-started?
Kind regards
Sean.

PR #2776, opened 2017-06-06
Make sure we dispose CancellationTokenRegistration

Mpdreamz:
When the passed cancellation token is shared for a longer time then a single request, e.g bulkall/scrollall/reindex all spawn multiple requests over a single cancellationtoken it will leak Register() handles and memory will keep on growing because it will keep `requests` around indefinitely.

### DIFF
--- a/src/Elasticsearch.Net/Connection/HttpConnection.cs
+++ b/src/Elasticsearch.Net/Connection/HttpConnection.cs
@@ -180,7 +180,7 @@ namespace Elasticsearch.Net
 		}
 
 
-		private static void RegisterApmTaskTimeout(IAsyncResult result, WebRequest request, RequestData requestData) =>
+		private static RegisteredWaitHandle RegisterApmTaskTimeout(IAsyncResult result, WebRequest request, RequestData requestData) =>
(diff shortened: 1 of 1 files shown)


src/Elasticsearch.Net/Connection/HttpConnection.cs
```diff
@@ -194,42 +194,15 @@ private static void TimeoutCallback(object state, bool timedOut)
 			var builder = new ResponseBuilder<TReturn>(requestData, cancellationToken);
 			try
 			{
-				var request = this.CreateHttpWebRequest(requestData);
-				cancellationToken.Register(()=>request.Abort());
 				var data = requestData.PostData;
-
-				if (data != null)
+				var request = this.CreateHttpWebRequest(requestData);
+				using (cancellationToken.Register(() => request.Abort()))
```
russcam: Does this also need to include `await GetResponseAsync(requestData, request, builder);` in the using block?

Mpdreamz: Your right i think it does

Mpdreamz:
@forloop updated with one more commit mind giving this another review? 👍

Mpdreamz:
@russcam spotted an opportunity to clean up the code while manually merging to `2.x` forward ported these to `master` and `5x` too (see attached commits).

russcam:
Nice!

Issue #2777, opened 2017-06-07
SniffingConnectionPool SynchronizationLockException


sss13579:
It's my fault that i accidentally disposed SniffingConnectionPool, but i've got SynchronizationLockException with message "The read lock is being released without being held".

at System.Threading.ReaderWriterLockSlim.ExitReadLock()
   at Elasticsearch.Net.SniffingConnectionPool.CreateView(Action`2 audit) in C:\Projects\elastic\net-5\src\Elasticsearch.Net\ConnectionPool\SniffingConnectionPool.cs:line 84

Please move EnterReadLock & EnterWriteLock outside 'try' block. I prefer to see ObjectDisposedException rahter than SynchronizationLockException. Examples in msdn have Enter...Lock outside 'try' block (https://msdn.microsoft.com/en-us/library/system.threading.readerwriterlockslim.enterreadlock(v=vs.110).aspx).

Mpdreamz:
Thanks for the heads up, added a PR to address this @sss13579

PR #2780, opened 2017-06-08
move monitor enter readlock outside of try

Mpdreamz:
as per #2777


Mpdreamz:
backported to `2.x` and `master`

PR #2781, opened 2017-06-12
Throw an exception when Reindex() produces just enough documents for …

### LINKED ISSUE #2771: Reindex hangs when calling Subscribe
**NEST/Elasticsearch.Net version**: NEST 5.4

**Elasticsearch version**: 5.0.2

I'm trying to use the Reindex to take all products from one index, enrich them, and reindex them in a separate index.

The code I'm using is the following:

    var reindexObservable = elasticClient.Reindex<ProductDocument, ProductDocument>(pd =>
    {
        pd.DisplayName = "Foobar";
        return pd;
    },
    descriptor => descriptor
                    .ScrollAll(new Time(TimeSpan.FromMinutes(1)), 4,
                        scrollAllDescriptor => scrollAllDescriptor
                        .MaxDegreeOfParallelism(2)
                            .Search(searchDescriptor => searchDescriptor
                                .Index("currentcatalog")
                                .Type<ProductDocument>()
                                .MatchAll()
                                .Size(250)))
                    .BulkAll(bulkAllDescriptor => bulkAllDescriptor
                        .MaxDegreeOfParallelism(1)
                        .RefreshOnCompleted()
                        .Index("updatedcatalog")
                        .Type<ProductDocument>())
                    .OmitIndexCreation()
    );
    
    var obs = new ReindexObserver(
        onNext: br => Console.WriteLine($"Next: {br.Page}"),
        onError: ex =>
        {
            Console.WriteLine($"Error: {ex.Message}");
            throw ex;
        },
        onCompleted: () =>
        {
            Console.WriteLine("Completed");
        });
    
    reindexObservable.Subscribe(obs);

When the `reindexObservable.Subscribe(obs);` is called, the program never seems to move any further, and none of the onNext/Error/Complete are called.

Am I misconfiguring the Scroll/Bulk or is this a bug?

The index contains around 120k documents.

Mpdreamz:
…a single bulk but leaves no slots to call MoveNext() which would free up an awaiting task. fixes #2771

### DIFF
--- a/src/Nest/Document/Multiple/Reindex/ReindexObservable.cs
+++ b/src/Nest/Document/Multiple/Reindex/ReindexObservable.cs
@@ -131,6 +131,13 @@ namespace Nest
 					+ $" which is smaller then the bulkSize:{bulkSize}."
 				);
 
+			var funnelExact = producerBandwidth == bulkSize;
+			if (funnelExact)
+				throw new Exception("The back pressure settings are too conservative it provides enough documents for a single bulk but not enough room to advance "
+					+ $"searchSize:{searchSize} * maxConcurrency:{maxConcurrency} * backPressureFactor:{backPressureFactor} = {producerBandwidth}"
+					+ $" which is exactly the bulkSize:{bulkSize}. Increase the BulkAll max concurrency or the backPressureFactor"
+				);
+
 			var backPressure = new ProducerConsumerBackPressure(backPressureFactor, maxConcurrency);
 			return backPressure;
 		}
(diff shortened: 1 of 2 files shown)


src/Nest/Document/Multiple/Reindex/ReindexObservable.cs
```diff
@@ -131,6 +131,13 @@ private ProducerConsumerBackPressure CreateBackPressure(IBulkAllRequest<IHitMeta
 					+ $" which is smaller then the bulkSize:{bulkSize}."
 				);
 
+			var funnelExact = producerBandwidth == bulkSize;
+			if (funnelExact)
+				throw new Exception("The back pressure settings are too conservative it provides enough documents for a single bulk but not enough room to advance "
```
russcam: minor pedantic punctuation nitpick:

```
The back pressure settings are too conservative. They provide enough documents for a single bulk but not enough room to advance ...
```

Mpdreamz: 👍 added

Mpdreamz:
Only needed to be backported to `master`

Issue #2763, opened 2017-05-12
Null as parameter value in inline script for UpdateByQuery


robertkala:
**NEST/Elasticsearch.Net version**: 2.5.1

**Elasticsearch version**: 2.4.3

**Description of the problem including expected versus actual behavior**:

`elasticClient.UpdateByQueryAsync<Type>("IndexName", Type, x => x
                .Query(q => q.Term(t => t.Field(f => f.Field).Value(Value)))
                .Script(s => s.Inline("ctx._source.field2 = field2;")
                            .Params(p => p.Add("field2", null))));`

In this code parameter is skipped in script parameters because of `null` value. Most probably problem is in json serialization. Parameters are empty. In case when in paremeter `field2` there's not null value everything works fine. 

Is there a way to pass null value in parameter for inline script?

Mpdreamz:
Not currently but i've opened: https://github.com/elastic/elasticsearch-net/pull/2783

Which would allow you to do so in the next version of `2.x` and `5.x`. 

Thanks for reporting this @robertkala

PR #2783, opened 2017-06-12
allow null to be passed as value on params to scripts

Mpdreamz:
fix https://github.com/elastic/elasticsearch-net/issues/2763


Mpdreamz:
ported to `master` and `2.x`

Issue #2408, opened 2016-11-22
Top hits aggregation - support multitype type hits


llRandom:
**Describe the feature**:
Current version doesn't allow to get deserialized documents from `TopHitsAggregate` aggregation object if they have different types. An attempt to use an `object` as a generic type leads to an exception. The only workaround I've found is to use `Newtonsoft.Json.Linq.JObject` type for getting hits and then cast hits using `ToObject(targetType)` method:

    var doss = topHitsAgg.Hits<Newtonsoft.Json.Linq.JObject>()
        .Select(hit =>
        {
            var type = GetType(hit.Type);
            return hit.Source.ToObject(type);
        });

Mpdreamz:
We have a special `ILazyDocument` type that you can use to defer the deserialization of your document type.

Then you can call `.As<MyObject>()` on these to materialize them.

llRandom:
Thank you, that might work as a workaround instead of `Newtonsoft.Json.Linq.JObject` type but this interface doesn't allow to specify cast type at runtime (I can't pass required deserialization type as an argument). Besides, it looks odd to me that approaches for root hits and hits from top aggregation are so different.

Mpdreamz:
@llRandom while I agree this would be a super handy feature, in NEST 5.x it would be incredibly hard to introduce without backwards compatible breaking changes. Closing this as `Elasticsearch 6.x` no longer supports multiple types this should be ideally handled by the serializer itself. 

In NEST 6.x and up it will be possible to configure a serializer **just** for `_source`'s and field values so the responsibility will shift.

PR #2791, opened 2017-06-27
Use serializer when deserializing LazyDocument to T

### LINKED ISSUE #2788: Unable to deserialize `number` field into `TimeSpan` for inner hits
**NEST/Elasticsearch.Net version**: 5.4.0

**Elasticsearch version**: 5.4.0

**Description of the problem including expected versus actual behavior**: Unable to deserialize `number` field into `TimeSpan` for inner hits. An attempt to deserialize inner hits into object leads to an exception: `Newtonsoft.Json.JsonSerializationException: Error converting value 37230000000 to type 'System.TimeSpan'. Path 'startTime'.`

**Steps to reproduce**:
```c#
// sample mapping with nested objects with TimeSpan field
class Root
{
    [Nested]
    public Child[] Children { get; set; }
}

class Child
{
    public TimeSpan StartTime { get; set; }

    public TimeSpan EndTime { get; set; }
}

var indexName = "sample";
var client = new ElasticClient(new ConnectionSettings().DefaultIndex(indexName));

//create index with automapping
client.CreateIndex(indexName, create => create.Mappings(mappings => mappings.Map<Root>(map => map.AutoMap())));

// add sample document
client.Index(new Root
{
    Children = new[]
    {
        new Child
        {
            StartTime = new TimeSpan(1, 2, 3),
            EndTime = new TimeSpan(2, 3, 4)
        }
    }
}, index => index.Refresh(Refresh.WaitFor));

// search with inner hits
var result = client.Search<Root>(
    search => search.Query(
        query => query.Nested(
            nested => nested.Query(
                nestedQuery => nestedQuery.MatchAll()
            ).Path(i => i.Children).InnerHits()
        )
    )
);

// try to deserialize (exception)
var children = result.Hits.First().InnerHits.Single().Value.Documents<Child>();
```

russcam:
When converting a LazyDocument to type T, use the JsonNetSerializer configured, along with all registered contract resolvers.

Fixes #2788

### DIFF
--- a/src/Nest/CommonAbstractions/LazyDocument/LazyDocumentJsonConverter.cs
+++ b/src/Nest/CommonAbstractions/LazyDocument/LazyDocumentJsonConverter.cs
@@ -8,7 +8,7 @@ namespace Nest
 	{
 		public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
 		{
-			var d = value as LazyDocument;
+			var d = (LazyDocument)value;
 			if (d?._Value == null)
 			{
 				writer.WriteNull();
--- a/src/Nest/CommonAbstractions/LazyDocument/LazyDocumentJsonConverter.cs
+++ b/src/Nest/CommonAbstractions/LazyDocument/LazyDocumentJsonConverter.cs
@@ -10,19 +10,24 @@ namespace Nest
 		{
 			var d = value as LazyDocument;
 			if (d?._Value == null)
+			{
+				writer.WriteNull();
 				return;
+			}
+
 			writer.WriteToken(d._Value.CreateReader());
 		}
 
 		public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
 		{
 			var document = serializer.Deserialize(reader) as JToken;
-			return new LazyDocument { _Value = document };
+			return new LazyDocument
+			{
+				_Value = document,
+				_Serializer = serializer
+			};
 		}
 
-		public override bool CanConvert(Type objectType)
-		{
-			return true;
-		}
+		public override bool CanConvert(Type objectType) => true;
 	}
 }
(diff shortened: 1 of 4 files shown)


src/Nest/CommonAbstractions/LazyDocument/LazyDocumentJsonConverter.cs
```diff
@@ -10,19 +10,24 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
 		{
 			var d = value as LazyDocument;
 			if (d?._Value == null)
+			{
+				writer.WriteNull();
```
Mpdreamz: Why was this change needed?

Mpdreamz:
Can we also add an ovlerload to fix https://github.com/elastic/elasticsearch-net/issues/2408 as part of this PR. 

Two for one :)

src/Nest/CommonAbstractions/LazyDocument/LazyDocumentJsonConverter.cs
russcam: I think this is an oversight in the implementation; if `d` is `null` or `d.Value` is `null` and a value is not written to the writer for it, invalid json will be emitted

russcam:
What overload were you thinking of for #2408, @Mpdreamz?

Mpdreamz:
https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/CommonAbstractions/LazyDocument/LazyDocument.cs#L21

Also a `As(Type T)` for dynamic invocations.

russcam:
Gotcha

russcam:
ported to `2.x` in c77ee31e120b49a774cf5a463c1bb0cecf2cfb58
ported to `master` in bd71ff8f27585dc1477ed9bed6f47673190b2ac5

PR #2793, opened 2017-06-27
Runnable xunit tests with Resharper

russcam:
Visual Studio 2017 RTM and Resharper unit test runner are currently broken: https://youtrack.jetbrains.com/issue/RSRP-464233

This fix allows xunit tests to be runnable within Visual Studio by adding references to the following packages:

```
<PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
<PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
```

A consequence of doing this is that Microsoft.NET.Test.Sdk defines a Main entry point which means that Tests project cannot define one. Main renamed to TestMain until this is resolved.

### DIFF
--- a/src/Tests/Program.cs
+++ b/src/Tests/Program.cs
@@ -56,7 +56,7 @@ namespace Tests
 		private static  string SdkPath { get; }
 		private static  string OutputPath { get; }
 
-		public static void Main(string[] args)
+		public static void TestMain(string[] args)
 		{
 			if (args.Length == 0)
 				Console.WriteLine("Must specify at least one argument: TestAssemblyPath, Profile or Benchmark ");
--- a/src/Tests/Program.cs
+++ b/src/Tests/Program.cs
@@ -56,6 +56,9 @@ namespace Tests
 		private static  string SdkPath { get; }
 		private static  string OutputPath { get; }
 
+		// TODO: Renamed this from Main to TestMain because of a bug with Resharper and running unit tests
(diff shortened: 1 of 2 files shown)


src/Tests/Program.cs
```diff
@@ -56,7 +56,7 @@ static Program()
 		private static  string SdkPath { get; }
 		private static  string OutputPath { get; }
 
-		public static void Main(string[] args)
+		public static void TestMain(string[] args)
```
Mpdreamz: Add comment why we do this

src/Tests/Tests.csproj
```diff
@@ -23,14 +23,17 @@
     <PackageReference Include="BenchMarkDotNet" Version="0.10.0" />
     <PackageReference Include="DiffPlex" Version="1.4.1" />
     <PackageReference Include="System.Reactive" Version="3.1.1" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
     <!-- TODO only for Desktop CLR? -->
```
Mpdreamz: :Sad: :cry:

src/Tests/Program.cs
russcam: ++

russcam:
- ported to `master` in 3be2f67dc875dc56bcd78178c7c1b442cdae35e3
- ported to `2.x` in 99dde04b9e113afd27b0abd6c958cdd38aeb6dbc

PR #2807, opened 2017-07-14
5.5

Mpdreamz:
This updates the json spec to 5.5 and generates the code from that.  No big major changes to existing API's.

The machine learning json spec is included but we do not generate code from that folder just yet, this will follow in a `5.5.x` release.

`XPack security` has a new role_mapping API to map roles to users, and a token based access API. Support for these have been added with tests.

The `matrix_stats`, `sig_terms` and `sampler_agg` have additional properties as per 5.5 which tripped up our aggregation parser. 

GetAlias now return a 404 if you submit a query for e.g `a,b,c` and only alias `a` was found as per: https://github.com/elastic/elasticsearch/pull/25043


src/Elasticsearch.Net/Domain/RequestParameters/RequestParameters.Generated.cs
```diff
@@ -7001,6 +7001,40 @@ public class DeleteRoleRequestParameters : FluentRequestParameters<DeleteRoleReq
 		
 	}
 	
+	///<summary>Request parameters descriptor for XpackSecurityDeleteRoleMapping
+	///<pre>
+	///https://www.elastic.co/guide/en/x-pack/master/security-api-role-mapping.html#security-api-delete-role-mapping
+	///</pre>
+	///</summary>
+	public class DeleteRoleMappingRequestParameters : FluentRequestParameters<DeleteRoleMappingRequestParameters>
```
russcam: I think this should adhere to allowing a 404 status code response, similar to other Delete operations

src/Elasticsearch.Net/Domain/RequestParameters/RequestParameters.Generated.cs
```diff
@@ -7133,6 +7167,66 @@ public class GetRoleRequestParameters : FluentRequestParameters<GetRoleRequestPa
 		
 	}
 	
+	///<summary>Request parameters descriptor for XpackSecurityGetRoleMapping
+	///<pre>
+	///https://www.elastic.co/guide/en/x-pack/master/security-api-role-mapping.html#security-api-get-role-mapping
+	///</pre>
+	///</summary>
+	public class GetRoleMappingRequestParameters : FluentRequestParameters<GetRoleMappingRequestParameters>
```
russcam: I think this should also be consistent with other Get* responses in Elasticsearch in returning a `{ "found": false }` body along with 404 response. Speaking with @tvernum, this behaviour is consistent across Get* responses in X-Pack Security, but going to open up an issue to collectively discuss these.

src/Tests/XPack/Security/RoleMapping/PutRoleMapping/PutRoleMappingApiTests.cs
```diff
@@ -0,0 +1,109 @@
...
+			Roles = new [] { "admin"},
+			Metadata = new Dictionary<string, object>
+			{
+				{"x", "y"},
+				{"z", null}
+			},
+			Rules =
+				(new DistinguishedNameRule(_dn) | new UsernameRule(_username) | new RealmRule(_realm))
+				& new MetadataRule(_metadata.Item1, _metadata.Item2)
+				& !new GroupsRule(_groups)
```
russcam: does it make sense to use logical operators here as opposed to bitwise, similarly to how they are used within queries? For example,

```c#
(new DistinguishedNameRule(_dn) || new UsernameRule(_username) || new RealmRule(_realm))
&& new MetadataRule(_metadata.Item1, _metadata.Item2)
&& !new GroupsRule(_groups)
```

src/Tests/XPack/Security/RoleMapping/RoleMappingRules.doc.cs
```diff
@@ -0,0 +1,164 @@
...
+namespace Tests.Search
+{
+	/**[[role-mapping-rules]]
+	 * === Role Mapping Rules
+	 *
+	 * X-Pack Security allows you to map role rules through the API as of 5.5
+	 * NEST exposes this role rules DSL in a strongy typed fashion
+	 *
+	 */
+	public class RoleMappingRulesTests : SerializationTestBase
```
russcam: X-Pack docs 😃

src/CodeGeneration/ApiGenerator/RestSpecification/Core/indices.refresh.json
```diff
@@ -25,14 +25,6 @@
             "options" : ["open","closed","none","all"],
             "default" : "open",
             "description" : "Whether to expand wildcard expression to concrete indices that are open, closed or both."
-        },
```
codebrain: Fixed: https://github.com/elastic/elasticsearch-net/pull/2807/commits/7c9b72acb09ae97c1224435c027b35ef9ccfbfcd

src/Tests/XPack/Security/RoleMapping/PutRoleMapping/PutRoleMappingApiTests.cs
Mpdreamz: That sadly won't be possible since we need to return a different class when they are combined.

src/Elasticsearch.Net/Domain/RequestParameters/RequestParameters.Generated.cs
Mpdreamz: Good catch!

src/CodeGeneration/ApiGenerator/Overrides/Descriptors/RefreshDescriptorOverrides.cs
```diff
@@ -0,0 +1,23 @@
+using ApiGenerator.Domain;
+
+namespace ApiGenerator.Overrides.Descriptors
```
Mpdreamz: This can also be achieved by adding an `*.obsolete.json` companion file to the rest spec itself see:

https://github.com/elastic/elasticsearch-net/blob/100cf7c3f81553456fb039ca7a499a97c7990028/src/CodeGeneration/ApiGenerator/RestSpecification/Core/cat.thread_pool.obsolete.json

russcam: Can we move the *.obsolete.json and *.patch.json files to a different directory to the downloaded specs? I think it'll make it easier to manage

src/Tests/XPack/Security/RoleMapping/PutRoleMapping/PutRoleMappingApiTests.cs
russcam: Could we have implementations of `AnyFieldRule`, `AllFieldRule` and `ExceptFieldRule` that derive from `FieldRuleBase` and implicitly convert to the associated `RoleMappingRuleBase` implementations, respectively? Would that work?

I think there's potential for confusion with using bitwise operators, since we use binary operators for combining queries, aggregations and watcher actions.

Mpdreamz: Implicit conversions and operator overloading are not friends :)

I hear you on the deviation from other areas though but as this is icing and not the only way to construct this (ois/fluent are also there). I am leaning more on the rtfm side of things here tbh. 

We could call it out in the docs stronger though.

russcam: I think people might eat more of the icing than cake 🤣 

`FieldRuleCombinator` _a la_ `ActionCombinator`? https://github.com/elastic/elasticsearch-net/blob/1de5538efda53dac24b024ce18c1b0c5012973ce/src/Nest/XPack/Watcher/Action/ActionBase.cs#L51-L73.

Mpdreamz: `ActionBase` overloaded operators return `ActionBase` so binary operators work. `FieldRuleBase` needs to return `RoleMappingRuleBase` so it can't.

PR #2806, opened 2017-07-14
Feature/5.x build output

russcam:
This PR is a return to outputting the build for all projects from the command line to `build/output`. All projects build there with the following dir structure `build/output/<Project>/<TFM>`

- package nuget packages from `build/output`
- remove superfluous properties from `Paths` module
- specify Json.NET version dependency in one place and parameterize in nuspec file.

PR #2813, opened 2017-07-25
Benchmarking implementation

codebrain:
_First PR!_

Benchmarking

- Sets up the basic benchmark plumbing in the FAKE scripts;
    - Optional command line switch for 'non-interactive' that allows you to run all discovered benchmarks.
    - Can index into an ES instance (command line arguments for url/user/pass)
- "Clean" target now always performs a clean.
- Example benchmarks now include PostData benchmarks.
- Using a custom JSON exporter, until this issue is resolved: https://github.com/dotnet/BenchmarkDotNet/issues/505


build/scripts/Benchmarking.fsx
```diff
@@ -10,26 +16,131 @@ open System.Linq
...
 module Benchmarker =
-   let private testsProjectDirectory = Path.GetFullPath(Paths.Source("Tests"))
-   let private benchmarkOutput = Path.GetFullPath(Paths.Output("benchmarks")) |> directoryInfo
 
-   let private copyToOutput file = CopyFile benchmarkOutput.FullName file
+    let pipelineName = "benchmark-pipeline"
+    let indexName = IndexName.op_Implicit("benchmark-reports")
+    let typeName = TypeName.op_Implicit("benchmarkreport")
+    
+    type Memory(gen0Collections:int, gen1Collections: int, gen2Collections: int, totalOperations:int64, bytesAllocatedPerOperation:int64) =
```
Mpdreamz: Do we have plans to create a custom Elasticssearch BenchMarkDotNet exporter? 

https://github.com/dotnet/BenchmarkDotNet/blob/master/docs/guide/Configs/Exporters.md

That would be such a cool nuget package

Mpdreamz: While this works and I know we have this types because the json type provider garbled certain types I rather in the long run we don't maintain these POCO's as part of our build script.

build/scripts/Building.fsx
```diff
@@ -70,12 +72,8 @@ module Build =
         compileCore incremental
 
     let Clean() =
-        match (quickBuild, getBuildParam "target" = "clean") with
```
Mpdreamz: Why was this removed? this section helped when running tests from the command line to rely on incremental builds. Incremental builds are not that fast on .NET core 1.1 but they still help a great deal.

build/scripts/Benchmarking.fsx
```diff
@@ -10,26 +16,131 @@ open System.Linq
...
-
-            DotNetCli.RunCommand(fun p ->
-                { p with
-                    WorkingDir = testsProjectDirectory
-                }) "run Benchmark"
+            if runInteractive then
+                DotNetCli.RunCommand(fun p ->
+                    { p with
+                        WorkingDir = testsProjectDirectory
+                    }) "run -f net46 -c Release Benchmark"
```
Mpdreamz: Should we hardcode `net46` here? The fake command line parser snoops `-all` off `target` names so we can e.g do `built test` and `build test-all` or `build integrate-all` which would tests all the TFM's we support.

Should we do something similar for `benchmark` ?

https://github.com/elastic/elasticsearch-net/blob/4fa7daef451f629286c4986672a6b1740a00eb55/build/scripts/Testing.fsx#L29

Mpdreamz:
Appveyor is not happy because it can not find the DocGenerator

https://ci.appveyor.com/project/Mpdreamz/elasticsearch-net/build/1.0.245

Output path changes not completely in sync with: https://github.com/elastic/elasticsearch-net/pull/2806

build/scripts/Benchmarking.fsx
codebrain: I completely agree, it's really not ideal - I will switch to just using the raw JSON output at some point.

codebrain: In terms of the exporter, it might be a nicer way for us to get the data into ES than using the F# script :)

russcam: ++ I think this is good to merge now for _progression over perfection_  and build an exporter afterwards

build/scripts/Benchmarking.fsx
russcam: Discussed offline. The net46 framework is only the framework under which to run Tests.exe and the benchmark switcher. The framework used for benchmarking tests is chose through the benchmark configuration for the chosen benchmark.

PR #2820, opened 2017-07-31
Replaced pre_offset and post_offset with offset

jeroenkoknl:
`pre_offset` and `post_offset` are not supported in the search API, `offset` is.


jeroenkoknl:
Could someone please review this PR?

russcam:
Thanks for the PR @jeroenkoknl, I'll take a look

src/Nest/Aggregations/Bucket/Histogram/HistogramAggregation.cs
```diff
@@ -26,11 +26,8 @@ public interface IHistogramAggregation : IBucketAggregation
 		[JsonProperty("extended_bounds")]
 		ExtendedBounds<double> ExtendedBounds { get; set; }
 
-		[JsonProperty("pre_offset")]
-		long? PreOffset { get; set; }
-
-		[JsonProperty("post_offset")]
-		long? PostOffset { get; set; }
```
russcam: For binary compatibility reasons, these can't be removed in a major version i.e. they can be removed in NEST 6.x but not 5.x. These should be marked with `[Obsolete("removed in Elasticsearch 2.0. Will be removed in the next major version of NEST")]` or something to that effect, and the `JsonPropertyAttribute`s removed, so that setting them becomes a no-op.

src/Nest/Aggregations/Bucket/Histogram/HistogramAggregation.cs
```diff
@@ -26,11 +26,8 @@ public interface IHistogramAggregation : IBucketAggregation
...
+		[JsonProperty("offset")]
+		double? Offset { get; set; }
```
russcam: [It looks like this can be a `Time` type](https://github.com/elastic/elasticsearch/blob/9f3a34efa34b500147f3235a31043ceab685d410/core/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/DateHistogramAggregationBuilder.java#L212-L212) for greater flexibility.

jeroenkoknl: I may be misunderstanding your request but I'm not sure this is correct. This change applies to the `HistorygramAggregation` and not to `DateHistogramAggregation`. Also in the [java counterpart `HistogramAggregationBuilder`](https://github.com/elastic/elasticsearch/blob/9f3a34efa34b500147f3235a31043ceab685d410/core/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/HistogramAggregationBuilder.java#L149) doesn't have this method. I may be missing something so could you please explain?

jeroenkoknl:
@russcam, any ideas why CI failed? The error message:
>No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Mpdreamz:
@jeroenkoknl seems like an intermittent travis bootstrap error. Again many thanks for the PR, much appreciated 👍 , rebased and squashed this commit without the merge commits via https://github.com/elastic/elasticsearch-net/pull/2845 and forward ported it to `master`.

jeroenkoknl:
Great! Thank you!

PR #2823, opened 2017-08-08
Fixes to Stack Overflow links in NEST

NickCraver:
This updates "stackoverflow" to "Stack Overflow" and moves to `https://`.


Mpdreamz:
We should have known better, `Elasticsearch` misspellings are a pet peeve of mine too 🙂

PR #2826, opened 2017-08-09
Add AggregationsHelper for SerialDifferencing

russcam:
Use lag in the usage test that return values.

https://discuss.elastic.co/t/using-serial-differencing-aggregation-in-net/96402/3


Mpdreamz:
ported to `master`

PR #2839, opened 2017-08-25
Fix the property names

tishion:
Correct the mismatch of the property name


Mpdreamz:
Thank you for this PR @tishion the situation is bit more complicated in `5.x` and `2.x` its `_status` but in master its `status`. Its always `metadata` though. Closing this in favor of fixing the branches manually but will be sure to attribute you in the release notes :smile:

tishion:
That's sad, because we need to use watch API now. Anyway we will find our own solution before there is a release.

PR #2845, opened 2017-09-05
deprecate PreOffset and PostOffset and introduced Offset

Mpdreamz:
Reintroduced removed members for backwards compatibility

rebased and squashed version of #2820 by @jeroenkoknl


Mpdreamz:
ported to master

Issue #2831, opened 2017-08-15
NU1603 Warnings


jwu-au:
**NEST/Elasticsearch.Net version**:
5.4/5.5

**Description of the problem including expected versus actual behavior**:
```
warning NU1603: NEST 5.4.0 depends on Newtonsoft.Json (>= 10.0.0 && < 11.0.0) but Newtonsoft.Json 10.0.0 was not found. An approximate best match of Newtonsoft.Json 10.0.1 was resolved.
```
Shouldn't have these warnings when build in latest vs2017

**Steps to reproduce**:
 1. Updated VS2017 to Version 15.3.0
 2. Build the project having reference on NEST 5.4/5.5

benhilty:
Same problem with versions 2.4.6/2.4.6.

> error NU1603: NEST 2.4.6 depends on Newtonsoft.Json (>= 9.0.0 && < 10.0.0) but Newtonsoft.Json 9.0.0 was not found. An approximate best match of Newtonsoft.Json 9.0.1 was resolved.

anurse:
I also hit this. Specifically when trying to use NEST in a .NET Core project.

This is caused by NEST's `nuspec` file setting a lower-bound on Newtonsoft.Json of `10.0.0` but that version does not exist, the best match would be `10.0.1`. Releasing new packages with updated lower-bounds to that would fix this issue. The reason these warnings are reported is that it negatively impacts package restore performance because the NuGet has to check the package feed for version `10.0.0` each time a restore is done (though there is some caching) before it falls back to `10.0.1`.

See https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings#nu1603 for more.

I'd have sent a PR but I haven't had a chance to dig into the build scripts enough or set things up locally to test. I believe this line is the culprit:

https://github.com/elastic/elasticsearch-net/blob/2efdb55a6d40e71ca60b42ed9a23ddc7e6125863/build/scripts/Releasing.fsx#L37

kwilliamssbs:
I've taken @anurse 's suggestion and raised a PR - it's not a blocker for us, but it does mean that all dependencies that reference Newtonsoft.Json have warning triangles in VS, which makes it hard to find the actual warnings.

xplatsolutions:
Any update when the next release fixing this will be available?

russcam:
@xplatsolutions Hoping to get a new 5.x release out within the next week. We've been focused on getting the 6.x version of the client shipped as a priority.

PR #2862, opened 2017-09-20
Update Releasing.fsx

kwilliamssbs:
Removing the computed properties for the current/next version of Newtonsoft.Json in favour of supplying the strings raw, to allow us to use "10.0.1" as the correct minimum version. I've removed the computation and simply specified the versions as strings.

This resolves issue #2831 by specifying a valid value for the minimum version of Newtonsoft.Json, and thus fixes any NU1603 compiler warnings.

### DIFF
--- a/build/scripts/Releasing.fsx
+++ b/build/scripts/Releasing.fsx
@@ -34,10 +34,8 @@ module Release =
 
             let year = sprintf "%i" DateTime.UtcNow.Year
 
-            let jsonDotNetVersion = 10
-
-            let jsonDotNetCurrentVersion = sprintf "%i" jsonDotNetVersion
-            let jsonDotNetNextVersion = sprintf "%i" (jsonDotNetVersion + 1)
+            let jsonDotNetCurrentVersion = "10.0.1"
+            let jsonDotNetNextVersion = "11.0"
 
             let properties =
                 let addKeyValue (e:Expr<string>) (builder:StringBuilder) =


russcam:
Thanks for submitting a PR @kwilliamssbs. I'll take a look now

build/scripts/Releasing.fsx
```diff
@@ -34,10 +34,8 @@ module Release =
 
             let year = sprintf "%i" DateTime.UtcNow.Year
 
-            let jsonDotNetVersion = 10
-
-            let jsonDotNetCurrentVersion = sprintf "%i" jsonDotNetVersion
-            let jsonDotNetNextVersion = sprintf "%i" (jsonDotNetVersion + 1)
+            let jsonDotNetCurrentVersion = "10.0.1"
+            let jsonDotNetNextVersion = "11.0"
```
russcam: I think this would be better derived from the current version, as before.

build/scripts/Releasing.fsx
```diff
@@ -34,10 +34,8 @@ module Release =
 
             let year = sprintf "%i" DateTime.UtcNow.Year
 
-            let jsonDotNetVersion = 10
-
-            let jsonDotNetCurrentVersion = sprintf "%i" jsonDotNetVersion
-            let jsonDotNetNextVersion = sprintf "%i" (jsonDotNetVersion + 1)
+            let jsonDotNetCurrentVersion = "10.0.1"
```
russcam: It would be awesome if this value were retrieved from the `<PackageReference Include="Newtonsoft.Json" Version="10.0.1" />` in Nest.csproj. Using an XML type provider (or even just LINQ to XML), could grab the current version from thecsproj. Then, using `SemVerHelper` from F# FAKE, can parse the semantic version to get the next major version for `jsonDotNetNextVersion` .

kwilliamssbs: OK cool, I'll have a go at that and see how it looks

russcam: Awesome 😃 ping me if you have questions

russcam:
@kwilliamssbs I have a bit of time and can make the necessary changes, if you haven't already started?

kwilliamssbs:
@russcam that would be great, although I feel a little guilty for proposing the PR and then making you do all the work :)

russcam:
@kwilliamssbs I don't want to infringe if you're already working on it, or discourage you from opening a PR in the future 😄

russcam:
Closing this in favour of #2865. Thank you @kwilliamssbs for kicking this off 👍

kwilliamssbs:
No problem, I don’t think my F#-fu was quite up to the challenge, but glad to have gotten this moving!

PR #2872, opened 2017-10-11
Deserialize error property on MultiGetHit

### LINKED ISSUE #2871: Deserialize error property on MultiGetResponse items
[From stackoverflow.](https://stackoverflow.com/questions/46682274/elasticsearch-nest-errors-are-ignored-in-multigetrequests)

`error` properties on items in the `docs` array on an `IMultiGetResponse` are not currently deserialized onto `IMultiGetHit<object>`, resulting in no way to determine if an individual get response has resulted in an error. Further, `IMultiGetResponse` is still determined to be valid (`.IsValid` is true on response). 

I have a fix underway for this and will open a PR tomorrow.

russcam:
Closes #2871

### DIFF
--- a/src/Nest/Document/Multiple/MultiGet/Response/MultiGetHit.cs
+++ b/src/Nest/Document/Multiple/MultiGet/Response/MultiGetHit.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using Elasticsearch.Net;
 using Newtonsoft.Json;
 
 namespace Nest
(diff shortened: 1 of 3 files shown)


src/Nest/Document/Multiple/MultiGet/Response/MultiGetHit.cs
```diff
@@ -34,22 +37,22 @@ public class MultiGetHit<T> : IMultiGetHit<T>
 	{
 		public FieldValues Fields { get; internal set; }
 
-		[JsonProperty(PropertyName = "_source")]
```
Mpdreamz: 👍

PR #2878, opened 2017-10-19
Set Format when Missing is specified

### LINKED ISSUE #2864: Append date_optional_time to Format when Missing specified
from [stack overflow](https://stackoverflow.com/questions/46316244/missing-method-datehistogramaggregationdescriptor-doesnt-seems-to-work-in-a-s).

When using DateHistogram aggregation where ExtendedBounds are specified, [the `date_optional_time` format is appended to the `Format`](https://github.com/elastic/elasticsearch-net/blob/9a00e6ce3496b1d1d13e8240b735e7d3190f0c02/src/Nest/Aggregations/Bucket/DateHistogram/DateHistogramAggregation.cs#L54-L65), such that Elasticsearch can deserialized the ExtendedBounds. This is ultimately because the bounds are not serialized using the given format.

`date_optional_time` should also be appended to the `Format` if `Missing` has a value.

russcam:
Closes #2864

### DIFF
--- a/src/Nest/Aggregations/Bucket/DateHistogram/DateHistogramAggregation.cs
+++ b/src/Nest/Aggregations/Bucket/DateHistogram/DateHistogramAggregation.cs
@@ -53,15 +53,12 @@ namespace Nest
 
 		public string Format
 		{
-			get
-			{
-				return !string.IsNullOrEmpty(_format) &&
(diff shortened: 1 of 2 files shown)


Mpdreamz:
ported to `2.x` and `master`

PR #2891, opened 2017-11-08
Disable AllowWriteStreamBuffering 

russcam:
- Disable `AllowWriteStreamBuffering` on HttpWebRequest as requests do not require retrying with buffered content.
- Remove superfluous Keep-Alive header on .NET Core HttpConnection


src/Elasticsearch.Net/Connection/HttpConnection-CoreFx.cs
```diff
@@ -222,9 +222,6 @@ protected virtual HttpRequestMessage CreateRequestMessage(RequestData requestDat
 			}
 			requestMessage.Headers.Connection.Clear();
 			requestMessage.Headers.ConnectionClose = false;
-			requestMessage.Headers.Connection.Add("Keep-Alive");
```
Mpdreamz: Happy we have integration tests for this now :)

PR #2906, opened 2017-11-17
Feature/dependencies scrub

Mpdreamz:
This PR shades our `Newtonsoft.Json` dependency. Meaning its internalized and its namespaces are renamed to `Nest.Json`. 

image

I've written new docs on how its implemented and how to use it here:

https://github.com/elastic/elasticsearch-net/blob/0eae3c71a23e0e0deeb8b8a7bf68630dad261926/src/Tests/ClientConcepts/HighLevel/Serialization/CustomSerialization.doc.cs

To facilitate testing this our integration tests are now truly random. Before our integration tests seeded Bogus with `1337`. Now we use a random seed on each run and report it back on the commandline as part of the `build` command to repeat the failing run.

The integration tests now randomly uses a custom `SourceSerializer` and if it does asserts in the expected places whether this `SourceSerializer` reads and writes the extra properties only it knows about.

This PR also introduces a new package `NEST.JsonNetSerializer` which helps provide an instance `IElasticsearchSerializer` that knows how to handle `NEST` types as part of the `_source`.


Mpdreamz:
[zip of the nupkg's this PR produces](https://github.com/elastic/elasticsearch-net/files/1481986/_packages.zip)

Mpdreamz:
Fixed the type load exceptions, attribute constructors needed to be touched too.

Found a preliminary way to run our tests using the staged local NuGet packages at the end of the release rather than the `<ProjectReference>` it will normally use.

For backwards compatibility sake we'd need to support `Newtonsoft.Json.JsonPropertyAttribute` and `Newtonsoft.Json.Converters.JsonConverterAttribute` attributes on user types in `NEST` even if those types are not available. This will require some dynamic typeloading and reflection based proxies in the case of  `Newtonsoft.Json.Converters.JsonConverter`. 

or should we drop this completely and require folks to reference `NEST.JsonNetSerializer` ?

I'm leaning heavily towards the latter, thoughts @codebrain @russcam ?

Mpdreamz:
https://github.com/elastic/elasticsearch-net/pull/2912 needs merging in this PR before pulling.

PR #2909, opened 2017-11-20
Set Transfer-Encoding: chunked when a POST/PUT with body

### LINKED ISSUE #2908: Calling Search() or Index() throws ProtocolViolationException Exception
Preview
NEST/Elasticsearch.Net version: 6.1.x

Elasticsearch version: 6.0

Description of the problem including expected versus actual behavior: Any index or search operations made using NEST result in the following error:-

ProtocolViolationException: When performing a write operation with AllowWriteStreamBuffering set to false, you must either set ContentLength to a non-negative number or set SendChunked to true.

Steps to reproduce:

Create a new ElasticClient
Call Search() or Index()
Provide ConnectionSettings (if relevant):
var settings = new ConnectionSettings(new Uri("http://localhost:9200")).BasicAuthentication("elastic", "password");

russcam:
AllowWriteStreamBuffering = false means the request is not buffered before wirting to the request stream. As a consequence, the .NET Framework does not calculate a Content-Length value for the request, so it either needs to be calculated upfront and set, or the request should use chunked transfer encoding by setting SendChunked = true. This commit adds the latter.

Fixes #2908

### DIFF
--- a/src/Elasticsearch.Net/Connection/HttpConnection.cs
+++ b/src/Elasticsearch.Net/Connection/HttpConnection.cs
@@ -65,7 +65,6 @@ namespace Elasticsearch.Net
 			if (requestData.HttpCompression)
 			{
 				request.AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate;
-				request.Headers.Add("Accept-Encoding", "gzip,deflate");
 				request.Headers.Add("Content-Encoding", "gzip");
 			}
 			if (!requestData.RunAs.IsNullOrEmpty())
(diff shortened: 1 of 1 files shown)


src/Elasticsearch.Net/Connection/HttpConnection.cs
```diff
@@ -66,7 +66,6 @@ protected virtual HttpWebRequest CreateWebRequest(RequestData requestData)
 			if (requestData.HttpCompression)
 			{
 				request.AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate;
-				request.Headers.Add("Accept-Encoding", "gzip,deflate");
```
codebrain: Did you mean to delete this too?

russcam: Yes, `request.AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate;` sets the `"Accept-Encoding"` header

src/Elasticsearch.Net/Connection/HttpConnection.cs
```diff
@@ -80,12 +79,18 @@ protected virtual HttpWebRequest CreateWebRequest(RequestData requestData)
...
+			//which goes against RFC's and might break i.e IIS when used as a proxy.
+			//see: https://github.com/elasticsearch/elasticsearch-net/issues/562.
+			//Transfer-Encoding: chunked must be set for POST and PUT where there is a body.
+			var method = requestData.Method.GetStringValue();
+			request.Method = method;
+			if (method != "HEAD" && method != "GET")
+			{
+				if (requestData.PostData == null)
+					request.ContentLength = 0;
+				else
```
Mpdreamz: Does this need [SupportsTransferEncodingChunked](https://msdn.microsoft.com/en-us/library/system.net.http.httpclienthandlerextensions.supportstransferencodingchunked(v=vs.118).aspx) and if so should `AllowWriteStreamBuffering = false` depend on that too?

src/Elasticsearch.Net/Connection/HttpConnection.cs
codebrain: :+1:

russcam:
look at after 6.x release

russcam:
Added `AllowWriteStreamBuffering=false` to this PR, to keep in the same place as `SendChunked = true`

russcam:
I'm going to close this PR for now; the underlying HTTP implementations within newer versions of .NET Core have been rewritten, and the classic `HttpConnection` built on top of `HttpWebRequest` will likely be sunsetted down the line in light of this. We can always reopen in the future to revisit if necessary.

PR #2911, opened 2017-11-21
Set HttpClientHandler.UseProxy=false to disable automatic proxy detection

codeh:
With HttpClientHandler.Proxy = null HttClient still tries to detect and use system proxy.
I've fixed this by using HttpClientHandler.UseProxy flag, which prevents HttpClient from using proxy at all.
"Else if" is used to avoid preventing usage of proxy, if client setted ProxyAddress.


codebrain:
Looks good - shall I merge in, and bring into `master` and `2.x` ? @Mpdreamz

russcam:
yes please @codebrain

russcam:
needs porting to 5.x too (PR targets 5.5)

codebrain:
Ported to `5.x`, `2.x` and `master`

PR #2912, opened 2017-11-21
Feature/unit test packages release

Mpdreamz:
* Canary builds now run the unit tests using `Nest` as `<ProjectReference>` and as `<PackageReference>`. This validates the nuget package that has `Json.NET` il merged and its namespaces renamed.

* `Rename` `Ignore` attributes, since `[JsonProperty]` by default references the internal `Nest.Json.JsonPropertyAttribute` we can no longer support this OOTB, so we include `[Rename]` and `[Ignore]` as mechanism to control the internalized `Json.NET`.  Folks can always reference `Nest.JsonNetSerializer` to keep using `[JsonProperty]`.

* Some useful converters can be applied with `[StringEnum]`, `[TimeSpanString]` and `[MachineLearningDateTime]`, further more `enum` properties that have `[Text]` `[Keyword]` applied will default to using the internal `StringEnum` converter and `AutoMap()` takes this as a hint to map them as a keyword/text as well.

Mind if you run `build canary` locally your `%USERPROFILE%\.nuget\packages\NEST\` folder will cache each `-ci<date>` version you run the test against.


Mpdreamz:
Merging this in so we can continue on #2906

Issue #2656, opened 2017-03-07
IndicesCreatePost not working with JSON string


GaryTurpin:
**NEST/Elasticsearch.Net version**:
2.5.2

**Elasticsearch version**:
2.4.4

**Description of the problem including expected versus actual behavior**:
I am upgrading from ES 1.X to ES 2.X and I am having a problem creating an index with mappings/settings from JSON.

In ES 1.X Nest, I would do the following and everything would work perfectly.
```
public bool CreateIndex(string json)
{
  var client = new Nest.ElasticClient(new ConnectionSettings(this.ESIndex.ClusterUri));
  var resp = client.Raw.IndicesCreatePost(this.ESIndex.SyncIndex, json);
  return resp.Success;
}
```

In ES 2.X Nest, I do the following and the mappings/settings never get added to the blank index.
```
public bool CreateIndex<T>(string json) where T : class
{
  var client = new Nest.ElasticClient(this.ESIndex.ClusterUri);
  var body = new PostData<string>(json);
  var resp = client.LowLevel.IndicesCreatePost<T>(this.ESIndex.SyncIndex, body);
  return resp.Success;
}
```

If I look at fiddler, it looks like the only thing in the body of the request is
`{"type":1}`

But, I expect to have the JSON that have for my mappings/settings.  Any idea what I am doing incorrectly so I can make this work?

Mpdreamz:
This is a little confusing but its because of `PostData<string>`

```csharp
  var body = new PostData<object>(json);
var resp = client.LowLevel.IndicesCreatePost<string>("someindex5", body);
```

Will work or just sending json directly as the second argument. 

It happens because that `PostData<string>` gets passed internally to argument taking `PostData<object>` and the implicit operator for `T` kicks in wrapping it.

PR #2916, opened 2017-11-23
Prevent boxing on PostData<object>

Mpdreamz:
`PostData<object>` was introduced as a union'esque type that
`byte[]/string/object/IEnumerable<string>/<IEnumerable<object>` could
all implicitly convert to automagically making the lowel client easy
to use without exploding the already massive method list on
ILowLevelClient.

This abstraction meant that we always boxed for NEST types

In preparation for all the serialization/perf work we want to do in 6.x
it would be great if we can take `T` that we we want to serialize all
the way to the `IElasticsearchSerializer` which we can do with this PR
now.

We sacrifice on the implicit operators in the low level client (for the
most part: byte[] and string still do). TBF most of the time I've seen
people use the low level client they would instantiate
`PostData<object>` directly already. Or worse see: #2656. We weren't
really helping anyone here.

`PostData<object>` now has a base class `PostData` that we use
throughout. There is also `SerializableData<T>` subclass that we now use
for anything that needs serializing.

This dispatcher is now aware of what generics an `IRequest`
implementation has. Because of this our hack to make `Index(document)`
and `CreateIndex(document)` work when document == IIndexRequest or not
no longer works. Rather than fighting C# resolution of overloads (and
its lack of not constraints :() the methods are now called
`IndexDocument()` and `CreateDocument()`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ solution: src/Elasticsearch.sln
 script: ./build.sh test
 dist: trusty
 mono: 4.6.2
-dotnet: 1.0.1
+dotnet: 2.0.3
 env:
   global:     
     - DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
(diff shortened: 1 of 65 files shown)


Mpdreamz:
Merged after walk through with @codebrain and @russcam

PR #2917, opened 2017-11-23
AutoMap() maxRecursion defaulted to 0 which may not be a safe default…

Mpdreamz:
…, chose 20 instead. 20 levels deep oughta be enough for everyone right? :)

https://stackoverflow.com/questions/47059094/elastic-nest-automap/47462235#47462235


Mpdreamz:
Scratch this, `0` is means `no recursion on the same type` already :)

PR #2922, opened 2017-11-30
Allow non-string values for Sort "missing" field

### LINKED ISSUE #2857: Support non-string values for sort "missing"
[From stackoverflow](https://stackoverflow.com/questions/46222510/secondary-sorting-not-working-correctly-for-elasticsearch-documents).

[Sort "missing" field should support other types such as `int` in addition to `string`](https://github.com/elastic/elasticsearch/blob/v5.0.0/core/src/main/java/org/elasticsearch/search/sort/FieldSortBuilder.java#L110)

russcam:
This PR allows non string values to be passed for the value of the "missing" field by using `object` as the property type.

Rename `PinTo()` to `Points()` to align the fluent method for setting Points to the object initializer name.

Requires backporting to 2.x and 5.x in a non-binary breaking way

Closes #2857

### DIFF
--- a/src/Tests/Search/Request/SortUsageTests.cs
+++ b/src/Tests/Search/Request/SortUsageTests.cs
@@ -38,6 +38,12 @@ namespace Tests.Search.Request
 						}
 					},
 					new {
+						numberOfCommits = new {
+							missing = -1,
+							order = "desc"
+						}
+					},
+					new {
 						_geo_distance = new {
 							location = new [] {
 								new {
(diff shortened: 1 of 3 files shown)


src/Tests/Search/Request/SortUsageTests.cs
```diff
@@ -85,13 +91,18 @@ public class SortUsageTests : SearchUsageTestBase
...
+					.Missing(-1)
+				)
 				.GeoDistance(g => g
 					.Field(p => p.Location)
 					.DistanceType(GeoDistanceType.Arc)
 					.Order(SortOrder.Ascending)
 					.Unit(DistanceUnit.Centimeters)
 					.Mode(SortMode.Min)
-					.PinTo(new GeoLocation(70, -70), new GeoLocation(-12, 12))
+					.Points(new GeoLocation(70, -70), new GeoLocation(-12, 12))
```
Mpdreamz: 💯

russcam:
Opened https://github.com/elastic/elasticsearch-net/pull/2943 for the backport

PR #2924, opened 2017-12-01
Include similarity in index settings

### LINKED ISSUE #2890: Include Classic similarity in index settings
`CreateIndex()` allows a similarity to be set

```c#
client.CreateIndex("index", c => c
	.Similarity(s => s
		.BM25("default", ds => ds)
	)
);
```

but there appear to be two issues with this

1. No way to set the [Classic Similarity](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-similarity.html#classic-similarity) (looks like it is still referred to as Default in source)
2. Similarity settings should be under `settings.index` e.g.

```json
PUT /my_index
{
  "settings": {
    "index": {
      "similarity": {
        "default": {
          "type": "classic"
        }
      }
    }
  }
}
```

but NEST currently produces

```json
PUT http://localhost:9200/messages?pretty=true 
{
  "similarity": {
    "default": {
      "type": "BM25"
    }
  }
}
```

russcam:
Similarity settings are nested under settings>index
Rename DefaultSimilarity to ClassicSimilarity
Add DFISimilarity
Add similarity settings to create index integration tests
Add XML comments for each similarity
Refactor SimilarityJsonConverter to not look up the NEST ISimilarity type with Type.GetType() but to infer it from the type property value directly.
Add Similarity to the updatable index settings

Closes #2890


russcam:
Requires backporting to 2.x and 5.x in a non-binary breaking way

PR #2928, opened 2017-12-07
Update suggest-usage.asciidoc

gkinsman:
Added example of indexing completion suggestions. No existing docs on this, or a hint that a CompletionField is how you generate the term/weight request.

Found it in this forum post: https://discuss.elastic.co/t/using-nest-no-results/106191/10

### DIFF
--- a/docs/search/request/suggest-usage.asciidoc
+++ b/docs/search/request/suggest-usage.asciidoc
@@ -22,6 +22,36 @@ The suggest feature suggests similar looking terms based on a provided text by u
 See the Elasticsearch documentation on {ref_current}/search-suggesters.html[Suggesters] for more detail.
 
 [float]
+=== Indexing example
+
+[source,csharp]
+----
+public class Question
+{
+    public CompletionField TitleSuggest { get; set; }
+}
+
+var createIndexResponse = client.CreateIndex("my_index", c => c
+    .Mappings(m => m
+        .Map<Question>(u => u
+            .AutoMap()
+        )
+    )
+);
+
+var question = new Question
+{
+    TitleSuggest = new CompletionField
+    {
+        Input = new[] { "This is the title" },
+        Weight = 5
+    }
+};
+
+client.Index(question, i => i.Index("my_index").Refresh(Refresh.WaitFor));
+----
+
+[float]
 === Fluent DSL example
 
 [source,csharp]


gkinsman:
Signed CLA, awaiting update

russcam:
Thanks for opening the PR @gkinsman!

[Documentation is generated from tests](https://github.com/elastic/elasticsearch-net#generating-documentation-from-tests) to ensure that it stays in line with changes in the client. 

I think this would be a good candidate for a "Search as you type" section under [Search](https://www.elastic.co/guide/en/elasticsearch/client/net-api/current/search.html) as opposed to extending the Usage test. It would follow a narrative, much like the ["writing queries" section](https://github.com/elastic/elasticsearch-net/blob/master/src/Tests/Search/WritingQueries.doc.cs). Would you like to contribute such a section? This [Stack Overflow answer](https://stackoverflow.com/questions/41746660/elasticsearch-5-x-context-suggester-nest-net/41753197#41753197) could be used as a starting basis for it and such a section would include

1. What is required in the mapping to make completion suggester work i.e. `CompletionField`
2. How one queries using the Completion Suggester, showing both NEST API code and the JSON request
3. How to handle the response within NEST
4. Adding context with the Context Suggester

gkinsman:
Ah! I totally missed the note at the top of the file :). I like that strategy a lot. I can certainly give it a red hot go. It looks like the tests should all be runnable locally using the in memory client - is that correct?

I'll give it a go tonight and see what I can turn out :)

russcam:
Awesome 😄 It's been on the list of TODOs for some time so any PRs for documentation are hugely welcomed!

We have integration tests already for the completion suggester so for the documentation, using the `InMemoryConnection` would be a good approach. When running

```batch
.\build.bat
```
in the root, the unit tests will run, so will ensure all is as intended. Some pointers that I hope will help:

1. Probably want to derive from `SerializationTestBase` and implement `IClusterFixture<ReadOnlyCluster>` similar to the writing queries section.
2. `[U]` attribute for unit tests
3. Code that is needed for the tests but should not appear in documentation can be "attributed" with a `//hide` code comment. `private` members don't need this as the doc generation process only looks at `public` members.
4. Anonymous types can be serialized and shown as JSON in documentation by "attributing" the anonymous type with `//json`
5. Code annotations can be applied with `// <1> comment here` where the number should increment across a source block for each subsequent annotation.
6. AsciiDoc is strict about heading level ordering. Take a look at [the Elastic docs README](https://github.com/elastic/docs) and ping me if you get stuck.

As a heads up, the DocGeneration project does not currently run and emit the asciidoc files; `MsBuildWorkspace` does not properly support the newer csproj format, meaning that all of the  source file documents under the project directory do not get automatically picked up and added to the `Project` type. We're working on fixing this, likely moving back to `AdHocWorkspace` for the moment.

Thansk again!

russcam:
Heya @gkinsman, just a ping to see if you were still interested in looking at this?

gkinsman:
Ah hey, yes! Great to hear the doc gen is now fixed, I struggled a bit last time without it :). Will take a look this weekend I think - still using and loving elastic :).

PR #2929, opened 2017-12-07
Fix/server error

Mpdreamz:
⚠️ PR THAT IS WAY TOO BIG ⚠️ 

I let this branch grow in to a bit of a monstrosity, I'll try my best to summarize but as its nearly midnight here I might revise and explain more tomorrow. 

# Short circuit on error status codes are gone

With `Elasticsearch 2.x `and `NEST 2.x` we were able to build a short circuit in the low level client for bad error status codes to deserialize the `error/status` codes. This served us well but already in 5.x Elasticsearch responses started to break away from this model [and some API calls returned bodies](https://github.com/elastic/elasticsearch/issues/17548) on bad request that weren't just a deserialized `Throwable`. With 6.x this has only increased so this way of modeling responses in NEST no longer reflects reality.

# No more hand of between low and high level client

Because the low level client was responsible for deserializing bad status codes dispatched calls from NEST in to the low level client were basically always returned as `ElasticsearchResponse<NestResponse>` and unpacked. This never sat well with me this PR fixes that.

Both low level client and high level clients implement `IElasticsearchResponse` from the low level client which only dictates that such a response is able to get and set the `IApiCallDetails`, this is the object that tracks what url was hit, what the audit trail for the call is etc.

# BREAKING change in the low level client

Because the low level client now constraints `<T>` in `lowLevelClient.Search<T>` to be of type `IElasticsearchResponse` you can no longer pass `string` , `byte[]`, `dynamic` directly anymore. This seemed like a nice design when NEST 1.x released with feature parity of the other low level clients but that design dictated to much how the High Level client had to unwrap `<T>`. We no ship with dedicated `StringResponse`, `BytesResponse`, `StreamResponse`, `VoidResponse` and `DynamicResponse` in the low level client.

# BREAKING ServerError is not materialized on low level client.

`ServerError` is no longer part of `IApiCallDetails`. Because it was before we needed to branch in `ResponseBuilder` and possibly double read in certain cases (e.g ReindexOnServer API and others) it also used the built in slower low level json serializer for this. NEST is now fully in charge of deserializing bad responses too. 

In terms of NEST nothing has changed `ServerError` is still available. If you want it in the low level client `StringReponse` and `ByteResponse` have helpers or feed directly into `ServerError.Create()`.

# ResponseBuilder static

We now longer instantiate a `ResponseBuilder` for each request, the whole thing is now static and greatly simplified.

# Improved error parsing

[I investigated what additional data can be returned `error`'s in elasticsearch](https://github.com/elastic/elasticsearch/issues/27672) and mapped all of these on either `Error.Headers` or `Error.Metadata`. There are tests for these that cover both the low and high level client.

# ResponseBuilder content type aware.

We now flow the servers response `content-type` to the response builder and make sure its what we expect before we serialize. We had tests for proxies returning HTML which would never break in previous NEST versions because of the short circuiting but now because we always deserialize the body we need to be able to account for this differently.

# No more AllowStatusCodes needed

We special cased some API's to be valid for 404's e.g all `Exists` api's and *some* `Get` API's. The latter especially was problematic because some API's were valid on 404 GET and some were not. This was done at a time where e.g a get for a document in an index that does not exist returned a 404 with a valid body. In 6.x however this returns a serialized `Throwable` e.g `error` and `status` property. Again because this PR cuts all of the special handling out and *always* flows the body to the serializer all the API's behave the same.


src/CodeGeneration/ApiGenerator/Views/ElasticLowLevelClient.Generated.cshtml
```diff
@@ -16,8 +16,7 @@ using System.Threading.Tasks;
 using Elasticsearch.Net;
 using static Elasticsearch.Net.HttpMethod;
 
-//This file is automatically generated from https://github.com/elastic/elasticsearch/tree/@Model.Commit/rest-api-spec
-//Generated from commit @Model.Commit
+//This file is automatically generated from https://github.com/elastic/elasticsearch/tree/master/rest-api-spec
```
russcam: Should this include `@Model.Commit`?

Mpdreamz: No since I quite often end up copying the files over manually (especially from x-pack) this goes out of sync.  I want to set up a public repos with the rest spec that is sync'ed automatically to publish them more formally.

Mpdreamz:
rebased against `master`

Mpdreamz:
merged after run down with @russcam and @codebrain over zoom. Opting not to squash these commits to keep the history of the individual commits.

PR #2930, opened 2017-12-08
Do not capture synchronization context on async calls

### LINKED ISSUE #2904: ScrollAll Continues Before Retrieving All Matching Documents
**NEST/Elasticsearch.Net version**: 5.5.0

**Elasticsearch version**: 5.6.1

ScrollAll<T> not waiting for final slice, no matter how long the maximumRunTime is set for. Instead it continues past the Wait and then requests the final slice of documents from ES.

**Steps to reproduce**:
 1. Populate index with ~3,500 documents across 3 shards.
 2. Perform ScrollAll request
```
var docList = new List<docType>();
var scrollObserver = nestClient.ScrollAll<docType>("10s", 3,
                                s => s.Search(se => se.Query(q =>
                                    +q.Term(t => t.Field(f => f.TheField).Value(true))
                                ).Size(10000)).MaxDegreeOfParallelism(1))
                            .Wait(TimeSpan.FromSeconds(15),
                                r =>
                                {
                                    docList.AddRange(r.SearchResponse.Documents.ToList());
                                });
								
return docList;
```
 3. The first two scroll requests complete almost immediately (requests# 2605 and 2610 in the screenshot). The third request to retrieve the last documents doesn't occur until after the wait period has elapsed, no matter how long the wait time is (request# 2800). 

```
var nestConfig = new ConnectionSettings(connectionPool)
                .RequestTimeout(TimeSpan.FromSeconds(120))
                .DefaultIndex("index").MapDefaultTypeIndices(m => m.Add(typeof(docType), "index"));
```

untitled

russcam:
Relates elastic/support-dev-help#2934

The sychronization context is captured in some of the asychronous calls made within the paths of the *All observable methods. Whilst this does not exhibit problems with the synchronization context within a console application, it causes problems with the synchronization context within ASP.NET 4.5 MVC, resulting in the response to an MVC controller to be returned before the last response is returned from Elasticsearch:

screenshot 2017-12-08 15 05 39

Whilst there are integration tests for *All observable methods, this was not picked up because async methods within tests run with the same synchronization context as a console application.

With this PR, the behaviour is now correct within ASP.NET 4.5. MVC:

screenshot 2017-12-08 15 13 40

Other changes:

- Continue only when the count of tasks is less than degrees of parallelism
- Initialize the size of the task list to the degrees of parallelism

Closes #2904

### DIFF
--- a/src/Nest/CommonAbstractions/Extensions/Extensions.cs
+++ b/src/Nest/CommonAbstractions/Extensions/Extensions.cs
@@ -235,18 +235,18 @@ namespace Nest
 
 			try
 			{
-				var tasks = new List<Task>();
+				var tasks = new List<Task>(maxDegreeOfParallelism);
 				foreach (var item in lazyList)
 				{
 					tasks.Add(ProcessAsync(item, taskSelector, resultProcessor, semaphore, additionalRateLimitter, page++));
-					if (tasks.Count <= maxDegreeOfParallelism)
+					if (tasks.Count < maxDegreeOfParallelism)
 						continue;
 
-					var task = await Task.WhenAny(tasks);
+					var task = await Task.WhenAny(tasks).ConfigureAwait(false);
 					tasks.Remove(task);
 				}
 
-				await Task.WhenAll(tasks);
+				await Task.WhenAll(tasks).ConfigureAwait(false);
 				done(null);
 			}
 			catch (Exception e)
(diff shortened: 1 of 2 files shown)


russcam:
Check to see if this also requires porting to 2.x and master.

Mpdreamz:
ported to `2.x` and `master` and did another sweep on those branches as well.

Issue #2926, opened 2017-12-06
ConnectionSettings.MapDefaultTypeIndices in NEST 6.x


tdoman:
Hello all, I'm trying out the bleeding edge here and testing NEST 6.x w/ ES 6.  I'm using NEST 6.1.0-ci1205045848.

Where did ConnectionSettings.MapDefaultTypeIndices go?  I see `InferMappingFor<T>` is there but that would require me to do something at compile time that I was previously able to do at run time from another list that I'm already required to create to keep track of for other per index things (ie. search fields and aggregations I use, the `Type`, index name, etc.).  For example, "IndexTypes" is that list, so I can do the following:
`IndexTypes.ForEach(x => settings.MapDefaultTypeIndices(d => d.Add(x.Type, x.Index)));`
What's the 6.0 way?

russcam:
Hey @tdoman, it looks like we should look at exposing a non-generic `InferMappingFor()` method that allows `Type` to be passed.

In the meantime, connection settings can be cast to `IConnectionSettingsValues` interface to add the mappings directly

```c#
    var pool = new SingleNodeConnectionPool(new Uri("http://localhost:9200"));
    var connectionSettings = new ConnectionSettings(pool);

    IConnectionSettingsValues settings = connectionSettings;   
    settings.DefaultIndices.Add(typeof(Message), "foobar");

    var client = new ElasticClient(connectionSettings);
```

tdoman:
@russcam Ok, thanks Russ I'll do that until there's a non-generic `InferMappingFor()`.  BTW, I assume that even after types go away in 7.0, we'll still need a way to specify which POCO to use for each index via NEST no?
On a separate note, as far as I can tell, https://github.com/elastic/elasticsearch-net/issues/2927 is my only blocker from having all the functionality that we use working in an ES 6.0 system.  Once I get past that I may discover something else but we're close.

Mpdreamz:
Thanks for opening this @tdoman, opened https://github.com/elastic/elasticsearch-net/pull/2934 to address this.

PR #2934, opened 2017-12-08
can now map clr types dynamically on connection settings including an…

Mpdreamz:
… overload that takes a list of mappings to ease more dynamic configuration of types at run time

fixes: https://github.com/elastic/elasticsearch-net/issues/2926


src/Nest/CommonAbstractions/ConnectionSettings/PocoMapping.cs
```diff
@@ -0,0 +1,194 @@
...
+
+		/// <summary>
+		/// Allows you to set a default Id property on <typeparamref name="T" /> that NEST will evaluate
+		/// </summary>
+		public Expression<Func<T, object>> IdProperty { get; set; }
+
+		/// <summary>
+		/// When specified allows you to ignore or rename certain properties of clr type <typeparamref name="T" />
+		/// </summary>
+		public IList<IPocoPropertyMapping<T>> Properties { get; set; }
```
russcam: Should the non-generic `PocoMapping` also expose a `Properties` collection, since ultimately the properties are added to a `FluentDictionary<MemberInfo, IPropertyMapping>`?

src/Nest/CommonAbstractions/ConnectionSettings/PocoMapping.cs
```diff
@@ -0,0 +1,194 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+
+namespace Nest
+{
+	public interface IPocoMapping
```
russcam: I thought `ClrTypeMapping` was fairly descriptive 😄

PR #2943, opened 2017-12-12
Support non string values for sort missing

### LINKED ISSUE #2857: Support non-string values for sort "missing"

russcam:
Non-binary breaking backport
Closes #2857

(cherry picked from commit f5959d66c15574a23595cdee2777f6987319775b)

### DIFF
--- a/src/Nest/Search/Search/Sort/SortBase.cs
+++ b/src/Nest/Search/Search/Sort/SortBase.cs
@@ -4,47 +4,103 @@ using Newtonsoft.Json;
 
 namespace Nest
 {
+	/// <summary>
+	/// Allows to add one or more sort on specific fields.
+	/// </summary>
 	[JsonObject(MemberSerialization = MemberSerialization.OptIn)]
 	[ContractJsonConverter(typeof(SortJsonConverter))]
 	public interface ISort
 	{
+		/// <summary>
+		/// The field to sort on
+		/// </summary>
 		Field SortKey { get; }
 
-		[JsonProperty("missing")]
+		/// <summary>
+		/// Specifies how docs which are missing the field should be treated
+		/// </summary>
+		[JsonIgnore]
+		[Obsolete("Use MissingValue")]
 		string Missing { get; set; }
 
+		/// <summary>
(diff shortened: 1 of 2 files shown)


src/Nest/Search/Search/Sort/SortBase.cs
```diff
@@ -4,47 +4,103 @@
...
 
+	/// <inheritdoc/>
 	public abstract class SortBase : ISort
 	{
-		public string Missing { get; set; }
+		/// <inheritdoc/>
+		[Obsolete("Use MissingValue")]
+		public string Missing
+		{
+			get => MissingValue as string;
```
codebrain: `.ToString()` ?

src/Nest/Search/Search/Sort/SortBase.cs
```diff
@@ -4,47 +4,103 @@
...
 		where TDescriptor : SortDescriptorBase<TDescriptor, TInterface, T>, TInterface, ISort
 		where TInterface : class, ISort
 	{
 		Field ISort.SortKey => this.SortKey;
 
-		string ISort.Missing { get; set; }
+		[Obsolete("Use MissingValue")]
+		string ISort.Missing
+		{
+			get => Self.MissingValue as string;
```
codebrain: `.ToString()` ?

src/Nest/Search/Search/Sort/SortBase.cs
russcam: These shouldn't be `.ToString()` in my opinion because it means a non-string value set with `MissingValue` can be retrieved as a `string` using `Missing`.

Mpdreamz:
merged to 5.x and 2.x

PR #2946, opened 2017-12-12
fix #2817 allow {dynamic_type} to be set when doing a generic mapping…

### LINKED ISSUE #2817: Setting `{dynamic_type}` as part of the mapping
Reported on stackoverflow:  https://stackoverflow.com/questions/44933356/create-a-dynamic-type-mapping-in-elasticsearch-with-nest

The user is not able to generate the following using our `.Generic` construct that can map dynamic mappings like these.

```json
"mapping": {
    "type": "{dynamic_type}",
    "index": "not_analyzed",
    "doc_values": true
}
```

Mpdreamz:
…, cleaned up the obsolete constructors for PropertyBase(Descriptor) in the process as well'


Mpdreamz:
Needs a manual back port PR for `5.x`

russcam:
LGTM. Merged

PR #2947, opened 2017-12-12
upgrade to the latest greatest bogus to take advantage of local seeds…

Mpdreamz:
… and clone(), hopefully this solves our deadlock issues in Bogus's Generate() call chain


Mpdreamz:
back ported to `2.x` and `5.x`

PR #2950, opened 2017-12-13
Generate documentation again

russcam:
With the move to the newer csproj format, the DocGenerator project no longer generated documentation from the Tests source documents, due to Roslyn's `MsBuildWorkspace` lack of support for the new project format; it is possible to get the project correctly compiling, but none of the projects within a solution have documents associated with them.

This PR gets documentation generation working again by using components from Buildalyzer to create an `AdhocWorkspace`. Source documents within the Tests project still need to be manually added to the project because Buildalyzer also does not add documents within a project directory by convention either, similar to `MsBuildWorkspace`.

The DocGenerator project requires that the solution be compiled before running, so that the XML comment files exist on the filesystem. Whilst the process will not fail, since Source documents are not added to the Nest and Elasticsearch.Net projects, compiling those projects to emit a documentation file will yield an empty file.

Doc generation should be revisited when Roslyn's support for the new project format has improved.

### DIFF
--- a/src/CodeGeneration/DocGenerator/Buildalyzer/AnalyzerManager.cs
+++ b/src/CodeGeneration/DocGenerator/Buildalyzer/AnalyzerManager.cs
@@ -1,0 +1,163 @@
+﻿#region License
+//MIT License
+//
+//Copyright (c) 2017 Dave Glick
+//
+//Permission is hereby granted, free of charge, to any person obtaining a copy
+//of this software and associated documentation files (the "Software"), to deal
+//in the Software without restriction, including without limitation the rights
+//to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//copies of the Software, and to permit persons to whom the Software is
+//furnished to do so, subject to the following conditions:
+//
+//The above copyright notice and this permission notice shall be included in all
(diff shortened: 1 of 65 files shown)


src/CodeGeneration/DocGenerator/Buildalyzer/AnalyzerManager.cs
```diff
@@ -0,0 +1,163 @@
...
+using System.Linq;
+using System.Xml.Linq;
+using DocGenerator.Buildalyzer.Logging;
+using Microsoft.Build.Construction;
+using Microsoft.Build.Framework;
+using Microsoft.Extensions.Logging;
+
+namespace DocGenerator.Buildalyzer
+{
+ public class AnalyzerManager
```
Mpdreamz: Whats the reason we source include this dependency?

Mpdreamz:
rebased the commits into 1 except the documentation update commit so https://github.com/elastic/elasticsearch-net/commit/d2eecfdf3c1e05d3004a5d17e24f806fd1107241 is easier to cherry-pick to other branches.

Mpdreamz:
ported to `2x` and `master`

Issue #2935, opened 2017-12-08
.Lang() not allowed to be specified as parameter in 6.0 on update request.


tdoman:
NEST 6.1.0-ci1208153011

The NEST library allows it to be specified but ES 6.0 responds with:
"request [/myindex/mytype/1/_update] contains unrecognized parameter: [lang]"

`var response = await client.UpdateAsync<T, object>(id, u => u.Index(index).Type(type).Lang("painless").Script(s => s.Indexed(scriptFile).Params(parms)).RetryOnConflict(5));`

tdoman:
I'm seeing this in some other contexts in my code as well where the .Lang() function is being exposed in script contexts where it shouldn't.  [lang can no longer be specified...](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_60_scripting_changes.html#_literal_lang_literal_can_no_longer_be_specified_when_using_a_stored_script_as_part_of_a_request)

russcam:
Hey @tdoman, `Lang()` is allowed in the body, but not as a query string parameter.
Merged 6.x script changes in #2951 and #2933.

Thanks for reporting this 👍

russcam:
Apologies @tdoman, my last comment was incorrect. `Lang()` is not allowed in places that reference a script. It is allowed in the Put Script API

tdoman:
@russcam Yep, NEST had been allowing me to specify `.Lang()` but ES 6.0 was rejecting it as a query string parameter.  Thanks for fixing all these, are they in the 12/14 build?  If so, I can try 'em all out.  I believe this will also represent 100% coverage on all our current usages, woot!

tdoman:
@russcam I grabbed the 12/14 build, looks great!  Just more thorough testing to do, thanks for all the work, looking forward to the official ship of NEST 6.0!

russcam:
@tdoman [NEST-6.0.0-beta1 is up on Nuget](https://www.nuget.org/packages/NEST/6.0.0-beta1) if you're like to take that for a spin; it's very near to what we expect to be in the GA release and we have released it to nuget in order to garner wider feedback.

tdoman:
Sweet, I will pull that from Nuget and give it a spin, thanks Russ!

Issue #2923, opened 2017-11-30
6.0 Breaking changes check


codebrain:
# [Breaking Changes](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-6.0.html)

## [Aggregations changes](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_60_aggregations_changes.html)

- [x] Deprecated pattern element of include/exclude for terms aggregations has been removed
- [x] Numeric to and from parameters in date_range aggregation are interpreted according to format now

## [Cat API changes](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_60_cat_changes.html)

- [x] Unbounded queue size in cat thread pool

## [Clients changes](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_60_clients_changes.html)

- [x] Java High Level REST Client

## [Cluster changes](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_60_cluster_changes.html)

- [x] Cluster name no longer allowed in path.data

## [Document API changes](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_60_docs_changes.html)

- [x] `version` type force remove - **Deleted enum value**
- [x] Upserts no longer support versions
- [x] `created` field removed in the Index API
- [x] `found` field removed in the Delete API - **Deleted property**

## [Indices changes](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_60_indices_changes.html)

- [x] Index templates use `index_patterns` instead of `template`
- [x] Shadow Replicas have been removed
- [x] Open/Close index API allows wildcard expressions that match no indices by default
- [x] Delete a document
- [x] Indices aliases api resolves indices expressions only against indices
- [x] Delete index api resolves indices expressions only against indices
- [x] Support for + has been removed in index expressions
- [x] Translog retention

## [Ingest changes](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_60_ingest_changes.html)

- [x] Timestamp meta-data field type has changed
- [x] The format of the string-formatted ingest.timestamp field has changed

## [Java API changes](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_60_java_changes.html)

- [x] `setSource` methods require XContentType
- [x] `DeleteByQueryRequest` requires an explicitly set query
- [x] `InternalStats` and `Stats` getCountAsString() method removed
[...]

PR #2951, opened 2017-12-13
6.0 Script changes

codebrain:
PR may contradict breaking change information here: https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_60_scripting_changes.html#_stored_search_template_apis_removed

---

- Groovy, JavaScript, and Python languages removed
The Groovy, JavaScript, and Python scripting languages were deprecated in elasticsearch 5.0 and have now been removed. Use painless instead.

- lang can no longer be specified when using a stored script as part of a request
The lang variable can no longer be specified as part of a request that uses a stored script otherwise an error will occur. Note that a request using a stored script is different from a request that puts a stored script. The language of the script has already been stored as part of the cluster state and an id is sufficient to access all of the information necessary to execute a stored script.

- lang can no longer be used when putting, getting, or deleting a stored script
Stored scripts can no longer have the lang parameter specified as part of the url when performing PUT, GET, and DELETE actions on the _scripts/ path. All stored scripts must have a unique id as the namespace is only id now and no longer lang and id.

Addresses: https://github.com/elastic/elasticsearch-net/issues/2935
and items on https://github.com/elastic/elasticsearch-net/issues/2923

### DIFF
--- a/src/Nest/Modules/Scripting/IStoredScript.cs
+++ b/src/Nest/Modules/Scripting/IStoredScript.cs
@@ -1,4 +1,5 @@
-﻿using Newtonsoft.Json;
+﻿using Elasticsearch.Net;
+using Newtonsoft.Json;
 
 namespace Nest
 {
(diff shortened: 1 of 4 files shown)


src/Nest/Modules/Scripting/IStoredScript.cs
```diff
@@ -1,74 +1,37 @@
...
-using Newtonsoft.Json;
+﻿using Newtonsoft.Json;
 
 namespace Nest
 {
 	[JsonObject(MemberSerialization = MemberSerialization.OptIn)]
 	[JsonConverter(typeof(ReadAsTypeJsonConverter<StoredScript>))]
 	public interface IStoredScript
 	{
-		[JsonProperty("lang")]
```
Mpdreamz: You can still specify lang as part of the body, its just no longer part of the path in the url:

https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/script/StoredScriptSource.java#L157

src/Nest/Modules/Scripting/IStoredScript.cs
```diff
@@ -1,74 +1,37 @@
...
-	{
-		private static readonly string Lang = ScriptLang.Python.GetStringValue();
-		public PythonScript(string source) : base(Lang, source) { }
-	}
-	public class LuceneExpressionScript : StoredScript
-	{
-		private static readonly string Lang = ScriptLang.Expression.GetStringValue();
-		public LuceneExpressionScript(string source) : base(Lang, source) { }
-	}
-	public class MustacheScript : StoredScript
```
Mpdreamz: I think we should keep this around
https://github.com/elastic/elasticsearch/blob/master/modules/lang-mustache/src/main/java/org/elasticsearch/script/mustache/MustacheScriptEngine.java

src/Nest/Modules/Scripting/IStoredScript.cs
```diff
@@ -1,74 +1,37 @@
...
-	{
-		private static readonly string Lang = ScriptLang.JS.GetStringValue();
-		public JavaScriptScript(string source) : base(Lang, source) { }
-	}
-	public class PythonScript : StoredScript
-	{
-		private static readonly string Lang = ScriptLang.Python.GetStringValue();
-		public PythonScript(string source) : base(Lang, source) { }
-	}
-	public class LuceneExpressionScript : StoredScript
```
Mpdreamz: I think we should keep this one around:

https://github.com/elastic/elasticsearch/blob/master/modules/lang-expression/src/main/java/org/elasticsearch/script/expression/ExpressionScriptEngine.java

src/Nest/Modules/Scripting/PutScript/PutScriptRequest.cs
```diff
@@ -23,10 +23,5 @@ public partial class PutScriptDescriptor
 			Assign(a => a.Script = selector?.Invoke(new StoredScriptDescriptor()));
 
 		public PutScriptDescriptor Painless(string source) => Assign(a => a.Script = new PainlessScript(source));
-		public PutScriptDescriptor Groovy(string source) => Assign(a => a.Script = new GroovyScript(source));
-		public PutScriptDescriptor JavaScript(string source) => Assign(a => a.Script = new JavaScriptScript(source));
-		public PutScriptDescriptor Python(string source) => Assign(a => a.Script = new PythonScript(source));
-		public PutScriptDescriptor LuceneExpression(string source) => Assign(a => a.Script = new LuceneExpressionScript(source));
```
Mpdreamz: Keep `LuceneExpression` and `Mustache`

PR #2957, opened 2017-12-15
Include similarity in index settings (#2924)

### LINKED ISSUE #2890: Include Classic similarity in index settings

russcam:
Non binary breaking back port of 648c3cf9d016839bc6aa4dbd60da28d90db984b3

Similarity settings are nested under settings>index
Add ClassicSimilarity
Add DFISimilarity
Add similarity settings to create index integration tests
Add XML comments for each similarity
Refactor SimilarityJsonConverter to not look up the NEST ISimilarity type with Type.GetType() but to infer it from the type property value directly.
Add Similarity to the updatable index settings

Closes #2890


src/Tests/Framework/ManagedElasticsearch/Tasks/InstallationTasks/InstallationTaskBase.cs
```diff
@@ -35,7 +35,7 @@ protected void ExecuteBinary(string binary, string description, params string[]
 			Console.WriteLine($"Preparing to execute: {description} ...");
 			var timeout = TimeSpan.FromSeconds(420);
 			var handle = new XplatManualResetEvent(false);
-			Task.Run(() =>
+			var task = Task.Run(() =>
```
codebrain: :+1:

PR #2959, opened 2017-12-15
Add phrase_limit to highlighting

### LINKED ISSUE #2851: Add phrase_limit for FVH highlighting
[from stackoverflow](https://stackoverflow.com/questions/46162384/specifying-phrase-limit-for-elasticsearch-using-nest-client)

Fast Vector Highlighter should have [`phrase_limit`](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-highlighting.html#phrase-limit)

russcam:
- Change order from string to HighlighterOrder?
- Change TagsSchema from string to HighlighterTagsSchema?
- Remove BoundaryMaxSize
- Change Encoder to HighlighterEncoder?
- Remove encoder from highlight fields as it is only valid on the top level highlighter
- Add Encoder and PhraseLimit to integration test

Requires backporting to 5.x and 2.x in a non-binary breaking way.

Closes #2851

### DIFF
--- a/src/Nest/Search/Search/Highlighting/Highlight.cs
+++ b/src/Nest/Search/Search/Highlighting/Highlight.cs
@@ -54,10 +54,6 @@ namespace Nest
 		[JsonProperty("fragment_offset")]
 		int? FragmentOffset { get; set; }
 
-		[Obsolete("Bad mapping use BoundaryMaxScan instead")]
-		[JsonProperty("boundary_max_size")]
-		int? BoundaryMaxSize { get; set; }
-
 		/// <summary>
 		/// Controls how far to look for boundary characters. Defaults to 20.
 		/// </summary>
@@ -69,13 +65,13 @@ namespace Nest
 		/// It can be either default (no encoding) or html (will escape html, if you use html highlighting tags).
 		/// </summary>
 		[JsonProperty("encoder")]
-		string Encoder { get; set; }
+		HighlighterEncoder? Encoder { get; set; }
 
 		/// <summary>
 		/// The order in which highlighted fragments are sorted
 		/// </summary>
 		[JsonProperty("order")]
-		string Order { get; set; }
+		HighlighterOrder? Order { get; set; }
 
 		/// <summary>
 		/// Use a specific "tag" schemas.
@@ -88,7 +84,7 @@ namespace Nest
 		/// &lt;em class="hlt10"&gt;
 		/// </remarks>
 		[JsonProperty("tags_schema")]
-		string TagsSchema { get; set; }
+		HighlighterTagsSchema? TagsSchema { get; set; }
 
 		[JsonProperty("fields")]
 		[JsonConverter(typeof(VerbatimDictionaryKeysJsonConverter<Field, IHighlightField>))]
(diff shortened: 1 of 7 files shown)


src/Nest/Search/Search/Highlighting/Highlight.cs
```diff
@@ -215,13 +206,13 @@ public class HighlightDescriptor<T> : DescriptorBase<HighlightDescriptor<T> ,IHi
 			);
 
 		// <inheritdoc/>
-		public HighlightDescriptor<T> TagsSchema(string schema = "styled") => Assign(a => a.TagsSchema = schema);
+		public HighlightDescriptor<T> TagsSchema(HighlighterTagsSchema? schema) => Assign(a => a.TagsSchema = schema);
```
codebrain: No default value?

src/Nest/Search/Search/Highlighting/Highlight.cs
```diff
@@ -239,19 +230,17 @@ public class HighlightDescriptor<T> : DescriptorBase<HighlightDescriptor<T> ,IHi
...
-		public HighlightDescriptor<T> Encoder(string encoder) => Assign(a => a.Encoder = encoder);
+		public HighlightDescriptor<T> Encoder(HighlighterEncoder? encoder) => Assign(a => a.Encoder = encoder);
 
 		// <inheritdoc/>
-		public HighlightDescriptor<T> Order(string order) => Assign(a => a.Order = order);
+		public HighlightDescriptor<T> Order(HighlighterOrder? order) => Assign(a => a.Order = order);
 
 		// <inheritdoc/>
-		public HighlightDescriptor<T> RequireFieldMatch(bool requireFieldMatch) => Assign(a => a.RequireFieldMatch = requireFieldMatch);
+		public HighlightDescriptor<T> RequireFieldMatch(bool requireFieldMatch = true) => Assign(a => a.RequireFieldMatch = requireFieldMatch);
```
codebrain: New default of true changes behaviour - is this intentional?

src/Nest/Search/Search/Highlighting/HighlightField.cs
```diff
@@ -238,7 +241,7 @@ public class HighlightFieldDescriptor<T> : DescriptorBase<HighlightFieldDescript
 		public HighlightFieldDescriptor<T> AllField() => this.Field("_all");
 
 		/// <inheritdoc/>
-		public HighlightFieldDescriptor<T> TagsSchema(string schema = "styled") => Assign(a => a.TagsSchema = schema);
+		public HighlightFieldDescriptor<T> TagsSchema(HighlighterTagsSchema? schema) => Assign(a => a.TagsSchema = schema);
```
codebrain: No default?

src/Nest/Search/Search/Highlighting/Highlight.cs
russcam: Yes, it's intentional. Where methods take a `bool` parameter and the calling of the method is a positive intent (i.e. enables something), a default parameter value of `true` should be passed. I thought we had a test for this, but the test is for `bool?`:

https://github.com/elastic/elasticsearch-net/blob/5617675f61c368be30c93e710c9440adeec1289d/src/Tests/CodeStandards/Descriptors.doc.cs#L96

This should be updated to accept `bool?`. I think we also need a sweep before release to ensure we're adhering to this convention.

src/Nest/Search/Search/Highlighting/HighlightField.cs
russcam: I can make it the default, but the client generally doesn't have default values for nullable enums. I suspect this had a default value more as a hint for the accepted string value.

src/Nest/Search/Search/Highlighting/Highlight.cs
russcam: I can make it the default, but the client generally doesn't have default values for nullable enums. I suspect this had a default value more as a hint for the accepted string value.

russcam:
Changed the parameter to a `bool?` with a default value. @codebrain would you mind taking another quick look please?

russcam:
Ported to:
- 5.x: 5632144a04d4bb8025f420787b8e27223be16ea1 
- 2.x: efbb717b3819c5570a20a5d1523bf116798e35e7

PR #2965, opened 2017-12-18
Add Params to PhraseSuggestCollate

### LINKED ISSUE #2849: PhraseSuggestCollate needs a Params property to pass params to the collate query
NEST/ElasticSearch.Net 5.3
Elasticsearch 5.5

When running a collate query on a phrase suggester the json produced and sent to the server looks like this with the params object inside of the query object:

```json
        "collate": {
          "query": {
            "inline": "{\"bool\":{\"must\":[{\"match\":{\"{{field-name}}\":{\"query\":\"{{suggestion}}\",\"type\":\"phrase\"}}}],\"filter\":[{\"nested\":{\"query\":{\"bool\":{\"filter\":[{\"term\":{\"listings.site-ids\":{\"value\":\"{{site-id}}\"}}},{\"term\":{\"listings.region-id\":{\"value\":\"{{region-id}}\"}}},{\"term\":{\"listings.industry-id\":{\"value\":\"{{industry-id}}\"}}}]}},\"path\":\"listings\"}},{\"bool\":{\"must_not\":[{\"exists\":{\"field\":\"private-ids\"}}]}}]}}}",
            "params": {
              "field-name": "name",
              "site-id": "1",
              "region-id": "1",
              "industry-id": "4"
            }
          },
          "prune": true
        }
```

While a working query when setting up the json manually looks like this with the params object inside of the collate object instead: 

```json
        "collate": {
        	"query": {
        		"inline": "{\"bool\":{\"must\":[{\"match\":{\"{{field-name}}\":{\"query\":\"{{suggestion}}\",\"type\":\"phrase\"}}}],\"filter\":[{\"nested\":{\"query\":{\"bool\":{\"filter\":[{\"term\":{\"listings.site-ids\":{\"value\":\"{{site-id}}\"}}},{\"term\":{\"listings.region-id\":{\"value\":\"{{region-id}}\"}}},{\"term\":{\"listings.industry-id\":{\"value\":\"{{industry-id}}\"}}}]}},\"path\":\"listings\"}},{\"bool\":{\"must_not\":[{\"exists\":{\"field\":\"private-ids\"}}]}}]}}"
        	},
        	"params": {
        		"field-name": "name",
        		"site-id": "1",
        		"region-id": "1",
        		"industry-id": "4"
        		},
        	"prune": true
        }
```

Steps to reproduce:
 1. Setup a phrase suggester in json with a collate query that takes parameters
 2. Setup the same phrase suggester in .Net 
 3. Compare the request body of the .Net request to the json from Step 1 and see the params object is at the wrong level so in the .Net request the params are not replaced in the collate query

Adding a Params property to PhraseSuggestCollate fixes this and produces a working request.

russcam:
PhraseSuggestCollate Query takes an ITemplateQuery that allows the specification of parameters. The API of phrase suggester however specifies the parameters for the query at the same level as the query, not nested within the query object. This commit adds a Parameters dictionary to IPhraseSuggestCollate so that parameters are serialized to the correct query form. If parameters are specified on the ITemplateQuery, these are assigned to the parameters on IPhraseSuggestCollate.

Add XML documentation to indicate the usage of each property.

Closes #2849


russcam:
Requires a port to master. This port should probably not use `ITemplateQuery` and instead define a new type that does not expose a `Params` property

PR #2973, opened 2017-12-20
Feature/resolvable dictionaries

Mpdreamz:
This PR adds support for resolvable dictionaries on responses. In the past you would always have to use the resolved form as key e.g

```csharp
response.Indices["project"]["doc"].Properties.Should().NotBeEmpty();
```

Which is a shame if you already have an `IndexName` or `TypeName` or `Field` etc at hand. Now you can use those on the readonly dictionaries that are returned on responses and `IProperties`.

```csharp
response.Indices[Index<Project>()].Mappings[Type<Project>()].Properties.Should().NotBeEmpty();
response.Indices[Index<Project>()][Type<Project>()].Properties.Should().NotBeEmpty();
var properties = response.Indices[Index<Project>()][Type<Project>()].Properties;

var leadDev = properties[Property<Project>(p => p.LeadDeveloper)];
leadDev.Should().NotBeNull();
```


Mpdreamz:
TODO clean up isadictionary constructrors

PR #2976, opened 2017-12-21
Fix/5.x whole time units

### LINKED ISSUE #2818: TimeUnit week ("w") inconsistency
Hello,

In the latest NEST (5.5) and Elasticsearch (5.5):
Trying to use TimeUnit as in the code below:
```c#
.ExponentialDate(gd => gd
     .Field("update_date")
     .Origin(DateMath.Now)
     .Offset(TimeSpan.FromDays(60))
```

Offset expects Time type.
When I specify Offset("60d") above the query works fine, as "d" is supported by Elasticsearch.
When I specify Offset(TimeSpan.FromDays(60)), NEST transforms that to the following JSON:
```javascript
"exp": {
    "update_date": {
    "origin": "now",
    "scale": "8.57142857142857w"
```
and as soon as "w" (week) seems like not supported for TimeUnits the exception is thrown.

Supported TimeUnits:
https://www.elastic.co/guide/en/elasticsearch/reference/5.5/common-options.html#time-units
NEST Time:
https://www.elastic.co/guide/en/elasticsearch/client/net-api/5.x/time-units.html

Please update, thank you!

russcam:
This commit changes the implementation of Time within the client for better consistency with Elasticsearch's TimeValue. The Time type in NEST 5.x is used to represent both Elasticsearch's TimeValue (which allows from nanos up to days as unit), and as time modifiers within DateMath expressions, for which Elasticsearch accepts units of  y (years) M (months), w (weeks), d (days), h/H (hours), m (minutes) and s (seconds).

- When converting from a TimeSpan to Time, the serialized value is a whole number with the largest unit of days. This fixes the implementation to be in line with Elasticsearch 5.0+ with elastic/elasticsearch#19102 where weeks was removed as a supported unit for TimeValue.

- When constructing a Time using the constructor that takes a double of milliseconds, make the implementation consistent with the implicit conversion from double implementation; in the case of -1, both should be Time.MinusOne and the case of 0, both should be Time.Zero. If needing to express -1ms or 0ms, the constructor taking a factor and unit, or the constructor taking a string value with unit should be used.

- When constructing a Time from double or TimeSpan, try to reduce the value to a whole number factor and largest represented unit (up to unit of days). For values smaller than 1 nanosecond, retain the factor as a fraction of nanoseconds.

- When converting a Time to a TimeSpan, round the value to the nearest TimeSpan tick for microseconds and nanoseconds, and milliseconds for all other units.

- When converting to a Time within a DateMath expression using Time.ToFirstUnitYieldingInteger(), when the value is fractional nanoseconds, round to the nearest nanosecond. Although this method is only used within the client for DateMath expressions and can return times with units smaller than a second (the smallest unit that DateMath within Elasticsearch accepts), the method is public so could be used by consumers for purposes other than DateMath. Because of this, don't break the implementation for units smaller than one second.

- Allow the implicit conversion from string and the constructor that takes a string to continue to accept fractional units, which can be returned from Elasticsearch as per TimeValue's toString implementation https://github.com/elastic/elasticsearch/blob/f2501130399f6e3788e110d74dd62b3aad66804a/core/src/main/java/org/elasticsearch/common/unit/TimeValue.java#L253-L290

- Update documentation

Closes #2818

### DIFF
--- a/src/Nest/CommonOptions/TimeUnit/Time.cs
+++ b/src/Nest/CommonOptions/TimeUnit/Time.cs
@@ -45,16 +45,11 @@ namespace Nest
 		private double ApproximateMilliseconds { get; set; }
 
 		public static implicit operator Time(TimeSpan span) => new Time(span);
-
-		public static implicit operator Time(double milliseconds)
-		{
-			if (Math.Abs(milliseconds - (-1)) < FLOAT_TOLERANCE) return MinusOne;
-			if (Math.Abs(milliseconds) < FLOAT_TOLERANCE) return Zero;
-			return new Time(milliseconds);
-		}
+		public static implicit operator Time(double milliseconds) => new Time(milliseconds);
 		public static implicit operator Time(string expression) => new Time(expression);
 
 		public static Time MinusOne { get; } = new Time(-1, true);
+
 		public static Time Zero { get; } = new Time(0, true);
 
 		private Time(int specialFactor, bool specialValue)
@@ -141,6 +136,11 @@ namespace Nest
 			return 1;
 		}
 
+		/// <summary>
+		/// Converts an instance of <see cref="Time"/> with a fractional value to an instance of <see cref="Time"/>
+		/// with a whole value. The largest unit is <see cref="TimeUnit.Day"/>. For fractional values in
+		/// <see cref="TimeUnit.Nanoseconds"/>, value will be rounded to the nearest nanosecond.
+		/// </summary>
 		public static Time ToFirstUnitYieldingInteger(Time fractionalTime)
 		{
 			var fraction = fractionalTime.Factor.GetValueOrDefault(double.Epsilon);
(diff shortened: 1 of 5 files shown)


russcam:
The implementation of `Time` should be updated for 6.x, to split out into two types:

- a type to represent Elasticsearch's `TimeValue`
- a type to represent times within NEST's `DateMath`

src/Nest/CommonOptions/TimeUnit/Time.cs
```diff
@@ -31,23 +34,22 @@ public class Time : IComparable<Time>, IEquatable<Time>
...
 
 		public static implicit operator Time(TimeSpan span) => new Time(span);
 
 		public static implicit operator Time(double milliseconds)
 		{
-			if (Math.Abs(milliseconds - (-1)) < FLOAT_TOLERANCE) return Time.MinusOne;
-			if (Math.Abs(milliseconds) < FLOAT_TOLERANCE) return Time.Zero;
+			if (Math.Abs(milliseconds - (-1)) < FLOAT_TOLERANCE) return MinusOne;
+			if (Math.Abs(milliseconds) < FLOAT_TOLERANCE) return Zero;
 			return new Time(milliseconds);
```
russcam: Can now simply return `new Time(milliseconds);`

Mpdreamz: :+1:

Mpdreamz:
Awesome work @russcam 

>When constructing a Time using the constructor that takes a double of milliseconds, make the implementation consistent with the implicit conversion from double implementation; in the case of -1, both should be Time.MinusOne and the case of 0, both should be Time.Zero. If needing to express -1ms or 0ms, the constructor taking a factor and unit, or the constructor taking a string value with unit should be used.

I remember going to some effort to build it as it was since the constructor mentions passing milliseconds but the way this PR normalizes it makes a lot more sense

PR #2987, opened 2017-12-22
Feature/per request throw exceptions toggle

Mpdreamz:


Mpdreamz:
cc @russcam and @codebrain adding this in while you guys are out ❗️

Issue #3000, opened 2018-01-01
Failing to parse empty geo_centroid


MaJaHa95:
**NEST/Elasticsearch.Net version**: 5.6.0

**Elasticsearch version**: 6.1.1

**Description of the problem including expected versus actual behavior**:

Geo Centroid aggregate fails to parse with no results.

Essentially, NEST is unable to parse responses that look like this:

```
{
  ...
  "hits" : {
    "total" : 0
    ...
  },
  "aggregations" : {
    "centroid" : {
      "count" : 0
    }
  }
}

```

This resulted from a query that returned no results, with a Geo Centroid aggregation applied. But my actual issue first showed up in an empty sub-aggregation on a Date Histogram. Either way, I believe same root cause (albeit different exceptions).

Looking into the code and stack trace, I believe the issue is happening here:

https://github.com/elastic/elasticsearch-net/blob/3bde31718b7371c798b7829f8fc2040e0f5f3a55/src/Nest/Aggregations/AggregateJsonConverter.cs#L354

The `ReadAggregate` function in `AggregateJsonConverter` is looking at the name of the first property in the aggregation result, finding `count`, and incorrectly concluding that it's a Stats Aggregation. When it tries to parse it as a stats aggregation, it's surprised to find that that's the only property, so an exception is thrown.

Essentially, Elasticsearch doesn't include the `centroid` property when there are no records present, which is logical enough. But NEST doesn't parse that response properly.

I believe the issue itself is relatively clear, but looking at this all, I'm still not sure how to resolve it on my end. Because the parsing is occurring at query time, I'm not given the opportunity to stop it when `doc_count` is 0 on the parent aggregation.

As for a PR, the best I can think would be to check for whether the only property is `count`, and parse it as a `GeoCentroidAggregate` if so, with a `null` `Location`. But assuming an aggregate with only that property is inherently a geo_centroid seems error-prone, even if it does happen to be correct for now.

Any suggestions on a workaround in my code? I'm open to ideas. I might go with the arithmetic mean of the results of a `geo_bounds`, since it doesn't have this same limitation. Of course, that's not necessarily the centroid, but it's probably real-world close enough.

For reference, `geo_bounds` works because it does not include the `count` property (or any properties) when the parent is empty, so it just parses as `null`. Convenient, but not helpful for us here.

Mpdreamz:
Opened https://github.com/elastic/elasticsearch-net/pull/3002 to deal with this. 

[...]

Issue #2902, opened 2017-11-15
Composite agg 6.1


Mpdreamz:
https://github.com/elastic/elasticsearch/pull/26800

Might be good to try and implement this in `master` before `6.1` hits to see if we find anything that would be hard to map.

PR #3002, opened 2018-01-02
Fix/geocentroid agg empty

Mpdreamz:
fix for #3000 

It relies on `typed_keys=true` being send to the server so the aggregation names return `type#name` which all search request already do by default in the `6.x` client.

`AggregateDictionaryJsonConverter` now inspects the `type` portion. We only do it for `geo_centroid` for now but it should be easy to start adding all the others over the course of 6.x and support new ones like composite aggs #2902. `typed_keys=true` is a requirement from 6.x onwards. The fact that the old heuristics parser works in 98% of the cases is not good enough. 

Test classes can be marked with `[NeedsTypedKeys]` so that they don't run when the test configuration sets typed keys to false randomly. 

Can be tested with:

```
$  build integrate 6.0.0 "readonly" "aggregation" random:typedkeys
```

```
$  build integrate 6.0.0 "readonly" "aggregation" random:typedkeys:false
```

This PR also fixes an issue where this random value was not fed to the aggregation tests correctly.

### DIFF
--- a/src/Nest/Aggregations/AggregateDictionary.cs
+++ b/src/Nest/Aggregations/AggregateDictionary.cs
@@ -1,68 +1,10 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace Nest
 {
-	internal class AggregateDictionaryConverter : VerbatimDictionaryKeysJsonConverter<string, IAggregate>
-	{
-		private static readonly AggregateJsonConverter OldSchoolHeuristicsParser = new AggregateJsonConverter();
-
-		public override bool CanRead => true;
-
-		public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
-		{
-			var dictionary = new Dictionary<string, IAggregate>();
-			if (reader.TokenType != JsonToken.StartObject)
-			{
-				reader.Skip();
-				return new AggregateDictionary(dictionary);
-			}
-
-			var depth = reader.Depth;
-			while (reader.Depth >= depth)
-			{
-				reader.Read();
-				var typedProperty = reader.Value as string;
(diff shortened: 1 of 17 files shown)


src/Nest/Aggregations/AggregateDictionary.cs
```diff
@@ -1,40 +1,42 @@
...
 	/// Contains aggregates that are returned by Elasticsearch. In NEST `Aggregation` always refers to an aggregation
 	/// going to elasticsearch and an `Aggregate` describes an aggregation going out.
 	/// </summary>
 	[ContractJsonConverter(typeof(AggregateDictionaryConverter))]
 	public class AggregateDictionary : IsAReadOnlyDictionaryBase<string, IAggregate>
 	{
+		private static string TypedKeysTokens(string key)
+		{
+//typed_keys = true on results in aggregation keys being returned as "<type>#<name>"
+			var tokens = key.Split(TypedKeysSeparator, 2, StringSplitOptions.RemoveEmptyEntries);
```
codebrain: var tokens = TypedKeyTokens(key); ?

src/Tests/Framework/Configuration/EnvironmentConfiguration.cs
```diff
@@ -39,13 +39,18 @@ public EnvironmentConfiguration()
 
 			this.Random = new RandomConfiguration
 			{
-				SourceSerializer = BoolConfig("SOURCESERIALIZER", randomizer),
-				TypedKeys = BoolConfig("TYPEDKEYS", randomizer),
+				SourceSerializer = RandomBoolConfig("SOURCESERIALIZER", randomizer),
+				TypedKeys = RandomBoolConfig("TYPEDKEYS", randomizer),
```
codebrain: :+1:

src/Nest/Aggregations/AggregateDictionary.cs
Mpdreamz: whole method can be remove alltogether 👍

PR #3003, opened 2018-01-02
GetMapping helpers to get the TypeMapping for a type.

Mpdreamz:
We used to have a Mapping property which returns the first index's first type's mapping. Since folks typically call get mapping to only return a single mapping. We removed this because its a leaky abstraction but it might be hard for folks to figure out the best way now. Introduced GetMappingFor helpers and reintroduced the Mapping property with an obsolete to guide folks during upgrade.

cc @niemyjski

### DIFF
--- a/src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs
+++ b/src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs
@@ -41,7 +41,7 @@ namespace Nest
 	{
 		[JsonIgnore]
 		public IReadOnlyDictionary<IndexName, IndexMappings> Indices => Self.BackingDictionary;
-		
+
 		[JsonIgnore]
 		[Obsolete("Renamed to Indices, will be deleted from NEST 7.x")]
 		public IReadOnlyDictionary<IndexName, IndexMappings> Mappings => Indices;
@@ -49,14 +49,35 @@ namespace Nest
 		[Obsolete("Use GetMappingFor explicitly instead this is a leaky abstraction that returns the mapping of the first index's first type on the response")]
 		public ITypeMapping Mapping => this.Indices.FirstOrDefault().Value?.Mappings?.FirstOrDefault().Value;
 
-		public ITypeMapping GetMappingFor<T>() => this.Indices[typeof(T)]?[typeof(T)];
-		public ITypeMapping GetMappingFor(string index, string type) => this.Indices[index]?[type];
-		public ITypeMapping GetMappingFor(string index) => this.Indices[index]?.Mappings?.FirstOrDefault().Value;
 
 		public void Accept(IMappingVisitor visitor)
 		{
 			var walker = new MappingWalker(visitor);
 			walker.Accept(this);
 		}
+	}
+
+	public static class GetMappingResponseExtensions
+	{
+
+		public static ITypeMapping GetMappingFor<T>(this IGetMappingResponse response) => response.GetMappingFor(typeof(T), typeof(T));
+
+		public static ITypeMapping GetMappingFor(this IGetMappingResponse response, IndexName index, TypeName type)
+		{
+			if (index.IsNullOrEmpty() || type.IsNullOrEmpty()) return null;
+			TypeMapping mapping = null;
+			var hasValue = response.Indices.TryGetValue(index, out var typeMapping)
+			        && typeMapping?.Mappings != null
+			        && typeMapping.Mappings.TryGetValue(type, out mapping);
+			return mapping;
+		}
+
+		public static ITypeMapping GetMappingFor(this IGetMappingResponse response, IndexName index)
+		{
+			if (index.IsNullOrEmpty()) return null;
+			if (!response.Indices.TryGetValue(index, out var typeMapping)) return null;
+			return typeMapping?.Mappings?.FirstOrDefault().Value;
+		}
+
 	}
 }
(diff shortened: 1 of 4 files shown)


src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs
```diff
@@ -44,6 +44,13 @@ public class GetMappingResponse : DictionaryResponseBase<IndexName, IndexMapping
 		[JsonIgnore]
 		public IReadOnlyDictionary<IndexName, IndexMappings> Mappings => Indices;
 
+		[Obsolete("Use GetMappingFor explicitly instead this is a leaky abstraction that returns the mapping of the first index's first type on the response")]
+		public TypeMapping Mapping => this.Indices.FirstOrDefault().Value?.Mappings?.FirstOrDefault().Value;
+
+		public TypeMapping GetMappingFor<T>() => this.Indices[typeof(T)]?[typeof(T)];
+		public TypeMapping GetMappingFor(string index, string type) => this.Indices[index]?[type];
```
niemyjski: Should any argument validation checks happen here?

src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs
```diff
@@ -44,6 +44,13 @@ public class GetMappingResponse : DictionaryResponseBase<IndexName, IndexMapping
...
+		public TypeMapping GetMappingFor(string index) => this.Indices[index]?.Mappings?.FirstOrDefault().Value;
```
niemyjski: Nice little helper, but I wonder if this should be SingleOrDefault() so it throws if you have more than 1 mapping (to let you know you might get a random mapping back. Also this needs a null check after the FirstOrDefault()?.Value

src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs
```diff
@@ -44,6 +44,13 @@ public class GetMappingResponse : DictionaryResponseBase<IndexName, IndexMapping
 		[JsonIgnore]
 		public IReadOnlyDictionary<IndexName, IndexMappings> Mappings => Indices;
 
+		[Obsolete("Use GetMappingFor explicitly instead this is a leaky abstraction that returns the mapping of the first index's first type on the response")]
+		public TypeMapping Mapping => this.Indices.FirstOrDefault().Value?.Mappings?.FirstOrDefault().Value;
+
+		public TypeMapping GetMappingFor<T>() => this.Indices[typeof(T)]?[typeof(T)];
```
niemyjski: Is there any perf hit to doing a lookup of the same type? I'd assume this would be cached and also a micro optimizations.

src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs
```diff
@@ -44,6 +44,13 @@ public class GetMappingResponse : DictionaryResponseBase<IndexName, IndexMapping
 		[JsonIgnore]
 		public IReadOnlyDictionary<IndexName, IndexMappings> Mappings => Indices;
 
+		[Obsolete("Use GetMappingFor explicitly instead this is a leaky abstraction that returns the mapping of the first index's first type on the response")]
+		public TypeMapping Mapping => this.Indices.FirstOrDefault().Value?.Mappings?.FirstOrDefault().Value;
```
niemyjski: Needs a null check. I'm thinking this probably should be the old behavior and not SingleOrDefault as below.

src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs
niemyjski: On second thought, should this return ITypeMapping?

niemyjski:
I think we assumed that getting a value by the indexer would return null but it's throwing exceptions.

```
[7:05] 
{System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.
  at System.ThrowHelper.ThrowKeyNotFoundException()
  at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
  at Nest.ResolvableDictionaryProxy`2.get_Item(TKey key)
  at Foundatio.Parsers.ElasticQueries.ElasticQueryParserConfiguration.<>c__DisplayClass70_0`1.<UseMappings>b__0() in C:\Code\Foundatio.Parsers\src\Foundatio.Parsers.ElasticQueries\ElasticQueryParserConfiguration.cs:line 360
  at Foundatio.Parsers.ElasticQueries.ElasticQueryParserConfiguration.GetServerMapping() in C:\Code\Foundatio.Parsers\src\Foundatio.Parsers.ElasticQueries\ElasticQueryParserConfiguration.cs:line 329}
```

src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs
Mpdreamz: `typeof(T)` is not a compile time constant but yeah the IL for it is pretty minimal already :)

src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs
Mpdreamz: Multiple mappings is to be expected in this case (and the initial reason for removing the `.Mapping` helper property.

src/Nest/Indices/MappingManagement/GetMapping/GetMappingResponse.cs
Mpdreamz: ++ this could be a whole lot more defensive 👍

Mpdreamz:
Thanks @niemyjski updated so that the indexer is safe to call if they key is missing. Made the methods extension methods on `IGetMappingResponse` instead that return `ITypeMapping` and these methods are a lot more defensive in validation. Also added integration tests for the extension methods and the indexer in the case of key not found.

PR #3009, opened 2018-01-03
Feature/auto routing

Mpdreamz:
This PR strongly types the `routing` parameter to `Routing` and introduces a new feature called auto routing. 

This feature will automatically append `routing=key` if we see the Poco is using a `JoinField` in which case we will send the `parent id` provided there. If the join field itself is the root and only defines the root relation name NEST will send the id of the root object. This plays nicely when `_routing` is required on the mapping. This will hopefully aid people as they move to single types and the new way to do parent and child mappings. Also finished the documentation section for parent and child mappings 6.x style so that we can link to it from the 6.0 GA release blog post.

You can also configure a default routing lookup property on `ConnectionSettings` through `InferMappingFor<>()`. This takes precedence over NEST looking for a `JoinField` property. 

These defaults kick in for request and objects (LikeDocument. BulkOperation, MultiTermVectorOperation, PercolateQuery) that have a routing parameter AND a `.NET` document to infer on. For requests this is automated from the code generator. 

And lastly an explicitly set `Routing(x)` or `Routing = x` by the user takes precedence over all, and if `x == null` it will nullify the routing. 

Important to note: For POCO's that have neither a JoinField or a mapping set up for routing on the `ConnectionSettings` the `routing` parameter is not send automatically so existing code should not be affected.

See docs:

https://github.com/elastic/elasticsearch-net/blob/01b7731dab069e8967d3b3baccda0efa092f800d/src/Tests/ClientConcepts/HighLevel/Mapping/ParentChildJoins.doc.cs

https://github.com/elastic/elasticsearch-net/blob/01b7731dab069e8967d3b3baccda0efa092f800d/src/Tests/ClientConcepts/HighLevel/Inference/RoutingInference.doc.cs


src/CodeGeneration/ApiGenerator/Views/_Descriptors.Generated.cshtml
```diff
@@ -54,11 +55,18 @@ namespace Nest
...
+		var original = kv.Value.OriginalQueryStringParamName;
+		<text>@if (desc.Count == 1) {
+		<text>///<summary>@Raw(desc.First())</summary></text>
+		}
+		else 
+		{ <text>		///<summary>
+		@Raw(string.Join("\r\n\t\t", desc.Select(d=> "/// " + d)))
+		///</summary></text>
+		}</text><text>@if (!string.IsNullOrWhiteSpace(kv.Value.Obsolete))
+		{<text>		[Obsolete("Scheduled to be removed in 7.0, @kv.Value.Obsolete")]</text>}
```
codebrain: 👍

src/CodeGeneration/ApiGenerator/Views/_RequestParametersExtensions.Generated.cshtml
```diff
@@ -21,7 +21,7 @@ namespace Nest
 	{
 	@foreach (CsharpMethod method in Model.CsharpMethodsWithQueryStringInfo)
     {
-        if (!method.Url.Params.Any(p => p.Key.Contains("field") || p.Key.Contains("source_")))
+        if (!method.Url.Params.Any(p => p.Key.Contains("field") || p.Key.Contains("source_") || p.Key.Contains("routing")))
```
codebrain: These chained conditions always scare me as I have to read to work out what they mean.
Much prefer:

var nameOfCompositeCondition = ............................;
if (nameOfCompositeCondition) { ............ }

src/CodeGeneration/ApiGenerator/Views/_Requests.Generated.cshtml
```diff
@@ -80,10 +82,17 @@ namespace Nest
...
+<text>///<summary>@Raw(desc.First())</summary></text>
+		}
+		else 
+		{ <text>///<summary>
+		@Raw(string.Join("\r\n\t\t", desc.Select(d=> "/// " + d)))
+		///</summary></text>
+		}</text><text>@if(!string.IsNullOrWhiteSpace(kv.Value.DeprecatedInFavorOf))
 		{
-		<text>[Obsolete("Scheduled to be removed in 5.0, use @kv.Value.DeprecatedInFavorOf instead")]</text>
+		<text>[Obsolete("Scheduled to be removed in 7.0, use @kv.Value.DeprecatedInFavorOf instead")]</text>
```
codebrain: 👍

src/CodeGeneration/ApiGenerator/Views/_RequestParametersExtensions.Generated.cshtml
Mpdreamz: 👍

PR #3010, opened 2018-01-04
Fix/low level fluent

Mpdreamz:
Request parameters on low level client are no longer fluent. Since we also removed the selector as part of the method signature (the `r => r` bit) folks had to do `new XRequestParameters().Y().Z()` which is not idiomatic `c#`

Removed `RequestParameterExtensions` in the high level client no need for these as you typically set request parameters on the request/descriptor itself. 

cc @niemyjski


niemyjski:
Thanks! LGTM

PR #3012, opened 2018-01-04
[DOCS] Changed certgen to certutil

lcawl:
The certgen tool is deprecated, so the following page should refer to "certutil" instead in 6.1 and later: https://www.elastic.co/guide/en/elasticsearch/client/net-api/master/working-with-certificates.html


Mpdreamz:
Thanks @lcawl 👍

src/Tests/ClientConcepts/Certificates/WorkingWithCertificates.doc.cs
```diff
@@ -176,16 +176,16 @@ protected override void AssertHttpRequestException(HttpRequestException e)
...
 		 *
 		 * X-Pack also allows you to configure a {xpack_current}/pki-realm.html[PKI realm] to enable user authentication
-		 * through client certificates. The `certgen` tool included with X-Pack allows you to
-		 * {ref_current}/certgen.html[generate client certificates as well] and assign the distinguished name (DN) of the
+		 * through client certificates. The {ref_current}/certutil.html[`certutil` tool] included with X-Pack allows you to
+		 * generate client certificates as well and assign the distinguished name (DN) of the
 		 * certificate to a user with a certain role.
 		 *
-		 * certgen by default only generates a public certificate (`.cer`) and a private key `.key`. To authenticate with client certificates, you need to present both
+		 * By default, the `certutil` tool only generates a public certificate (`.cer`) and a private key `.key`. To authenticate with client certificates, you need to present both
```
tvernum: @lcawl  - thanks for the ping.

This is not true for `certutil` - the new tool generates p12 (PKCS#12) by default. (Actually it was out of date for cergen too, since that has an option to generate p12).

 I think these 2 paragraphs needs to be replaced with something like:

>  By default `certutil` generates PKCS#12 files (commonly referred to as p12 or pfx files). The files can be used "as is" by constructing a `new X509Certificate(pathToPfx)` and attaching it to requests.

@Mpdreamz That wording can probably be cleaned up a bit to match the right .NET/Nest terminology.
Also: could someone in your team confirm that `certutil` in ES 6.1 generates p12s that work correctly for NEST client auth? They should, but I'd like to know that the docs cover something we tested.

Issue #3017, opened 2018-01-08
Add 'docvalue_fields' to InnerHits


tsouza:
Since 5.0 Inner hits supports `docvalue_fields` but it seems that, until 5.6, NEST would only support `fielddata_fields` (https://github.com/elastic/elasticsearch-net/blob/5.6.0/src/Nest/Search/Search/InnerHits/InnerHits.cs#L36), which has been renamed to `docvalue_fields` in 5.0.

In master the `fielddata_values` has been removed, but we should add `docvalue_fields`.

PR #3018, opened 2018-01-08
Feature/render partial

Mpdreamz:
Got sidetracked implementing https://github.com/elastic/elasticsearch-net/issues/3017 

It bugged me that sometimes `docvalues_fields` and `stored_field` were going out in the body and sometimes on the query string.

Cleaned up the code generation patching a bit so that we can now declare certain documented querystring parameters as `partial` meaning they will get an automated property on the request interface but the actual implementation has to happen on our manual implementation portion of the request and descriptors. Because it's dictated on the interface though it won't compile. 

This is nicer than our current `Skip` mechanisms which don't dictate that the properties should be manually implemented. 

CodeGenerator now also has a crude warning mechanism when certain properties in overrides are defined but not available in the spec json files. Adding more of these warnings in the future should be straightforward.


src/CodeGeneration/ApiGenerator/Overrides/GlobalOverrides.cs
```diff
@@ -0,0 +1,19 @@
+using System.Collections.Generic;
+using ApiGenerator.Overrides.Descriptors;
+
+namespace ApiGenerator.Overrides
+{
+	public class GlobalOverrides : EndpointOverridesBase
+	{
+		/// <summary>
+		/// Force these be rendered as interface properties only, so that they'd have to be implemented manually
+		/// and become part of the body. This only takes affect for reques
```
codebrain: reques?

Mpdreamz: :+1: addressed.

PR #3022, opened 2018-01-09
Feature/parent deprecation

Mpdreamz:


Mpdreamz:
@russcam @codebrain merged this one in to not have two large PR's linger for to long

PR #3025, opened 2018-01-10
Feature/base class codegen

Mpdreamz:
* generates common parameters to `PlainRequestBase` and `RequestDescriptorBase`
* Cleans up `cshtml` files significantly by consolidating a lot of logic away from the views.
* Output of generated code now a lot more cleaner in terms of spacing.
* Long documentation lines are word wrapped to around `140` characters.
* Cat `h,v,s,ts` parameters are now removed.
* Requests were not listening to `Obsolete` but had their own `DeprecatedInFavorOf` which was never set. So not all `Parent` properties were marked obsolete yet.


Mpdreamz:
@codebrain @russcam merged this in.

PR #3026, opened 2018-01-11
Split out Time into separate types for Time values and DateMath times

### LINKED ISSUE #2983: Split Time out into two types for Time values and Date Math expressions
Relates https://github.com/elastic/elasticsearch-net/pull/2976

russcam:
The Time type has historically been used to represent both Elasticsearch's TimeValue and time within Date Math expressions. This is slightly misleading because

1. TimeValue accepts units from nanos -> seconds
2. Date Math expressions with time accept units from seconds -> years

In both cases, Elasticsearch accepts only whole numbers in addition to a unit of measure.

This commit introduces a new DateMathTime type to represent times within DateMath expressions, with a DateMathTimeUnit enum to represent the valid units for this type. There are implicit conversions from string, TimeSpan and double such that implicit conversion usage in users upgrading from 5.x -> 6.x will not be affected by the change. The constructors that accept string, TimeSpan and double accept fractional numbers to make the type easier to work with, and the fractional number will be converted to the largest whole number and unit that can be represented, with rounding to the nearest second for fractional seconds. An optional rounding argument to determine how to round mid points can be passed to the constructors that accept TimeSpan, double and string.

The DateMathTime type implements equality and comparison members, comparing the seconds representation. An approximation of seconds is used in the case of years and months, with 365 * seconds in a day for years, and (365 / 12)  * seconds in a day for months, allowing for comparisons between months and years.

Remove support for units above a day from the Time type.

Closes #2983

### DIFF
--- a/src/Nest/CommonOptions/DateMath/DateMathOperation.cs
+++ b/src/Nest/CommonOptions/DateMath/DateMathOperation.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Runtime.Serialization;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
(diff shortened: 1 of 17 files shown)


src/Nest/CommonOptions/DateMath/DateMathOperation.cs
```diff
@@ -12,4 +13,22 @@ public enum DateMathOperation
 		[EnumMember(Value = "-")]
 		Subtract
 	}
-}
\ No newline at end of file
+
+	[JsonConverter(typeof(StringEnumConverter))]
```
Mpdreamz: StringEnumConverter?

russcam: good catch

russcam:
Would you mind taking another look @Mpdreamz ?

PR #3034, opened 2018-01-13
TransactionScope and Nest

AliCharper:
Hi,

I need to know is there any way to put Nest Bulk indexing inside a TransactionScope?.

I need to be able to RollBack  if something was wrong.

Tnx


russcam:
@AliCharper I don't think you meant to open this as a PR?

There is no notion of transactions within Elasticsearch and NEST. Each operation within a bulk request is independent to others so, for example, in a bulk indexing operation to index 10 documents, 9 documents may be successfully indexed and one may fail. You can use the `ItemsWithErrors` collection to determine which bulk operation items have failed and take appropriate action.

I'm going to close this PR. I think the better place to discuss this further if you would like is [the Elastic discourse forum](https://discuss.elastic.co/c/elasticsearch).

PR #3035, opened 2018-01-13
Fix fluent for BulkAliasDescriptor.RemoveIndex()

laurentlbm:
RemoveIndex on BulkAliasDescriptor was not fluent, returning instead the interface IBulkAliasRequest.


russcam:
Good catch @laurentlbm , thanks for submitting!

PR #3037, opened 2018-01-15
fix #3017 add docvalues_fields to innerhits

### LINKED ISSUE #3017: Add 'docvalue_fields' to InnerHits

Mpdreamz:
Needs backport to `5.x`


Mpdreamz:
backported to 5.x via #3045

PR #3038, opened 2018-01-15
fix #3019, converter should not nullify know keys in the base dictionary

### LINKED ISSUE #3019: UpdateIndexSettingsRequest does not work on adding well-known settings through Dictionary.Add() method
**NEST/Elasticsearch.Net version**: 5.6.0

**Elasticsearch version**: 5.4.0

I'm trying to write a generic code for updating index setting where the setting name and value are provided as inputs. If I initialize an object of `IndexSettings` class and add a setting using the `Dictionary.Add()` method for a well-known setting, then Nest throws saying `"no settings to update"`.

**Steps to reproduce**:
```
var settings = new IndexSettings();
settings.Add("index.unassigned.node_left.delayed_timeout", "10m");
var request = new UpdateIndexSettingsRequest(Indices.AllIndices)
{
    IndexSettings = settings
};
client.UpdateIndexSettings(request);

```

I understand that using `settings.UnassignedNodeLeftDelayedTimeout = "10m";` works but when I do not know what the setting is beforehand, I cannot use this. I also understand that I can achieve this using the low level client but I would like to avoid that as much as possible.

My ask is to allow configuring any setting by adding to a dictionary irrespective of whether it is a well-known setting or not.

Mpdreamz:


src/Nest/IndexModules/IndexSettings/Settings/IndexSettingsConverter.cs
```diff
@@ -20,76 +21,81 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
-			d[UpdatableIndexSettings.BlocksWrite] = ds.BlocksWrite;
-			d[UpdatableIndexSettings.BlocksMetadata] = ds.BlocksMetadata;
-			d[UpdatableIndexSettings.Priority] = ds.Priority;
-			d[UpdatableIndexSettings.RecoveryInitialShards] = ds.RecoveryInitialShards;
-			d[UpdatableIndexSettings.RequestsCacheEnable] = ds.RequestsCacheEnabled;
-			d[UpdatableIndexSettings.RoutingAllocationTotalShardsPerNode] = ds.RoutingAllocationTotalShardsPerNode;
-			d[UpdatableIndexSettings.UnassignedNodeLeftDelayedTimeout] = ds.UnassignedNodeLeftDelayedTimeout;
+			void Set(string knownKey, object newValue)
+			{
+				if (newValue != null) d[knownKey] = newValue;
```
russcam: Might you want to set a null value too?

Mpdreamz: That would always override what is set in the backing dictionary which is what triggered #3019

PR #3039, opened 2018-01-15
Automatically skip parameters already found in _common.json but still…

Mpdreamz:
… warn about them

This fixes compiler warnings because `Human` is generated on the base class and implementation.


src/CodeGeneration/ApiGenerator/ApiGenerator.cs
```diff
@@ -71,13 +71,17 @@ private static RestApiSpec CreateRestApiSpecModel(string downloadBranch, string[
 					.Where(f => f.EndsWith(".json") && !IgnoredApis.Contains(new FileInfo(f).Name))
 					.ToList()
 				);
+				var commonFile = Path.Combine(CodeConfiguration.RestSpecificationFolder, "Core", "_common") + ".json";
```
russcam: combine "_common" and ".json" in the `Path.Combine()` call?

PR #3043, opened 2018-01-15
Fix/ttl timestamp

Mpdreamz:


Mpdreamz:
We should, I opened an issue for these for the time being though.

PR #3044, opened 2018-01-15
Fix/obsolete and removed

Mpdreamz:
Wen't over the breaking changes docs one last time


src/Nest/QueryDsl/Query.cs
```diff
@@ -114,11 +110,11 @@ public static class Query<T> where T : class
...
 
-		public static QueryContainer Percolate(Func<PercolateQueryDescriptor<T>, IPercolateQuery> selector) =>
-			new QueryContainerDescriptor<T>().Percolate(selector);
-
-		public static QueryContainer Prefix(Expression<Func<T, object>> fieldDescriptor, string value, double? boost = null, MultiTermQueryRewrite rewrite = null, string name = null) =>
-			new QueryContainerDescriptor<T>().Prefix(fieldDescriptor, value, boost, rewrite, name);
+		public static QueryContainer Percolate(Func<PercolateQueryDescriptor<T>, IPercolateQuery> selector) =>
+			new QueryContainerDescriptor<T>().Percolate(selector);
+
+		public static QueryContainer Prefix(Expression<Func<T, object>> fieldDescriptor, string value, double? boost = null, MultiTermQueryRewrite rewrite = null, string name = null) =>
```
russcam: Wondering if we should remove all the shorthands that accept `Field`, `Expression<Func<T,object>>` et. al, except for on `Term()` and `Terms()`?

src/Nest/QueryDsl/Compound/Indices/IndicesQuery.cs
```diff
@@ -1,74 +0,0 @@
...
-using System.Collections.Generic;
-using System.Linq;
-using Newtonsoft.Json;
-
-namespace Nest
-{
-	[JsonObject(MemberSerialization = MemberSerialization.OptIn)]
-	[JsonConverter(typeof(ReadAsTypeJsonConverter<IndicesQueryDescriptor<object>>))]
-	[Obsolete("Deprecated. You can specify _index on the query to target specific indices")]
-	public interface IIndicesQuery : IQuery
```
russcam: I think this is a good candidate for an example in the documentation i.e. combining a query on `_index` with other queries to target a specific index  in a request that targets multiple indices

src/Nest/QueryDsl/Query.cs
Mpdreamz: Or atleast get rid of the nullable parameters. I think we should do this for `7.x` at this stage though

PR #3048, opened 2018-01-16
Fix/nullable descriptor arguments

Mpdreamz:
Fun one, catches all value type arguments on descriptors are nullable. Also validates nullable bool arguments on descriptors have a default value and that default value is `true`.

### DIFF
--- a/src/CodeGeneration/ApiGenerator/Domain/ApiQueryParameters.cs
+++ b/src/CodeGeneration/ApiGenerator/Domain/ApiQueryParameters.cs
@@ -46,7 +46,8 @@ namespace ApiGenerator.Domain
 		}
 
 		public string ClsArgumentName => this.ClsName.ToCamelCase();
-		public string DescriptorArgumentType => this.Type == "list" && this.TypeHighLevel.EndsWith("[]") ? "params " + this.TypeHighLevel : TypeHighLevel;
+		public string DescriptorArgumentType =>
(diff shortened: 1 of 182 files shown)


codebrain:
There are a number of interfaces that have changed as a result of this:

`IExtendedStatsAggregation` 
`IDelimitedPayloadTokenFilter`
`IClusterAllocationExplainRequest`
`IAllocateClusterRerouteCommand` 
`IAllocateReplicaClusterRerouteCommand` 
`IAllocateStalePrimaryRerouteCommand` 
`ICancelClusterRerouteCommand` 
`IMoveClusterRerouteCommand`
`IRolloverConditions`
`IDateIndexNameProcessor` 
`IFieldNamesField` 
`IGeoSuggestContext 
`ILinearInterpolationSmoothingModel`
`IStopDatafeedRequest`
`IPutRoleMappingRequest`
`IGetUserAccessTokenRequest`

Taking for example, `IGetUserAccessTokenRequest` - the property change...

`[JsonProperty("grant_type")]`
`AccessTokenGrantType GrantType { get; set; }`

...to now:

`[JsonProperty("grant_type")]`
`AccessTokenGrantType**?** GrantType { get; set; }`

The issue is that the receiving method will not accept a `null` (only a `string` of "password"), further compounded by body parameters not being part of the REST specification.

What do we want to do in these situations?

Mpdreamz:
Discussed with @codebrain, we default to `null` every where and rely on the server to tell us a property is required. This might change when we look at generating our types but for 6.x uniformity wins.

PR #3053, opened 2018-01-18
add support for DefaultType on ConnectionSettings, removed PluralizeT…

Mpdreamz:
…ypeNames()


src/Nest/CommonAbstractions/ConnectionSettings/ConnectionSettingsBase.cs
```diff
@@ -134,10 +125,14 @@ public TConnectionSettings DefaultIndex(string defaultIndex)
 			return (TConnectionSettings) this;
 		}
 
-		private string LowerCaseAndPluralizeTypeNameInferrer(Type type)
+		/// <summary>
+		/// A default Elasticsearch type name to use for CLR types. If you also set <see cref="DefaultTypeNameInferrer"/> this will only take affect if your callback
```
russcam: Took me more than a casual glance to understand what this comment was saying (_it is Monday morning though!_); I think it could be a little clearer. Maybe something like
```
Sets a default type name to use within Elasticsearch for all CLR types. If <see cref="DefaultTypeNameInferrer"/> is also set, a configured default type name will only be used when <see cref="DefaultTypeNameInferrer"/>returns null or empty. If unset, the default type name for types will be the lowercased CLR type name.
```
?

src/Nest/CommonAbstractions/ConnectionSettings/Inflector.cs
```diff
@@ -1,220 +0,0 @@
...
-using System.Collections.Concurrent;
-using System.Collections.Generic;
-using System.Text.RegularExpressions;
-
-namespace Nest
-{
-	/// <summary>
-	/// Pluralizes or singularizes words.
-	/// </summary>
-	public static class Inflector
```
russcam: 🎉 🎈

PR #3056, opened 2018-01-18
Fix/covariant search results

Mpdreamz:
Goodbye old friend, you will be missed!

SInce `types` are gone it makes no sense to keep this feature around, I love the built in covariant feature but on the upside this simplifies things a great deal. 

We can now also easily mark generic interfaces with a default generic type to deserialize to (see IHit).


src/Tests/ClientConcepts/HighLevel/CovariantHits/CovariantSearchResults.doc.cs
```diff
@@ -10,246 +10,49 @@ public class CovariantSearchResults
...
 		/**=== Covariant search results
 		 *
-		 * NEST directly supports returning covariant result sets.
-		 * Meaning a result can be typed to an interface or base class
-		 * but the actual instance type of the result can be that of the subclass directly
+		 * NEST used to have a feature that allowed you to map multiple types in an index back into a covariant list.
 		 *
-		 * Let's look at an example; Imagine we want to search over multiple types that all implement
-		 * `ISearchResult`
+		 * Since types are on the way in Elasticsearch itself this feature is no longer supported. Because you can
```
russcam:
```
Since types are removed in Elasticsearch 6.0
```
I think is a little clearer, maybe with a link to the removal of mapping types post/documentation?

PR #3060, opened 2018-01-23
Feature/custom implementation of interfaces

### LINKED ISSUE #3049: icu_collation_keyword property not supported
**NEST/Elasticsearch.Net version**: v6.0.0-rc1 

**Elasticsearch version**: 6.1.1

We need to use the icu_collation_keyword property type for our mappings. In the previous version of NEST, we used a custom implementation of `IProperty` (code below). However, since upgrading the package to v6.0.0-rc1, the properties marked with the `JsonProperty` don't get serialized anymore.

Expected behavior: either an already existing implementation of the `icu_collation_keyword` or the proper serialization of properties marked with `JsonProperty`.

**Steps to reproduce**:
 1. Use the below implementation of a property
 2. Field is mapped as "type": "icu_collation_keyword", but all other properties (language and so on) are missing

```
        public class IcuCollationKeywordProperty : IProperty
	{
		public IcuCollationKeywordProperty(string name, string language)
		{
			this.Name = new PropertyName(name);
			this.Type = "icu_collation_keyword";
			this.Language = language;
			this.Variant = "@collation=standard";
			this.Strength = "primary";
			this.Numeric = true;
		}

		public IcuCollationKeywordProperty(string language)
			: this(null, language)
		{
		}

		public PropertyName Name { get; set; }

		public IDictionary<string, object> LocalMetadata { get; set; }

		[JsonProperty("language")]

		[JsonProperty("strength")]
		public string Strength { get; set; }

		[JsonProperty("variant")]
		public string Variant { get; set; }

		[JsonProperty("numeric")]
		public bool Numeric { get; set; }
	}
```

Mpdreamz:
Renamed `RenameAttribute` to `PropertyNameAttribute`, added tests that custom `IProperty` implementations serialize to fix #3049


docs/client-concepts/certificates/working-with-certificates.asciidoc
```diff
@@ -80,9 +80,8 @@ public class AllowAllCertificatesCluster : SslAndKpiXPackCluster
 If your client application has access to the public CA certificate locally, Elasticsearch.NET and NEST ship with some handy helpers
 that can assert that a certificate the server presents is one that came from the local CA.
 
-If you use X-Pack's {ref_current}/certutil.html[+certutil+ tool] to generate SSL certificates, the generated node certificate
-does not include the CA in the certificate chain, in order to cut down on SSL handshake size. In those case you can use
-`CertificateValidations.AuthorityIsRoot` and pass it your local copy of the CA public key to assert that
+If you use X-Pack's {ref_current}/certutil.html`certutil` tool] to generate SSL certificates, the generated node certificate
```
russcam: This change will break the documentation; it needs to be fixed in https://github.com/russcam/asciidocnet. Should revert this change

docs/client-concepts/certificates/working-with-certificates.asciidoc
```diff
@@ -116,7 +115,7 @@ the local CA certificate is part of the chain that was used to generate the serv
 ==== Client Certificates
 
 X-Pack also allows you to configure a {xpack_current}/pki-realm.html[PKI realm] to enable user authentication
-through client certificates. The {ref_current}/certutil.html[+certutil+ tool] included with X-Pack allows you to
+through client certificates. The {ref_current}/certutil.html`certutil` tool] included with X-Pack allows you to
```
russcam: This change will break the documentation; it needs to be fixed in https://github.com/russcam/asciidocnet. Should revert this change

src/Tests/ClientConcepts/HighLevel/Serialization/ExtendingNestTypes.doc.cs
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		[U] public void InjectACustomIPropertyImplementation()
+		{
+			/**
+			 * `PropertyNameAttribute` can be used to mark properties that should be serialized. Without this attribute NEST won't pick up the property for serialization.
+			 *
+			 * Now that we have our own `IProperty` implementation we can add it to our propertes mapping when creating an index
+			 */
+			var descriptor = new CreateIndexDescriptor("myindex")
```
russcam: I think we should try to move away from using `CreateIndexDescriptor` and serializing directly in documentation examples, and instead make an actual client call using an `InMemoryConnection`, capture the serialized request and assert against this. I think it makes the usage clearer to people reading the docs

PR #3065, opened 2018-01-29
Update Parent/Child relationship documentation

russcam:
Remove the recommendation to always set the _routing field as required. It is only needed for the child mapping, not the parent


russcam:
This is not an issue 😃

PR #3068, opened 2018-01-31
fix on empty buckets

Zyklop:
fixed loop so the deserialisation doesn't crash when reciving en empty aggregation bucket.


psytraxx:
We experienced this problem when we were dealing with  empty aggregation bucket.  Changing the do-while loop into a while loop solved the issues for us. We are unsure if the same problem applies also for other branches but we would be very greatful if this fix makes it into the next releases of 5.x of  since we are currently depending on a custom compiled DLL in our production code.

russcam:
Thanks @Zyklop 👍

Zyklop:
@russcam We need this fix in Version 5.5. Could you please publish a new nuget package with this fix?

PR #3071, opened 2018-02-02
Feature/clean deps

Mpdreamz:
This moves master over to `netstandard1.3` only for our class libs.

Tests and other executables are now `netstandard2.0` only. 

Removes all the `DOTNETCORE` related ifdef's.

For now we keep the `WebRequest` based `HttpConnection` connection around in a separate package `Elasticsearch.Net.Connections.WebRequestConnection`


src/CodeGeneration/DocGenerator/Program.cs
```diff
@@ -10,15 +10,17 @@ static Program()
 			var currentDirectory = new DirectoryInfo(Directory.GetCurrentDirectory());
             if (currentDirectory.Name == "DocGenerator" && currentDirectory.Parent.Name == "CodeGeneration")
 			{
+				Console.WriteLine("IDE: " + currentDirectory);
```
russcam: Still needed? Or just used for debugging?

src/CodeGeneration/DocGenerator/Program.cs
```diff
@@ -10,15 +10,17 @@ static Program()
...
                 InputDirPath = @"..\..\";
 				OutputDirPath = @"..\..\..\docs";
                 BuildOutputPath = @"..\..\..\build\output";
 			}
 			else
 			{
-				InputDirPath = @"..\..\..\..\..\src";
-				OutputDirPath = @"..\..\..\..\..\docs";
-                BuildOutputPath = @"..\..\..\..\..\build\output";
+				Console.WriteLine("CMD: " + currentDirectory);
```
russcam: Still needed? Or just used for debugging?

src/Nest/CrossPlatform/SynchronizedCollection.cs
```diff
@@ -1,87 +1,71 @@
...
-using System;
-using System.Collections;
-using System.Diagnostics;
-using Nest;
+﻿using Nest;
 
 namespace System.Collections.Generic
 {
 
+	//TODO see if we can get rid of this
```
russcam: ++

Mpdreamz:
A lot of changes went in today too:

* `HttpWebRequestConnection` is now a nuget package as well that is part of the command line build
* `HttpWebRequestConnection` is randomly tested on windows when targetting `net46`
* `Elasticsearch.DotNetCore.sln` is gone, no longer a need for this :tada: :cake:
* `Profiling` is now loaded optionally, this removes the requirement of calling `build` once before building in the IDE. If you call `test(.exe) profile` it will exit with an error + error code when not compiled with profiling support
* `DocGenerator` can now run on non windows platforms.
* `canary` nuget package testing only happens on appveyor due to package caches 
* `DocGenerator` was always looking for the xml docs in `build/output` but if you do `git clean -fxd` and run from the IDE this folder does not exist.

ghavaun:
Is there any possible way 2fix this?? Would realy appreciate some asistants
please

PR #3075, opened 2018-02-05
Re-add Aggregation usage examples

russcam:
This PR re-adds aggregation usage examples back into the documentation. Refactoring as part of 6.0 introduced aggregation related fluent, initializer and expectjson properties which are not picked up by the Usages Roslyn walker/visitor.

### DIFF
--- a/docs/client-concepts/certificates/working-with-certificates.asciidoc
+++ b/docs/client-concepts/certificates/working-with-certificates.asciidoc
@@ -80,7 +80,7 @@ public class AllowAllCertificatesCluster : SslAndKpiXPackCluster
 If your client application has access to the public CA certificate locally, Elasticsearch.NET and NEST ship with some handy helpers
 that can assert that a certificate the server presents is one that came from the local CA.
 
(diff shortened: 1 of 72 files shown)


docs/client-concepts/certificates/working-with-certificates.asciidoc
```diff
@@ -115,7 +115,7 @@ the local CA certificate is part of the chain that was used to generate the serv
 ==== Client Certificates
 
 X-Pack also allows you to configure a {xpack_current}/pki-realm.html[PKI realm] to enable user authentication
-through client certificates. The {ref_current}/certutil.html[+certutil+ tool] included with X-Pack allows you to
```
Mpdreamz: Is this the correct way ? We broke the doc build on this before not sure if this ends up reverting it?

russcam: That link looks fine there, but I'll double check. If I get a chance today, I'll take a look at AsciiDocNet

PR #3099, opened 2018-02-19
Fix/contentbody verb

BastiaanLouis:
When sending a HEAD or GET message, the message does not contain any content. The used HttpClient class accepts no content at all using these type of verbs. The current implementation that creates an empty BytesArrayContent actually creates a message that will not be send by the HttpClient. A ProtocolViolationException will be thrown (see https://github.com/Microsoft/referencesource/blob/master/System/net/System/Net/HttpWebRequest.cs, private void CheckProtocol(bool onRequestStream)).
This small fixes leaves the Content of the requestMessage null, only for HEAD and GET messages for which no postdata is offered.

Unfortunately there are no tests available that I can update to validate the change. To verify the bug and it's fix, I've been using the StaticConnectionPool (actually through Serilog).

### DIFF
--- a/src/Elasticsearch.Net/Connection/HttpConnection-CoreFx.cs
+++ b/src/Elasticsearch.Net/Connection/HttpConnection-CoreFx.cs
@@ -241,14 +241,16 @@ namespace Elasticsearch.Net
 					data.Write(stream, requestData.ConnectionSettings);
 				stream.Position = 0;
 			}
-			else
+			// HttpClient does not allow any Content for GET and HEAD requests
+			else if ((requestMessage.Method != System.Net.Http.HttpMethod.Get) && (requestMessage.Method != System.Net.Http.HttpMethod.Head))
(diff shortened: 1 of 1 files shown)


karmi:
Hi @BastiaanLouis, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/3099.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

BastiaanLouis:
Will create a new pull request

karmi:
@BastiaanLouis, it's perfectly to just use `git commit --amend --reset-author` with a new config and force push the branch! Just so you know that you don't have to open a completely new PR in future. Thanks!

BastiaanLouis:
Thx, learned something today :)

PR #3108, opened 2018-02-23
Fixed grammar issues

acidshards:


russcam:
Thanks for the PR @acidshards! Would it be possible to edit the original source file at `https://github.com/elastic/elasticsearch-net/tree/6.x/src/Tests/ClientConcepts/LowLevel/GettingStarted.doc.cs` from which the documentation is generated? Also, have you [signed the CLA](https://www.elastic.co/contributor-agreement)?

karmi:
Hi @acidshards, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/3108.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

acidshards:
Hi @russcam, no problem, I will edit the source file after the weekend. 
Hi @karmi, I deleted the e-mail used in my Git commit from Github because I couldn't access it any more. Maybe we should close this pull request without merging and I will create a new one together with the changes to the source file?

russcam:
Sounds good to me @acidshards , feel free to close

PR #3111, opened 2018-02-27
Remove user from hipchat watcher config

russcam:
The validation for watcher settings became stricter in 6.2. As a result, the user setting in hipchat configuration that is not used now causes the bootstrap checks to fail. Remove the setting.


Mpdreamz:
ported to `6.x`

PR #3117, opened 2018-03-01
fix #3116 instances of gethashcode in equals implementation

### LINKED ISSUE #3116: Incorrect usages of GetHashCode?
Considering the incorrect usage of GetHashCode in issue #2881, I decided to look through the code to see if there may be similar issues elsewhere. And I might have found a few.

I believe using GetHashCode() to determine equality is wrong, or at least bad practice. Two unequal objects may return the same hash code, depending on the GetHashCode() implementation.

This is from Indices.Equals:

```
		public override bool Equals(object obj)
		{
			var other = obj as Indices;
			if (other == null) return false;
			return this.Match(
				all => other.Match(a => true, m => false),
				many => other.Match(
					a => false,
					m => this.GetHashCode().Equals(other.GetHashCode())
				)
			);
		}

```

Similar code in:

Types.Equals
IndexName.EqualsMarker
RelationName.EqualsMarker
TypeName.EqualsMarker

Thoughts on this?

Mpdreamz:


src/Nest/CommonAbstractions/Infer/Types/Types.cs
```diff
@@ -88,17 +88,24 @@ string IUrlParameter.GetString(IConnectionConfigurationValues settings)
...
 				)
 			);
 		}
 
+		private static bool EqualsAllTypes(IReadOnlyList<TypeName> indicesCurrent, IReadOnlyList<TypeName> indicesOther)
+		{
+			if (indicesCurrent == null && indicesOther == null) return true;
+			if (indicesCurrent == null || indicesOther == null) return false;
+			if (indicesCurrent.Count != indicesOther.Count) return false;
+			return indicesCurrent.Zip(indicesOther, Tuple.Create).All(t=>t.Item1.Equals(t.Item2));
```
russcam: I think this should also take into account the order

Mpdreamz:
All good points @russcam, addressed them in followup commit.

Mpdreamz:
backported to `5.x` and 6.x`

PR #3121, opened 2018-03-06
Set Body on low level response for BytesResponse

russcam:
This commit fixes a bug that was introduced when moving to typed
ElasticsearchResponse<T> responses from the low level client,
whereby the response.Body was not being set in the case of BytesResponse.


src/Elasticsearch.Net/Transport/Pipeline/ResponseBuilder.cs
```diff
@@ -146,7 +146,7 @@ private static bool SetSpecialTypes<TResponse>(byte[] bytes, out TResponse cs)
 
 			if (responseType == typeof(StringResponse))
 				cs = new StringResponse(bytes.Utf8String()) as TResponse;
-			else if (responseType == typeof(byte[]))
+			else if (responseType == typeof(BytesResponse))
```
Mpdreamz: Doh!!

PR #3126, opened 2018-03-13
Structural equals for infer and IUrlParameter types

Mpdreamz:
All `IUrlParameter` and `Infer` types now implement a structural `Equals()` and not a referential `Equals()`

* Trim() is bad on singular types, removed this from them as it only hides bugs. Moved the trim to plural type parsing of multi valued comma separated strings.
* Made sure all implementations implement `IEquatable<>`
* Made sure implicit conversion from null/empty string/empty comma separated string all behave the same (return null) for IUrlParameters and Infer types.
* Made sure null flows through all the Equals implementations and fixed existing tests that failed due to the updated implementation

`Field` now takes boost value into account, this was previously explicitly documented as desired behaviour but I can remember why this was the case, @russcam  any ideas?

### DIFF
--- a/src/Nest/CommonAbstractions/Extensions/Extensions.cs
+++ b/src/Nest/CommonAbstractions/Extensions/Extensions.cs
@@ -183,6 +183,13 @@ namespace Nest
 			return list != null && list.Any();
 		}
 
+		internal static bool IsEmpty<T>(this IEnumerable<T> list)
+		{
+			if (list == null) return true;
+			var enumerable = list as T[] ?? list.ToArray();
+			return !enumerable.Any() || enumerable.All(t => t == null);
+		}
+
 		internal static void ThrowIfNull<T>(this T value, string name, string message = null)
 		{
 			if (value == null && message.IsNullOrEmpty()) throw new ArgumentNullException(name);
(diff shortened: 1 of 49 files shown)


src/Nest/CommonAbstractions/Extensions/Extensions.cs
```diff
@@ -193,6 +200,16 @@ internal static bool IsNullOrEmpty(this string value)
 		{
 			return string.IsNullOrWhiteSpace(value);
 		}
+		internal static bool IsNullOrEmptyCommaSeparatedList(this string value, out string[] split)
```
codebrain: Use a `.All(...)` LINQ extension? might save initialising the array.

codebrain: ...PEBKAC, just noticed the `out` parameter!

src/Nest/CommonAbstractions/Infer/IndexName/IndexName.cs
```diff
@@ -19,24 +19,25 @@ public class IndexName : IEquatable<IndexName>, IUrlParameter
...
 			this.Name = index;
 			this.Type = type;
-			this.Cluster = cluster?.Trim();
+			this.Cluster = cluster;
 		}
 
 		public static IndexName From<T>() => typeof(T);
 		public static IndexName From<T>(string clusterName) => From(typeof(T), clusterName);
 		private static IndexName From(Type t, string clusterName) => new IndexName(t, clusterName);
+		// TODO private?
```
codebrain: TODO?

Mpdreamz: left it in there to tackle at a later stage (not related to commit).

Mpdreamz:
Yeah this stuff should come baked *cough* `F#` *cough`.

Mpdreamz:
Ammended to not take `boost` into account on `Field` this would cause fields with different boost values to be cached separately which is undesirable. Conceptually i can live with `field^1 == field^3` too.

russcam:
> Field now takes boost value into account, this was previously explicitly documented as desired behaviour but I can remember why this was the case, @russcam any ideas?

IIRC, it was to do with field expression caching. If memory serves, the original expression overload did not accept a boost in the early 2.x days, and was added later.

PR #3127, opened 2018-03-13
Fix #3115 bump json.net to version 11

### LINKED ISSUE #3115: Add support for JSON.NET 11.0 for NEST 5/6 clients
Is there any timeline or plans for upgrading JSON.NET to 11.0 for the current clients?

Mpdreamz:
This also removes the +1 major exclusive restriction on the dependency range.

I am OK with folks upgrading at their own risk given that we internalize our own `JSON.NET` now.


Mpdreamz:
@codebrain @russcam merging this in so that TC can do integration tests.

Mpdreamz:
ported to `6.x`

PR #3133, opened 2018-03-15
when 'dotnet xunit' returns a non zero exit code forcefully fail the …

Mpdreamz:
…build, in cases where our xunit setup fails no tests will be run and will end up going unreported

cc @codebrain @russcam


Mpdreamz:
ported to `6.x` and `5.x`

PR #3139, opened 2018-03-20
Remove multiple types per index from documentation

### LINKED ISSUE #3138: Outdated Attribute Mapping Doc Page
This page' example is outdated. 

https://www.elastic.co/guide/en/elasticsearch/client/net-api/current/attribute-mapping.html

1. The [Object] attribute doesn't support Store=false
2. It uses ElasticsearchType attribute with two type names which is not supported in ES6x

russcam:
Closes #3138


russcam:
Requires documentation to be regenerated in 6.x after merging

PR #3188, opened 2018-04-05
Fix synchronous HEAD request

VanHuychemG:
When making a HEAD requests the response does not contain a ContentType. The Async version of `Request` already has this check, however the synchronous call fails with a NullReferenceException.


russcam:
Thanks for the PR @VanHuychemG 👍

CumpsD:
Thanks!

PR #3195, opened 2018-04-11
Change Fields to be optional on MultiMatchQuery

### LINKED ISSUE #3159: Make fields optional on multi_match query
[Rely on index.query.default_field by default when no fields specified](https://github.com/elastic/elasticsearch/pull/27380)

Remove Fields from the conditionless check

https://github.com/elastic/elasticsearch-net/blob/23ae899d42cc0dc55e8068d905fc8125f3d1ca6b/src/Nest/QueryDsl/FullText/MultiMatch/MultiMatchQuery.cs#L97

russcam:
This commit adds the ability to send multi_match query without providing any fields.
When no fields are provided the multi_match query will use the fields
defined in the index setting index.query.default_field
(which in turns defaults to *).

Closes #3159


codebrain:
Ported to `6.x`

PR #3201, opened 2018-04-14
Generate release notes

russcam:
This commit adds the generation of draft release notes to build/output.

Closed issues for labelled with the target release label are fetched from
Github and grouped by Bug,Feature and Deprecation labels, appearing in
more than one group if featuring more than one of those labels.

Comparison link is generated from the version in global.json and the
version being released.


Mpdreamz:
:gandalf:

PR #3212, opened 2018-04-25
Fix #3210 , missing decision enum values

### LINKED ISSUE #3210: Json Serialization Exception on Decision Enum
When calling the Elastic Search Client method:
           client.ClusterAllocationExplainAsync(ed => ed
                .ErrorTrace()
                .Human()
                .Index("my_index")
                .Primary(true));

I am getting a Json Serialization Exception trying to Deserialize the following json snip found in part of the response.

"node_allocation_decisions": [{
    "node_id": "_3APDoyWQUGJC2J_LbxQVw",
    "node_name": "node",
    "transport_address": "10.10.10.10:9300",
    "node_attributes": {
	    "ml.max_open_jobs": "10",
	    "rack_id": "2",
	    "ml.enabled": "true",
	    "machinetype": "warm"
    },
    "node_decision": "worse_balance",
    "weight_ranking": 4
	}
]

The NodeAllocationExplanation.NodeDecision property is of type Nullable<Decision>
The Decision enum values only include yes and no.
This is causing the json deserialization to fail.
Is there a way we can get the possible values added to the enum?
Is there a workaround we could use without having to manually interrogate the raw json? 

ES version 5.6.5
NEST package 5.5.0

Thanks in advance

narph:
based on https://github.com/elastic/elasticsearch/blob/5.6/core/src/main/java/org/elasticsearch/cluster/routing/allocation/AllocationDecision.java

### DIFF
--- a/src/Nest/Cluster/ClusterAllocationExplain/ClusterAllocationExplainResponse.cs
+++ b/src/Nest/Cluster/ClusterAllocationExplain/ClusterAllocationExplainResponse.cs
@@ -251,7 +251,25 @@ namespace Nest
 		Yes,
 
 		[EnumMember(Value = "no")]
-		No
+		No,
+
+		[EnumMember(Value = "worse_balance")]
+		WorseBalance,
+
+		[EnumMember(Value = "throttled")]
+		Throttled,
+
+		[EnumMember(Value = "awaiting_info")]
+		AwaitingInfo,
+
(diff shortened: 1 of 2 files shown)


russcam:
Thanks for the PR @narph! 

I'm going to slightly change the test to use the test client fixed response.

russcam:
squash merged in 2f285b23e70351ae6aa6f9d3098b78ea87a287a7

Issue #3213, opened 2018-04-26
Refactor geo shapes and geo shape queries


russcam:
Geo shapes require some refactoring for 7.x:

1. `IGeometryCollection` should implement `IGeoShape`
2. Remove `IgnoreUnmapped` from `IGeoShape` as it exists at the same level as `_name` and `boost` on the query.
3. Refactor `GeoShape*` queries into a single quey that accepts `IGeoShape`

PR #3214, opened 2018-04-26
Obsolete IgnoreUnmapped on IGeoShape

russcam:
This commit deprecates IgnoreUnmapped on IGeoShape by applying Obsolete attribute.
IgnoreUnmapped exists at the same level as _name and boost on a geo_shape query.

Map IgnoreUnmapped from IGeoShape.IgnoreUnmapped to IGeoShapeQuery.IgnoreUnmapped if
assigned.


russcam:
When ported to master, `Obsolete` attributes can be removed as part of refactoring in #3213

PR #3218, opened 2018-04-27
Add Visit methods for *_range properties

russcam:
This commit adds mapping and property visitor methods for
the range data types.

Map a QueryContainer property as a percolator data type by default.


russcam:
Approved by @codebrain in chat

PR #3223, opened 2018-04-30
Implement IUrlParameter on Time (#3222)

### LINKED ISSUE #3220: Scroll property in SearchRequest throws a System.InvalidCastException
**NEST/Elasticsearch.Net version**: 5.6.0

**Elasticsearch version**: 5.6.0

**Description of the problem including expected versus actual behavior**: Scroll property in SearchRequest throws a System.InvalidCastException when trying to retrieve it. In GetQueryStringValue<TOut>() in FluentRequestParameters.cs, it is trying to explicitly cast the Scroll value (a string) into a Time object. Time only has implicit casts in it right now, so this throws an exception.

**Steps to reproduce**:
 1. Create a SearchRequest with the Scroll property set
 2. Try to read this property
 3. Observe that it throws an exception when trying to evaluate it

russcam:
* Implement IUrlParameter on Time

This commit implements IUrlParameter on Time so that the conversion to a
string value happens when building the url.

Add condition to UrlFormatProvider to convert TimeSpan to Time unit
string value when building the url.

Remove special case handling for Time and TimeSpan and re-run API generation

Fixes #3220


russcam:
> LGTM, Resharper created the switch statement refactoring?

I made a couple of small changes and it did

PR #3264, opened 2018-05-28
Closes #3251. Rename levenstein to levenshtein.

### LINKED ISSUE #3251: [6.2] Deprecating `levenstein` in favor of `levensHtein`
https://www.elastic.co/guide/en/elasticsearch/reference/current/release-notes-6.2.0.html

Suggesters: Deprecating `levenstein` in favor of `levensHtein` - https://github.com/elastic/elasticsearch/pull/27409

codebrain:
Unsure if we should rename the enumeration value at this stage as it would break binary compatibility.

Potentially consider implementing another enumeration value with the correct spelling? @russcam - thoughts?

### DIFF
--- a/src/Nest/Search/Suggesters/TermSuggester/StringDistance.cs
+++ b/src/Nest/Search/Suggesters/TermSuggester/StringDistance.cs
@@ -16,7 +16,7 @@ namespace Nest
 		Internal,
 		[EnumMember(Value = "damerau_levenshtein")]
 		DamerauLevenshtein,
-		[EnumMember(Value = "levenstein")]
+		[EnumMember(Value = "levenshtein")]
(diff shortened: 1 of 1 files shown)


codebrain:
Ported to `6.x`

russcam:
@codebrain this can't be backported to 6.x because e.g. Elasticsearch 6.0 uses `levenstein`

https://github.com/elastic/elasticsearch/blob/6002b41b5f8d6ccfc5b5569ee00958ca565bb7d2/core/src/main/java/org/elasticsearch/search/suggest/term/TermSuggestionBuilder.java#L586

Mpdreamz:
Think this was merged to soon, should still rename the enum in `master` IMO.

PR #3285, opened 2018-06-12
add editorconfig formatting rules (that are supported by rider/reshar…

Mpdreamz:
…per 2018.1

WIP have not fixed all isues found yet but opening this for discussion 😄


Mpdreamz:
It's only a `warning` for now, lets discuss this once during the next sync and then never talk about it again 🗡 💥 💣 😸

PR #3291, opened 2018-06-14
Check stream is seekable when checking length

russcam:
This commit checks that the response stream is seekable
before checking the length; Streams that are not seekable
will throw.


russcam:
ported to 6.x

PR #3292, opened 2018-06-15
Add DocValueFields to TopHits Aggregation

### LINKED ISSUE #3274: TopHits aggregation is missing DocValueFields
Currently, Nest only supports DocValueFields in `InnerHits`, however, this is also a valid argument for `TopHits` aggregations. However, I see that in Nest there is the option to specify FieldDataFields which DocValueFields will ultimately make use of if available. FieldDataFields is currently not listed as a supported per hit feature in the documentation. https://www.elastic.co/guide/en/elasticsearch/reference/6.2/search-aggregations-metrics-top-hits-aggregation.html

Should this be renamed or is there a reason for using FieldDataFields?

russcam:
closes #3274


russcam:
ported to master and 5.x

PR #3293, opened 2018-06-15
Add _ScriptedUpsert to BulkUpdateOperation

### LINKED ISSUE #3270: ScriptedUpsert has no effect on BulkUpdateOperation
**NEST/Elasticsearch.Net version**: 6.1.0

**Elasticsearch version**: 6.0.0

Serialization of `BulkUpdateOperation<>` does not appear to include `scripted_upsert` value in serialized form.

**Steps to reproduce**:

 1. Create a (collection of) `BulkUpdateOperation<object, object>` object, with `Id`, `Index`, `Type`, `Upsert`, and `Script` set to applicable values, and `ScriptedUpsert = true`.
 2. Execute something like...

```
ElasticClient ec = ...;
IList<IBulkOperation> operations = ...; //including from above
await ec.BulkAsync(new BulkRequest { Operations = operations }); 
```
 3. Fiddler shows that there's no accompanying value for `scripted_upsert` in the request body.

Seems like a pretty straight-forward issue; enough that I'm surprised if I'm the first to catch it, albeit a bit of a specialized case. I played around with leaving `Upsert` out and such, but that had no effect on this field.

Are there other prerequisites or something that make the serializer ignore this field?

russcam:
Closes #3270

### DIFF
--- a/src/Nest/Document/Multiple/Bulk/BulkOperation/BulkUpdate.cs
+++ b/src/Nest/Document/Multiple/Bulk/BulkOperation/BulkUpdate.cs
@@ -92,7 +92,8 @@ namespace Nest
 			_PartialUpdate = this.Doc,
 			_Script = this.Script,
 			_Upsert = this.Upsert,
-			_DocAsUpsert = this.DocAsUpsert
+			_DocAsUpsert = this.DocAsUpsert,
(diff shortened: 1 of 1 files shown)


russcam:
ported to 5.x and master

PR #3299, opened 2018-06-20
Long range query support

annashlyak:
Hello, everybody,

I want to make range queries that correctly work with long type. So, I have added support of long range query. 

Small test to demonstrate trouble that I want to solve .
We want to work with timestamps that are long type:
```
q.Range(c => c
	.Name("named_query")
	.Boost(1.1)
	.Field(p => p.Description)
	.GreaterThan(636634079999999999)
	.GreaterThanOrEquals(636634080000000000)
	.LessThan(636634080000000000)
	.LessThanOrEquals(636634079999999999)
	.Relation(RangeRelation.Within))
```
converts numbers to double type:
```
   query = new {
          range = new {
            description = new {
              _name = "named_query",
              boost = 1.1,
              gt = 6.3663408E+17,
              gte = 6.3663408E+17,
              lt = 6.3663408E+17,
              lte = 6.3663408E+17,
              relation = "within"
            }
          }
        }
```
and loses precision.
Elasticsearch himself is able to make such requests, but in NEST this functionality is missed.

I will be grateful if you merge my pull request and I'm ready to discuss my decision.

### DIFF
--- a/src/Nest/QueryDsl/TermLevel/Range/RangeQueryJsonConverter.cs
+++ b/src/Nest/QueryDsl/TermLevel/Range/RangeQueryJsonConverter.cs
@@ -37,34 +37,28 @@ namespace Nest
 
 		private static IRangeQuery GetRangeQuery(JsonSerializer serializer, JObject jo)
 		{
-			var nameIsValid = !jo.Properties().Any(p => p.Name == "format" || p.Name == "time_zone");
+			IRangeQuery fq = FromJson.ReadAs<DateRangeQuery>(jo.CreateReader(), serializer);
+			var isNumeric = false;
+			var isLong = false;
 
-			var isLong = nameIsValid && CheckType(jo, JTokenType.Integer);
-			var isNumeric = nameIsValid && CheckType(jo, JTokenType.Float);
-
-			IRangeQuery fq;
-
+			foreach (var property in jo.Properties())
+			{
+				if (property.Name == "format" || property.Name == "time_zone")
+					return fq;
+				if (_rangeKeys.Contains(property.Name))
+				{
+					if (property.Value.Type == JTokenType.Float)
+						isNumeric = true;
+					else if (property.Value.Type == JTokenType.Integer)
+						isLong = true;
+				}
+			}
 			if (isNumeric)
-			{
-				fq = FromJson.ReadAs<NumericRangeQuery>(jo.CreateReader(), serializer);
-			}
-			else if (isLong)
-			{
-				fq = FromJson.ReadAs<LongRangeQuery>(jo.CreateReader(), serializer);
-			}
-			else
-			{
-				fq = FromJson.ReadAs<DateRangeQuery>(jo.CreateReader(), serializer);
-			}
+				return FromJson.ReadAs<NumericRangeQuery>(jo.CreateReader(), serializer);
+			if (isLong)
+				return FromJson.ReadAs<LongRangeQuery>(jo.CreateReader(), serializer);
 
 			return fq;
-		}
-
-		private static bool CheckType(JObject jo, JTokenType jTokenType)
-		{
-			return jo.Properties().Any(p =>
-				       _rangeKeys.Contains(p.Name) &&
-				       p.Value.Type == jTokenType);
 		}
 
 		private static TReturn GetPropValue<TReturn>(JObject jObject, string field)
--- a/src/Nest/QueryDsl/TermLevel/Range/RangeQueryJsonConverter.cs
+++ b/src/Nest/QueryDsl/TermLevel/Range/RangeQueryJsonConverter.cs
@@ -37,14 +37,13 @@ namespace Nest
 
 		private static IRangeQuery GetRangeQuery(JsonSerializer serializer, JObject jo)
 		{
-			IRangeQuery fq = FromJson.ReadAs<DateRangeQuery>(jo.CreateReader(), serializer);
 			var isNumeric = false;
 			var isLong = false;
 
 			foreach (var property in jo.Properties())
 			{
 				if (property.Name == "format" || property.Name == "time_zone")
-					return fq;
+					return FromJson.ReadAs<DateRangeQuery>(jo.CreateReader(), serializer);
 				if (_rangeKeys.Contains(property.Name))
 				{
 					if (property.Value.Type == JTokenType.Float)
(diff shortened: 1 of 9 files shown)


karmi:
Hi @annashlyak, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/3299.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

annashlyak:
Hi @karmi, I have added email to my profile

Mpdreamz:
Thanks for the PR @annashlyak 😄 this LGTM.

src/Nest/QueryDsl/TermLevel/Range/RangeQueryJsonConverter.cs
```diff
@@ -26,25 +26,40 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
+
+			return fq;
+		}
+
+		private static IRangeQuery GetRangeQuery(JsonSerializer serializer, JObject jo)
+		{
+			var isLong = !jo.Properties().Any(p => p.Name == "format" || p.Name == "time_zone")
+			             && jo.Properties().Any(p => _rangeKeys.Contains(p.Name) && p.Value.Type == JTokenType.Integer);
 
+			var isNumeric = !jo.Properties().Any(p => p.Name == "format" || p.Name == "time_zone")
```
russcam: is it possible to refactor this to remove the duplicated code within `isLong` and `isNumeric` checks. It may be possible to reduce down to one enumeration/iteration of `jo.Properties` to determine `IRangeQuery`

src/Nest/QueryDsl/Visitor/DslPrettyPrintVisitor.cs
```diff
@@ -79,7 +79,9 @@ private void Write(string queryType, Field field = null)
 
 		public virtual void Visit(INumericRangeQuery query) => Write("numeric_range");
 
-        public virtual void Visit(ITermRangeQuery query) => Write("term_range");
+		public virtual void Visit(ILongRangeQuery query) => Write("numeric_range");
```
russcam: `long_range`?

russcam:
would you like me to take another look at the PR @annashlyak - is it ready to review again?

annashlyak:
@russcam  yes, you can watch review again

src/Nest/QueryDsl/TermLevel/Range/RangeQueryJsonConverter.cs
```diff
@@ -26,28 +26,47 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
 
+			return fq;
+		}
+
+		private static IRangeQuery GetRangeQuery(JsonSerializer serializer, JObject jo)
+		{
+			var nameIsValid = !jo.Properties().Any(p => p.Name == "format" || p.Name == "time_zone");
+
+			var isLong = nameIsValid && CheckType(jo, JTokenType.Integer);
+			var isNumeric = nameIsValid && CheckType(jo, JTokenType.Float);
```
russcam: I think this can be simplified further. 

For `isNumeric` and DateRangeQuery cases, `CheckType` will end up being called twice. It looks like it should be possible to enumerate/iterate over `jo.Properties()` once and determine from the existence of "format", "time_zone" or token type, which `IRangeQuery` it is. I was thinking something like

```c#
private static IRangeQuery GetRangeQuery(JsonSerializer serializer, JObject jo)
{
	foreach (var property in jo.Properties())
	{
		if (property.Name == "format" || property.Name == "time_zone")
			return FromJson.ReadAs<DateRangeQuery>(jo.CreateReader(), serializer);

		if (_rangeKeys.Contains(property.Name))
		{
			if (property.Value.Type == JTokenType.Integer)
				return FromJson.ReadAs<LongRangeQuery>(jo.CreateReader(), serializer);					
			if (property.Value.Type == JTokenType.Float)
				return FromJson.ReadAs<NumericRangeQuery>(jo.CreateReader(), serializer);
			
			return FromJson.ReadAs<DateRangeQuery>(jo.CreateReader(), serializer);
		}
	}
}
```

what do you think?

annashlyak: Yes, thank you. LGTM. "format" and "time_zone" checks must be done for all properties in jo. When we check for a type, we can not immediately return the result, because LongRange is used, if there was no property of type float, but there was a property of type int. However, we can iterate once and I think that the code will looks better.

src/Nest/QueryDsl/TermLevel/Range/RangeQueryJsonConverter.cs
```diff
@@ -37,36 +37,30 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
 
 		private static IRangeQuery GetRangeQuery(JsonSerializer serializer, JObject jo)
 		{
-			var nameIsValid = !jo.Properties().Any(p => p.Name == "format" || p.Name == "time_zone");
+			IRangeQuery fq = FromJson.ReadAs<DateRangeQuery>(jo.CreateReader(), serializer);
```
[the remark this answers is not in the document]
annashlyak: Earlier we deserialize to DateRangeQuery always when check for numeric Type was false. I think that we can just not declare fq.

src/Nest/QueryDsl/TermLevel/Range/RangeQueryJsonConverter.cs
```diff
@@ -37,36 +37,30 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
-			}
-			else
-			{
-				fq = FromJson.ReadAs<DateRangeQuery>(jo.CreateReader(), serializer);
+				if (property.Name == "format" || property.Name == "time_zone")
+					return fq;
+				if (_rangeKeys.Contains(property.Name))
+				{
+					if (property.Value.Type == JTokenType.Float)
+						isNumeric = true;
```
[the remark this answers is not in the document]
annashlyak: Earlier was check !jo.Properties().Any(p => p.Name == "format" || p.Name == "time_zone") It is the same as jo.Properties().All(p => p.Name != "format" && p.Name != "time_zone"). So, we have to done this for all properties and I dont want return here NumericRangeQuery

russcam:
This looks great, @annashlyak! Thank you for your effort 👍

PR #3302, opened 2018-06-22
net46 sets WebException as OriginalException where netcore sets Elast…

Mpdreamz:
…icsearchClientException, becuase its IConnection throws and that WebException is now preserved with the backport of #3266


src/Elasticsearch.Net/Transport/Transport.cs
```diff
@@ -3,6 +3,7 @@
 using System.Threading;
 using System;
 using System.Linq;
+using System.Net;
```
Mpdreamz: pragma as well

codebrain:
Forward ported to `master`

PR #3304, opened 2018-06-25
Honour DataMember and DataContract attributes

### LINKED ISSUE #3107: "DataMember" attribute is not applied when deserializing searching query
**NEST/Elasticsearch.Net version**: 6.0.1

**Elasticsearch version**: 6.0.1

I have a document model as below:

    [DataContract(Name = "source_entity")]
    public class SourceEntity
    {
        [DataMember(Name = "name")]
        public string Name { get; set; }
        
        [DataMember(Name = "display_name")]
        public string DisplayName { get; set; }
    }

And here is the code for inserting random source entities:

    public class AddEntitySample
    {
        private ElasticClient elasticClient;

        public AddEntitySample()
        {
            var node = new Uri("http://localhost:9200");
            var pool = new SingleNodeConnectionPool(new Uri("http://localhost:9200"));
            var settings = new ConnectionSettings(pool);
            
            settings.DefaultMappingFor<SourceEntity>(t => t.IndexName("source_entity_fun"));
            settings.DisableDirectStreaming();
            settings.OnRequestCompleted((t) =>
            {
                var a = t.DebugInformation;
            });
            elasticClient = new ElasticClient(settings);
        }

        public void AddSampleData()
        {
            elasticClient.DeleteIndex(Indices.Index<SourceEntity>());
            var enumberation = Enumerable.Range(0, 1000).Select(index => new SourceEntity
            {
                Name = "Name " + index,
                DisplayName = "DisplayName" + index
            });        
            var response = elasticClient.BulkAsync(t => t.IndexMany(enumberation), 
                new System.Threading.CancellationToken()).Result;
        }        
    }

After insert some items in elastic db, I use this command "GET source_entity_fun/_mapping" (in Kibana dev console) to get the schemas of SourceEntity in elastic db.
And here is the json format of the document

    {
      "source_entity_test": {
        "mappings": {
          "sourceentity": {
            "properties": {
              "display_name": {
                "type": "text",
                "fields": {
                  "keyword": {
                    "type": "keyword",
                    "ignore_above": 256
                  }
                }
              },
              "name": {
                "type": "text",
                "fields": {
                  "keyword": {
                    "type": "keyword",
                    "ignore_above": 256
                  }
                }
              }
            }
          }
        }
      }
    }

When I make a simple query:

    var query = elasticClient.Search<SourceEntity>(t => t.Query(u => u.Terms(k => k.Field(h => h.DisplayName).Terms("23"))));
    
And here is the json query generated (using DebugInformation):

    {"query":{"terms":{"displayName":["23"]}}}

As the result, there are no documents returned.
I would expected is should be :

    {"query":{"terms":{"display_name":["23"]}}}
 (with underscore)

I tried to use [NEST.JsonNetSerializer](https://www.nuget.org/packages/NEST.JsonNetSerializer/) but it did not work.
Could you help me, please?
Thanks.

russcam:
This commit updates the behaviour of TypeNameResolver and
PropertyMapping to respect the usage of DataMember and
DataContract attributes for all requests.

Prior behaviour to this commit was that indexing would honour
DataMember attributes on a document due to the internal implementation of
JSON.Net honouring them. With other requests such as Search however,
where a Field type is used, the the resolution of the Field string name did
not honour the attributes.

Closes #3107

### DIFF
--- a/docs/client-concepts/high-level/inference/field-inference.asciidoc
+++ b/docs/client-concepts/high-level/inference/field-inference.asciidoc
@@ -114,7 +114,8 @@ Expect("name")
 To ease creating a `Field` instance from expressions, there is a static `Infer` class you can use
 
 [TIP]
-This example uses the https://msdn.microsoft.com/en-us/library/sf0df423.aspx#Anchor_0[static import] `using static Nest.Infer;` in the using directives to shorthand `Nest.Infer.Field<T>()`
+This example uses the https://msdn.microsoft.com/en-us/library/sf0df423.aspx#Anchor_0[static import] `using static Nest.Infer;`
+in the using directives to shorthand `Nest.Infer.Field<T>()`
 to simply `Field<T>()`. Be sure to include this static import if copying any of these examples.
 
 [source,csharp]
@@ -322,6 +323,7 @@ Expect("leadDeveloper.firstName.raw.evendeeper").WhenSerializing(multiSuffixFiel
 Expect("metadata.hardcoded.raw.evendeeper").WhenSerializing(multiSuffixFieldExpressions[4]);
 ----
 
+[[field-name-attribute]]
 ==== Attribute based naming
 
 Using NEST's property attributes you can specify a new name for the properties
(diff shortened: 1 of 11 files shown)


D9001:
I'm using .NET Core 2.2 with NEST 7.0.1, and whenever I have a class with DataContract and DataMember on a nested property, the objects in the nested property array are created as `[{},{}]`  (if I have 3 properties, there would be 3 empty objects inside). When I take DataContract and DataMember attributes off, everything is serialized just fine.. Is this a known issue?
 
Is there any way around this? I'd like to avoid creating extra classes with the same properties but different attributes.

russcam:
@D9001 It's best to open a separate issue about this and link to this PR; a comment is likely to end up being lost as it's not as visible as an issue. Please open an issue an provide as much detail as you can, ideally with a reproducible example, so that it helps understand more quickly.

niemyjski:
We have our own json.net serializer defined in NEST 7.5.1 and we are doing our own naming strategy that ignores DataMember Attributes (on purpose) and we are hitting our break points as expected. But we noticed that the models that are being serialized that have the data member name are preserved even though our naming strategy is hit.. Any idea why this would be happening in NEST 7, was working as expected in NEST 5.

PR #3333, opened 2018-07-16
build cluster command

Mpdreamz:
Can now start a test cluster from the root of the project by calling
`build cluster <clustername> <version>`

This will build and run the specified cluster from a temporary location.
Allowing us to develop and build under src/Test while it's running.

When running tests `Elastic.Xunit` now supports allowing it to assume a
cluster is already started. With a cluster already running, running the
integration tests from the IDE will be more responsive as we do not have
to wait for the whole cluster setup and teardown.

This brings back the final missing piece of functionality we had before
breaking `Elastic.Xunit` into its own project


Mpdreamz:
ported to `master` and `5.x`

PR #3345, opened 2018-07-31
Fix #3303 Cat Response error should bubble out

### LINKED ISSUE #3303: Expose IndexNotFoundException on CatIndices()
**NEST/Elasticsearch.Net version**: v6.0

**Elasticsearch version**: v6.2.4 running on Elastic Cloud - Amazon (EU Ireland)

When executing a ElasticClient.CatIndices() request on an a cluster which contains no indices (or no indices matching the specified pattern) AND that is protected with X-Pack security (for example an index hosted in Elastic Cloud) the NEST API throws an exception:

> Nest.Json.JsonSerializationException: Cannot deserialize the current JSON object (e.g. {"name":"value"}) into type 'System.Collections.Generic.IReadOnlyCollection`1[Nest.CatIndicesRecord]' because the type requires a JSON array (e.g. [1,2,3]) to deserialize correctly.

This is due to the fact that Elastic Cloud will return an IndexNotFoundException and not an Array:

> Elasticsearch clusters with X-Pack security enabled apply the /_all wildcard, and all other wildcards, to the indices that the current user has privileges for, not the set of all indices on the cluster. While creating or retrieving aliases by providing wildcard expressions for alias names, if there are no existing authorized aliases that match the wildcard expression provided an IndexNotFoundException is returned.

(Source: https://www.elastic.co/guide/en/elastic-stack-overview/current/security-limitations.html#_changes_in_index_wildcard_behavior)

I wouldn't it be better to expose the IndexNotFoundException?
No it's impossible to detect why the method throws the exception. Is it because there is a network issue? Or is it because there are no indices found? Would be cool if we could do something like this:

> try
> {
> _client.CatIndices();
> }
> catch(IndexNotFoundException e){
> // Index does not exist
> }
> catch(Exception e){
> // Possible network issue.
> }

Mpdreamz:
Right now we always attempt to deserialize to list of records but when a
server side error is returned we have can't deserialize the array as as
an object so we need to special case this in the converter callback

### DIFF
--- a/src/Nest/Cat/ElasticClient-Cat.cs
+++ b/src/Nest/Cat/ElasticClient-Cat.cs
@@ -13,8 +13,14 @@ namespace Nest
 		private CatResponse<TCatRecord> DeserializeCatResponse<TCatRecord>(IApiCallDetails response, Stream stream)
 			where TCatRecord : ICatRecord
 		{
+			var catResponse = new CatResponse<TCatRecord>();
+
+			if (!response.Success) return catResponse;
+
 			var records = this.RequestResponseSerializer.Deserialize<IReadOnlyCollection<TCatRecord>>(stream);
-			return new CatResponse<TCatRecord> { Records = records };
+			catResponse.Records = records;
+
+			return catResponse;
 		}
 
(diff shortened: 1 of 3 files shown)


Mpdreamz:
ported to `master` and `5.x`

PR #3352, opened 2018-08-06
Refactor/cut up tests projects

Mpdreamz:
This PR is a tad ridiculous in size but unavoidable with the amount of 🧀 this PR moves around.

`Tests` has always been our one stop artifact to do a variety of development processes:

* Documentation based tests
* Reproduce tests
* Profiling (automated though jetbrains self host API)
* Benchmarking (both micro and macro through Benchmark.NET)
* Unit tests
* Integration tests 
* Being able to start `ClusterBase` implementations used in tests in isolation.

Keeping all of these in one project stopped serving us well a long time ago. What prompted me to do this now though is that Rider 2018.2 will get DotCover support and will be able to automatically run impacted tests on build. This workflow has been long missed since we no longer use VS to develop for the past two years and no longer have NCrunch in our arsenal.  Incremental compiles of `Tests` using `netcoreapp1.1` were `40s` not exactly ideal when continuous testing.

This PR moves everything to `netstandard2.0` and `netcoreapp2.1`. Cutting the `Tests` project into smaller pieces really helps too.

This PR cuts the single `Tests` project up in to:

* `Tests.Configuration`
* `Tests.Domain`
* `Tests.Core`

and the following separate tools:
* `Tests`
Our trusted unit and integration tests that also act as our documentation through literate programming and lots of roslyn magic 🚀 
* `Tests.Benchmarking`
Macro benchmarks using a live cluster using BenchmarkDotNet. Indexes the results in Elasticsearch by the way 😎 
* `Tests.ClusterLauncher`
A tool that can start clusters declared in tests. By calling `build cluster <clustername> <version>` you can start a cluster in isolation in `%TEMP%` and continue to compile the solution. Furthermore our test framework sees the cluster and skips starting a cluster each test run. This really helps speed the development cycle.
* `Tests.Profiling`
A tool that can profile any code and output Performance/Memory/Timeline profile dumps using the dotTrace self host API.
* `Tests.Reproduce`
A tests project solely for reproducing Github/StackOverflow issues. Hopefully this makes it easier to discover for others as well.
* `Tests.ScratchPad`
Micro benchmarking (BenchmarkDotNet) and profiling (through attaching a profiler (http://www.getcodetrack.com/ rules)). This is more used for quick exploratory use.

**Refactor**

Buried in here is also a pretty substantial refactor in how our tests do json serialization testing.

`SerializationTestBase` was a horrible beast and of course the base class for everything. It relied on subclasses implementing virtual properties methods to assert a single json output. Lots of virtual calls in constructors 😢. Very hard to follow. On top of that most of our lovely json output verification, comparison, pretty diffing routines were married into this class. 

`RoundTripper` was a recently new construct that allowed us to break free from `SerializationTestBase` by using static imports one could write e.g

`Expect(jsonString).DeserializesTo<SearchResponse>()`

Super handy however `Expect` could be anything e.g:  

`Expect(new SearchRequest()).ToSerializeTo(jsonString)`

This uni-direction was super hard to follow inside `RoundTripper` since its all boxed as `object`

`RoundTripper` is now much cleaner and seperated in `JsonRoundTripper` and `ObjectRoundTripper`

`SerializationTester` is now a single class that takes care of our json serialization testing routing e.g

* are two unsorted json string equal.
* Does this object serialize to Y and does Y deserialize back to Y?

Both `SerializationTestBase` and `RoundTripper` offload to `SerializationTester`. `SerializationTestBase` has been renamed to `ExpectJsonTestBase` and does very little now.

### TestClient

https://github.com/elastic/elasticsearch-net/blob/57c44dba240e48f2110a61e59fc91bdd662e810d/src/Tests/Framework/TestClient.cs

now only holds the static default clients for tests to pick if they don't need their own `ConnectionSettings`. All other responsibilities have been moved away or refactored to no longer being needed.


Mpdreamz:
Travis and Appveyor are happy, merging this 👹

PR #3358, opened 2018-08-12
Fix #3330 Add additional translog properties as per elastic/elasticsearch#28613…

### LINKED ISSUE #3330: Add Translog files age to TranslogStats
Relates: https://github.com/elastic/elasticsearch/pull/28613

At the same time, review the properties exposed on `TranslogStats` to ensure that they align with those exposed in Elasticsearch.

Mpdreamz:
… and add properties we weren't mapping

Implements #3330

### DIFF
--- a/src/Nest/CommonOptions/Stats/TranslogStats.cs
+++ b/src/Nest/CommonOptions/Stats/TranslogStats.cs
@@ -9,7 +9,7 @@ namespace Nest
 		public long Operations  { get; set; }
 
 		[JsonProperty("uncommitted_operations")]
-		public long UncommittedOperations  { get; set; }
+		public int UncommittedOperations  { get; set; }
 
 		[JsonProperty("size")]
 		public string Size { get; set; }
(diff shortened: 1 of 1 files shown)


src/Nest/CommonOptions/Stats/TranslogStats.cs
```diff
@@ -8,11 +8,23 @@ public class TranslogStats
...
 		public long SizeInBytes  { get; set; }
 
+		[JsonProperty("uncommitted_size")]
+		public string UncommittedSize { get; set; }
+
+		[JsonProperty("uncommitted_size_in_bytes")]
+		public long UncommittedSizeInBytes  { get; set; }
+
+		[JsonProperty("earliest_last_modified_age")]
+		public long EarliestLastModifiedAge  { get; set; }
```
russcam: Maybe a comment to indicate that this is valid only for Elasticsearch 6.3.0+? (or maybe `long?`)

src/Nest/CommonOptions/Stats/TranslogStats.cs
```diff
@@ -8,11 +8,23 @@ public class TranslogStats
 		[JsonProperty("operations")]
 		public long Operations  { get; set; }
 
+		[JsonProperty("uncommitted_operations")]
+		public long UncommittedOperations  { get; set; }
```
russcam: slight nit: Should this be an `int`?

https://github.com/elastic/elasticsearch/blob/6.3/server/src/main/java/org/elasticsearch/index/translog/TranslogStats.java#L37

PR #3381, opened 2018-08-27
Bump to latest Elastic.Xunit and Proc

Mpdreamz:
Clusters can now quite easily provide an expected shutdown period.

[Proc](https://github.com/proc-net/proc) waits for the streamreaders to emit EOL after the process has exitted and if this does not happen before a timeout OnError on the observable is called.

In Elastic.Xunit we call `elasticsearch.bat` (host_ which spawns `elasticsearch.jar` (child) and so `ElasticsearchNode` tracks both process id's, however [Proc](https://github.com/proc-net/proc) tracks only the host. 

So while [Proc](https://github.com/proc-net/proc) was shutting down the host OK as that triggers the read wait for EOL on the console out streams these pipes (coming from `elasticsearch.jar`) can stay active quite a while. When we give up the jar is kept running (in shutdown mode).

This manifested itself later on when `EphemeralClusterComposer` wanted to run its clean up routines and the ephemeral directories were still in use.

NodeConfiguration can now provide `WaitForShutDown` which will set `Proc`'s `WaitForExit` and `WaitForStreamReaders`. `ClusterBase` also provided an overridable `ModifyNodeConfiguration` so that this can be set per cluster with relatively low ceremony.

This `WaitForShutdown` does not have to be accurate if its too short `ElasticsearchNode` now does a hard kill on both the host and and the java pid in a new hook provided by `ObservableProcessBase` `OnBeforeWaitForEndOfStreamsError`.

### DIFF
--- a/src/Tests/Tests.Configuration/tests.default.yaml
+++ b/src/Tests/Tests.Configuration/tests.default.yaml
@@ -16,6 +16,7 @@ elasticsearch_version: 6.3.0
 # do not spawn nodes as part of the test setup if we find a node is already running
 # this is opt in during development in CI we never want to see our tests running against an already running node
 test_against_already_running_elasticsearch: true
+# elasticsearch_out_after_started: true
 
 #random_source_serializer: true
 #random_old_connection: true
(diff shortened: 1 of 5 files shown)


Mpdreamz:
backported to `6.x` `6.3` and `6.4`

PR #3382, opened 2018-08-29
Support Ids in MultiTermVectors API

### LINKED ISSUE #3219: How to use the MultiTermVectorsAsync API
**NEST/Elasticsearch.Net version**: 6.0.2

**Elasticsearch version**: 6.2.3

I am trying to call the below query using NEST
```
GET 123_original/_doc/_mtermvectors
{
  "ids": [
    "9a271078-086f-4f4b-8ca0-16376c2f49a7",
    "481ce3db-69bf-4886-9c38-fcb878d44925"
  ],
  "parameters": {
    "fields": ["*"],
    "positions": false,
    "offsets": false,
    "payloads": false,
    "term_statistics": false,
    "field_statistics": false
  }
}
```
The NEST API (I think) would look something like this
```
var term = await elasticClient.MultiTermVectorsAsync(x =>
{
    return x.Index(indexOriginal)    // 123_original
        .Type(typeName)              // _doc
        .GetMany<ElasticDataSet>(ids.Keys) // list of strings
        .Fields("*")
        .FieldStatistics(false)
        .Positions(false)
        .Offsets(false)
        .TermStatistics(false)
        .Payloads(false);
});
```
The problem is that the above API is returning the following error
`Index name is null for the given type and no default index is set. Map an index name using ConnectionSettings.DefaultMappingFor<TDocument>() or set a default index using ConnectionSettings.DefaultIndex().`

And this is the query that its trying to execute which has the index in it and is missing the ids, but works in Kibana when the ids are set.
`123_original/_doc/_mtermvectors?fields=%2A&field_statistics=false&positions=false&offsets=false&term_statistics=false&payloads=false`

I cannot find a documentation on how to use the Multi Term Vector using NEST.

related Stackoverflow issue: https://stackoverflow.com/questions/50058822/how-to-use-multitermvectorsasync

russcam:
This commit adds support for providing a set of Ids to MultiTermVectors API to be used in
conjunction with index and type provided in the URI.

Index() and Type() methods added to MultiTermVectorOperation to allow the default typeof(T)
values to be overidden.

Closes #3219

### DIFF
--- a/src/Nest/Document/Multiple/MultiTermVectors/MultiTermVectorsRequest.cs
+++ b/src/Nest/Document/Multiple/MultiTermVectors/MultiTermVectorsRequest.cs
@@ -5,43 +5,84 @@ using Newtonsoft.Json;
 
 namespace Nest
 {
+	/// <summary>
+	/// A Multi termvectors API request
+	/// </summary>
 	public partial interface IMultiTermVectorsRequest
 	{
+		/// <summary>
+		/// The documents for which to generate term vectors
+		/// </summary>
 		[JsonProperty("docs")]
 		IEnumerable<IMultiTermVectorOperation> Documents { get; set; }
+
+		/// <summary>
+		/// The ids of documents within the same index and type
+		/// for which to generate term vectors. Must be used in
+		/// conjunction with <see cref="Index"/> and <see cref="Type"/>
+		/// </summary>
+		[JsonProperty("ids")]
+		IEnumerable<Id> Ids { get; set; }
 	}
 
+	/// <inheritdoc cref="IMultiTermVectorsRequest"/>
 	public partial class MultiTermVectorsRequest
 	{
+		/// <inheritdoc />
 		public IEnumerable<IMultiTermVectorOperation> Documents { get; set; }
+
+		/// <inheritdoc />
+		public IEnumerable<Id> Ids { get; set; }
 	}
 
+	/// <inheritdoc cref="IMultiTermVectorsRequest"/>
 	[DescriptorFor("Mtermvectors")]
 	public partial class MultiTermVectorsDescriptor
 	{
-		private List<IMultiTermVectorOperation> _operations = new List<IMultiTermVectorOperation>();
+		private List<IMultiTermVectorOperation> _operations;
+
+		private List<IMultiTermVectorOperation> Operations =>
+			this._operations ?? (this._operations = new List<IMultiTermVectorOperation>());
+
 		IEnumerable<IMultiTermVectorOperation> IMultiTermVectorsRequest.Documents
 		{
-			get { return this._operations; }
-			set { this._operations = value?.ToList(); }
+			get => this._operations;
+			set => this._operations = value?.ToList();
 		}
 
+		IEnumerable<Id> IMultiTermVectorsRequest.Ids { get; set; }
+
+		// TODO: Rename to Documents in 7.x
+		/// <summary>
+		/// A document for which to generate term vectors
+		/// </summary>
 		public MultiTermVectorsDescriptor Get<T>(Func<MultiTermVectorOperationDescriptor<T>, IMultiTermVectorOperation> getSelector)
 			where T : class =>
-			Assign(a => this._operations.AddIfNotNull(getSelector?.Invoke(new MultiTermVectorOperationDescriptor<T>())));
+			Assign(a => this.Operations.AddIfNotNull(getSelector?.Invoke(new MultiTermVectorOperationDescriptor<T>())));
 
+		// TODO: Rename to Documents in 7.x
+		/// <inheritdoc cref="IMultiTermVectorsRequest.Documents"/>
 		public MultiTermVectorsDescriptor GetMany<T>(IEnumerable<long> ids,
 			Func<MultiTermVectorOperationDescriptor<T>, long, IMultiTermVectorOperation> getSelector = null)
 			where T : class =>
-			Assign(a => this._operations.AddRange(ids.Select(id => getSelector.InvokeOrDefault(new MultiTermVectorOperationDescriptor<T>().Id(id), id))));
+			Assign(a => this.Operations.AddRange(ids.Select(id => getSelector.InvokeOrDefault(new MultiTermVectorOperationDescriptor<T>().Id(id), id))));
 
+		// TODO: Rename to Documents in 7.x
+		/// <inheritdoc cref="IMultiTermVectorsRequest.Documents"/>
 		public MultiTermVectorsDescriptor GetMany<T>(IEnumerable<string> ids, Func<MultiTermVectorOperationDescriptor<T>, string, IMultiTermVectorOperation> getSelector = null)
 			where T : class =>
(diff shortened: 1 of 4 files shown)


src/Tests/Tests/Document/Multiple/MultiTermVectors/MultiTermVectorsApiTests.cs
```diff
@@ -123,4 +123,84 @@ private static void AssertTermVectors(TermVector vectors)
...
+		protected override int ExpectStatusCode => 200;
+		protected override HttpMethod HttpMethod => HttpMethod.POST;
+		protected override string UrlPath =>
+			$"/devs/developer/_mtermvectors?field_statistics=true&payloads=true&term_statistics=true&positions=true&offsets=true";
+
+		protected override bool SupportsDeserialization => false;
+
+		protected override object ExpectJson { get; } = new
+		{
+			ids = Developer.Developers.Select(p => (Id)p.Id).Take(2)
```
codebrain: Balmer-object orientation.

src/Tests/Tests/Document/Multiple/MultiTermVectors/MultiTermVectorsApiTests.cs
```diff
@@ -123,4 +123,84 @@ private static void AssertTermVectors(TermVector vectors)
...
+		};
+
+		protected override void ExpectResponse(IMultiTermVectorsResponse response)
+		{
+			response.ShouldBeValid();
+			response.Documents.Should().NotBeEmpty().And.HaveCount(2).And.OnlyContain(d => d.Found);
```
codebrain: And.OnlyContain... nice!

src/Nest/Document/Multiple/MultiTermVectors/MultiTermVectorsRequest.cs
```diff
@@ -5,43 +5,84 @@
...
+
+		private List<IMultiTermVectorOperation> Operations =>
+			this._operations ?? (this._operations = new List<IMultiTermVectorOperation>());
+
 		IEnumerable<IMultiTermVectorOperation> IMultiTermVectorsRequest.Documents
 		{
-			get { return this._operations; }
-			set { this._operations = value?.ToList(); }
+			get => this._operations;
+			set => this._operations = value?.ToList();
```
codebrain: Just out of curiosity, is there any reason why we `.ToList()` on setters like this; other than to force the execution of the `IEnumerable` at this point? Seems like we could save some allocations in these instances...

src/Tests/Tests/Document/Multiple/MultiTermVectors/MultiTermVectorsApiTests.cs
russcam: I thought about providing overloads for `string`, `long`, `Guid`, etc. much like the existing `GetMany<T>(...)`.

src/Nest/Document/Multiple/MultiTermVectors/MultiTermVectorsRequest.cs
russcam: I don't know about this specific case, but in general would be to force execution as you say.

src/Tests/Tests/Document/Multiple/MultiTermVectors/MultiTermVectorsApiTests.cs
codebrain: Could be useful, but easy to add later :)

src/Nest/Document/Multiple/MultiTermVectors/MultiTermVectorsRequest.cs
russcam: What is unusual with respect to the general API design in NEST, is that `GetMany<T>(..)` is additive rather than assignative to the operations. I understand why this is, and it's not the only place where a method call is additive

src/Tests/Tests/Document/Multiple/MultiTermVectors/MultiTermVectorsApiTests.cs
russcam: that's what I err'ed on!

Mpdreamz:
`build seed:58181 test "multitermvectorsids` fails with:

```
   Tests.Document.Multiple.MultiTermVectors.MultiTermVectorsIdsApiTests.SerializesInitializer [FAIL]
      Expected True because RoundTrip: 0 itteration
      Deserialization has result: False
      RoundTripResult`1 success: False
      This is the first time I am serializing
        {
          "ids": [
      -     {},
      -     {}
      +     1001,
      +     1002
          ]
        }
      
       C# approximation of actual ------ 
       new {
        ids = new [] {
          1001,
          1002
        }
      };
       C# approximation of expected ------ 
       new {
        ids = new [] {
          {},
          {}
        }
      };, but found False.
      Stack Trace:
           at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message)
           at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args)
           at FluentAssertions.Primitives.BooleanAssertions.BeTrue(String because, Object[] becauseArgs)
        C:\projects\elasticsearch-net\src\Tests\Tests.Core\Extensions\SerializationTesterAssertionExtensions.cs(9,0): at Tests.Core.Extensions.SerializationTesterAssertionExtensions.ShouldBeValid(SerializationResult result, String message)
        C:\projects\elasticsearch-net\src\Tests\Tests.Core\Extensions\SerializationTesterAssertionExtensions.cs(33,0): at Tests.Core.Extensions.SerializationTesterAssertionExtensions.AssertSerialize[T](SerializationTester tester, T object, Object expectedJson, String message, Boolean preserveNullInExpected)
        C:\projects\elasticsearch-net\src\Tests\Tests.Core\Serialization\ExpectJsonTestBase.cs(24,0): at Tests.Core.Serialization.ExpectJsonTestBase.RoundTripsOrSerializes[T](T object)
        C:\projects\elasticsearch-net\src\Tests\Tests\Framework\EndpointTests\ApiTestBase.cs(37,0): at Tests.Framework.ApiTestBase`5.SerializesInitializer()
    Tests.Document.Multiple.MultiTermVectors.MultiTermVectorsIdsApiTests.SerializesFluent [FAIL]
      Expected True because RoundTrip: 0 itteration
      Deserialization has result: False
      RoundTripResult`1 success: False
      This is the first time I am serializing
        {
          "ids": [
      -     {},
      -     {}
      +     1001,
      +     1002
          ]
        }
      
       C# approximation of actual ------ 
       new {
        ids = new [] {
          1001,
          1002
        }
      };
       C# approximation of expected ------ 
       new {
        ids = new [] {
          {},
          {}
        }
      };, but found False.
```

src/Nest/Document/Multiple/MultiTermVectors/MultiTermVectorsRequest.cs
Mpdreamz: Places where we are additive is AFAIK only in the bulk related API helpers since you can `IndexMany`<X> and then `IndexMany<Y>` afterwards in a single bulk. That was the rationale behind the deviation. Definitely up for further discussion 😄

PR #3389, opened 2018-09-03
SeedNode() on DefaultSeeder parallelizes where possible

Mpdreamz:
Before:
prior

After:

after

Speeds up launching an integration cluster (`ReadOnlyCluster` and others who use `DefaultSeeder`).

### DIFF
--- a/src/Tests/Tests.Core/ManagedElasticsearch/NodeSeeders/DefaultSeeder.cs
+++ b/src/Tests/Tests.Core/ManagedElasticsearch/NodeSeeders/DefaultSeeder.cs
@@ -1,4 +1,7 @@
+using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
 using Nest;
 using Tests.Configuration;
 using Tests.Core.Client;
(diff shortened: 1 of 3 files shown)


src/Tests/Tests.Core/ManagedElasticsearch/NodeSeeders/DefaultSeeder.cs
```diff
@@ -36,22 +39,46 @@ public DefaultSeeder(IElasticClient client, IIndexSettings indexSettings)
...
+			t.Wait(TimeSpan.FromSeconds(40));
+		}
+		public void SeedNodeNoData()
+		{
+			if (!TestClient.Configuration.ForceReseed && this.AlreadySeeded()) return;
+
+			var t = Task.Run(async () => await this.SeedNodeNoDataAsync());
+
+			t.Wait(TimeSpan.FromSeconds(40));
 		}
```
russcam: Should we consider removing both sync methods in favour of making `SeedCluster()` async on the cluster?

Mpdreamz: We should! Wanted to keep this small before the 6.3.0 release

Mpdreamz:
ported to `master`

PR #3390, opened 2018-09-04
Split all the BulkAll and ScrollAll related tests into their own file…

Mpdreamz:
…s and classes making it easier to read what they are doing


Mpdreamz:
ported to `master` and `5.x`

PR #3393, opened 2018-09-10
Update TypeNameResolver.cs

msarilar:
check name if null before using it


Mpdreamz:
@msarilar this LGTM thank you for submitting this. Mind signing the Elastic CLA https://www.elastic.co/contributor-agreement so they we may pull these bits in?

Mpdreamz:
Ported to `6.x` and `5.x`. Thanks for the PR @msarilar 👍

PR #3394, opened 2018-09-10
scripts is now an SDK style project as well that is part of the solut…

Mpdreamz:
…ion build


Mpdreamz:
ported to `master` and `5.x`

PR #3405, opened 2018-09-14
Fix #3404 bulk all improvements

### LINKED ISSUE #3404: Bulk Insert is failing silently
**NEST/Elasticsearch.Net version**: 6.3.0

**Elasticsearch version**: 6.2.3

Bulk insert is failing silently. I have a BulkAllObserver with a listener on `onError` to log out any errors that occur during the batching but no errors are being output. This has been hiding data loading errors for weeks in production and we only just noticed. This seems to be true for all error conditions, but I will provide a single case that should be testable

**Steps to reproduce**:
 1. Create a new, blank index and define a data schema
 2. Create a bulk operation to insert a ton of documents into this index but have some of the object properties violate the schema (e.g. try to insert Foo : string when the schema requires Foo : int)
 3. Subscribe a BulkAllObserver to output errors during the operation 
 4. Verify that no errors were logged out
 5. Verify that the index contains no documents

Here is the code I am currently using:
``` csharp
var operationId = Guid.NewGuid();
Logger.Information("Beginning bulk insert operation to ElasticSearch. Operation ID: {operationId}", this, operationId);

var waitHandle = new CountdownEvent(1);
var bulkInsert = _client.BulkAll(vehicles, b => b
	.Index("inventory")
	.BackOffRetries(2)
	.BackOffTime("30s")
	.RefreshOnCompleted()
	.MaxDegreeOfParallelism(4)
	.Size(100));

bulkInsert.Subscribe(new BulkAllObserver(
	onError:     ex => Logger.Fatal("Bulk insert error {message} on operation {operationId}", this, ex, ex.Message, operationId),
	onCompleted: () =>
	{
		Logger.Information(
			"Successfully inserted {count} documents. Operation ID: {operationId}", 
			this, 
			vehicles.Count, 
			operationId);

		waitHandle.Signal();
	}));

waitHandle.Wait();
```

Mpdreamz:
The current `BulkAll()` implementation was written to steamroll its way to completion. This leads to a lot of unnecessary request in the case where most documents are expected to fail e.g due to a bad mapping.

This assumption is not clear from the contract nor our docs hence the bug label.

This PR halts `BulkAll()` if any documents with errors are returned that can not be retried.
It also halts or retries based on the `_bulk` requests status itself. 
It introduces a more explicit `FailedOverAllNodes` audit event which currently bubbles out as a `MaxRetries` only.

Finally you can opt in to the streamroll by setting `ContinueAfterDroppedDocuments` to true which will feed all documents it can not retry to an optional `DroppedDocumentCallback` which can put it in a DLQ for instance.


Mpdreamz:
backported to `master` `5.x`

Mpdreamz:
f7e326225ac48f072c1fbc50a2c1e4077fc05450

Introduces a couple of adjustments, the predicate was checking `response.IsValid` in the `Where` to select retry and dropped documents. I changed this to a more explict `item.IsValid`. As this was considering too many non retryable documents as dropped if the request was invalid but the item itself was.

Another change is that the `DroppedDocumentCallback` is always called no matter if `ContinueAfterDroppedDocuments` is true or not. If its `false` it will return the dropped documents of the failed buffer which is very usefull.

This change is now in `6.x` and `master` as well.

PR #3408, opened 2018-09-18
Use AssertionScope to always include response.DebugInformation

Mpdreamz:
When we do our response testing in the API integration tests only.

This changes the way assertions are executed though, reporting on all ALL the failures. 
Since we do quite a few assertions in a single method this would become noisy so we hack the message to only include the first failure as before


Mpdreamz:
@russcam @codebrain using my discretion to pull this

Mpdreamz:
ported to `master` and `5.x`

PR #3416, opened 2018-09-24
obsolete copy_settings as of 6.4 it will no longer allow false and wi…

Mpdreamz:
…ll be removed on the server in 8.x and in the client in 7.x


src/CodeGeneration/ApiGenerator/Overrides/GlobalOverrides.cs
```diff
@@ -35,7 +35,8 @@ public class GlobalOverrides : EndpointOverridesBase
 		public override IDictionary<string, string> ObsoleteQueryStringParams { get; set; } = new Dictionary<string, string>
 		{
 			{ "parent", "the parent parameter has been deprecated from elasticsearch, please use routing instead directly."},
-			{ "update_all_types", "Elasticsearch 6.x only allows a single type per index so this parameter is now useless"}
+			{ "update_all_types", "Elasticsearch 6.x only allows a single type per index so this parameter is now useless"},
+			{ "copy_settings", "Elasticsearch 6.4 will throw an exception if this is turned off see elastic/elasticsearch#30404"}
```
russcam: Having the link to the commit was useful to understand the behaviour. Do we want to surface it in the Obsolete message though (too much information)?

PR #3417, opened 2018-09-24
Add support for weighted average aggregation

Mpdreamz:
https://www.elastic.co/guide/en/elasticsearch/reference/6.x/search-aggregations-metrics-weight-avg-aggregation.html

### DIFF
--- a/src/Nest/Aggregations/Metric/WeightedAverage/WeightedAverageAggregation.cs
+++ b/src/Nest/Aggregations/Metric/WeightedAverage/WeightedAverageAggregation.cs
@@ -1,0 +1,78 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using Newtonsoft.Json;
+
+namespace Nest
+{
+	[JsonObject(MemberSerialization = MemberSerialization.OptIn)]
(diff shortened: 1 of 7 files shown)


src/Nest/Aggregations/Metric/WeightedAverage/WeightedAverageAggregation.cs
```diff
@@ -0,0 +1,78 @@
...
+		[JsonProperty("weight")]
+		IWeightedAverageValue Weight { get; set; }
+		/// <summary> The optional numeric response formatter </summary>
+		[JsonProperty("format")]
+		string Format { get; set; }
+		/// <summary> A hint about the values for pure scripts or unmapped fields </summary>
+		[JsonProperty("value_type")]
+		// TODO map as on server enum ?
+		// https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/search/aggregations/support/ValueType.java
+		string ValueType { get; set; }
```
russcam: Weighted Avg Aggregation is not marked as beta or experimental, so mapping as an enum might be better, as it would help guide users as to which values are supported.

PR #3426, opened 2018-09-28
add support for remove duplicates token filter

Mpdreamz:
pending #3424 and #3425 should be re-targetted to #3424


src/Nest/Analysis/TokenFilters/RemoveDuplicatesTokenFilter.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Nest
+{
+	/// <summary>
+	/// A token filter of type multiplexer will emit multiple tokens at the same position, each version of the token having
+	/// been run through a different filter. Identical output tokens at the same position will be removed.
+	/// </summary>
```
russcam: Documentation looks like it's for the Multiplexer Token Filter

PR #3427, opened 2018-09-28
add support for the char_group tokenizer

Mpdreamz:
pending #3424


src/Nest/Analysis/Tokenizers/CharGroupTokenizer.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+	/// <summary>
+	/// The char_group tokenizer breaks text into terms whenever it encounters a character which is in a defined set. It is mostly useful
+	/// for cases where a simple custom tokenization is desired, and the overhead of use of the pattern tokenizer is not acceptable.
+	/// </summary>
+	public interface ICharGroupTokenizer : ITokenizer
+	{
+		/// <summary>
+		/// The maximum token length. If a token is seen that exceeds this length then it is discarded. Defaults to 255.
+		/// </summary>
```
russcam: I think the documentation should be

```
A list containing a list of characters to tokenize the string on. Whenever a character from this list is encountered, a
new token is started. This accepts either single characters like eg. -, or character groups: whitespace, letter, digit,
punctuation, symbol.
```

src/Nest/Analysis/Tokenizers/CharGroupTokenizer.cs
```diff
@@ -0,0 +1,44 @@
...
+		[JsonProperty("tokenize_on_chars")]
+		IEnumerable<string> TokenizeOnCharacters { get; set; }
```
russcam: specialized type that takes a union of enum and `char`? `string` is no doubt easier to use.

Issue #3423, opened 2018-09-27
Nuget package does not include 4.5 framework files for Nest/Elasticsearch.Net v6.2.0


gokussx4:
**NEST/Elasticsearch.Net version**: 6.2.0

**Description of the problem including expected versus actual behavior**:
The nuspec file claims support for .net framework 4.5 but the files included do not include a .net framework 4.5 build.

**Steps to reproduce**:
 1. Review nuspec file for tag 6.2.0, you will see the files listed do not include a .net framework 4.5

gokussx4:
Not sure but looks like a backport merge wiped out the net45 files.

https://github.com/elastic/elasticsearch-net/commit/f719e968ca2550d1ea743848250f1a18ff05b427#diff-cef6b0470ec2e19aa9a8ccaf79f38d00

Mpdreamz:
Hi @gokussx4 thanks for taking the time to report.

This is a duplicate of https://github.com/elastic/elasticsearch-net/issues/3348

gokussx4:
Hi @Mpdreamz, the nuspec file should be updated one way or the other though. Your public nuspec file has a targetframework of 4.5 in it and causes a failure because the lib/net45 wasn't included in the files sections. 
So if you aren't supporting .net 4.5 in version 6.2.0 then remove the targetframework so those installing through nuget will clearly see 6.2.0 does not support .net 4.5. 
Otherwise just add the file section back that included that build.

Does that make sense?

Issue #3236, opened 2018-05-04
Performance hit in NEST 6.x


kaaresylow:
**NEST/Elasticsearch.Net version**:
5.6.2
6.1.0

**Elasticsearch version**:
5.5.1
6.2.3

**Description of the problem including expected versus actual behavior**:
In the process of upgrading our Elasticsearch setup from version 5.x to 6.x we have noticed a performance impact that translates into slower response time in our service that uses Elasticsearch as data storage. The slower response is mostly visible when dealing with large documents with complex structures e.g. geojson. 

The two Elasticsearch clusters performs equally when we are testing them directly via Apache ab so I guess the performance hit is related to the NEST library.

I have done some performance testing using BenchmarkDotNet (on a simplified implementation described further down) where NEST v. 6 has a noticeable higher response compared to v. 5.
I hope you can shed some light on why we are seeing this behaviour. 
Thanks

Method |     Mean |    StdDev |
------- |---------:|---------:|
NEST 5 | 73.11 ms |  2.558 ms |
NEST 6 | 92.15 ms |  3.054 ms |

**Steps to reproduce**:
 1. Add an index to Elasticsearch with this mapping
```
{
  "properties": {
    "boundaries": {
      "properties": {
        "boundary": {
          "type": "geo_shape"
        },
        "type": {
          "type": "text",
          "index": false
        }
      }
    }
  }
}
```

2. Insert document https://gist.github.com/kaaresylow/ece0d2dd88175653fb4fa44710297d14
 
3.

```
var connectionPool = new SingleNodeConnectionPool(new Uri("url"));
var settings = new ConnectionSettings(connectionPool).EnableHttpCompression();
var client = new ElasticClient(settings);
```

```
var response = client.Get<Document>(1, idx => idx
    .Index("documents")
    .Type("documentmodel"));
```

```
public class Document
    {
        public IEnumerable<PlaceBoundary> Boundaries { get; set; }
    }

    public class PlaceBoundary
    {
        public string Type { get; set; }

        public GeoJsonBoundary Boundary { get; set; }
    }

    public class GeoJsonBoundary
    {
        public string Type { get; set; }

        public object Coordinates { get; set; }

        public object Bbox { get; set; }

        public object Crs { get; set; }
    }
```

russcam:
Thanks for opening @kaaresylow. I'm keen to dive into this but I'm unlikely to get around to it until next week.

russcam:
[...]

Issue #1719, opened 2016-01-13
Elasticsearch .NET - Framework 4.0


Romanx:
The current 2.0.0-alpha on nuget seems to have removed the v4.0 build of this although NEST still seems to have this.

Is there a reason it was removed as I'm using nest with [Quartz](https://github.com/quartznet/quartznet) which only has a 4.0 build and we'd like to get ready for 2.0

cstlaurent:
From microsoft announcement: https://support.microsoft.com/en-us/gp/framework_faq/en-us?wa=wsignin1.0, .NET 4.x, 4.5, 4.5.1 are deprecated since January 12, 2016. I don't think the elasticsearch.net team will use their precious time to support it any longer. You should probably ask the Quartz team to create a build for 4.5.2 - 4.6.x or .NET core instead...

Romanx:
I did see that that was the case and all of our other software supports it. I agree with not necessarily supporting it actively but if you can still build against the target without issue then why take away the option, especially as NEST still has a 4.0 option but it can't be used at all without a 4.0 version of Elasticsearch.NET

Mpdreamz:
Hi @Romanx 

We already use features and api's that are a pain to backport to NET4.0. Not impossible but not worth the added complexity. 

I rather put that effort in getting pcl/uap/corefx support out and that for us takes priority over .NET 4.0 which has EOL'ed.

I agree with @cstlaurent, thanks for chiming in btw!, the onus probably lies with Quartz to get .NET 4.5+ builds out there. We have to move forward as a .NET community :smile:

Romanx:
Thanks for the response, I've opened an issue with Quartz to help get them into the future with the rest of us.

Issue #48, opened 2012-06-20
.NET 3.5 compliance?


macrauder:
Hi,

is it possible to create a build profile that is creating NEST as a .Net 3.5 compliant DLL?
We need .Net 3.5 because we like to use NEST in a SharePoint environment and these is only 3.5 compliant.
A manuel change in the solution to .Net 3.5 wasn't succesfully.
Thx Thomas

KamranShahid:
I also needed it for a 3.5 version. I have a .net 3.5 version of task Parallel library avaibale n codeplex which I have also used in my 3.5 version.Problem was use of dynamic feature of .net 4.0.
Is there any old repository of Nest for 3.5 where we can get the code and will try to compile it in 3.5?

PR #3432, opened 2018-09-28
Add .net 45 file source for 6.2.0

gokussx4:
Sorry for the incorrect base here but what I am proposing is make a branch off of the 6.2.0 tag so that this bug can be corrected. That same should be done for ElasticSearch.net.

@Mpdreamz I thought this might be a better way of explaining what I am saying. Thanks!
https://github.com/elastic/elasticsearch-net/issues/3423

### DIFF
--- a/build/NEST.nuspec
+++ b/build/NEST.nuspec
@@ -38,6 +38,9 @@
 		</dependencies>
 	</metadata>
 	<files>
+		<file src="output\Nest\net45\Nest.dll" target="lib\net45"/>
+		<file src="output\Nest\net45\Nest.XML" target="lib\net45"/>
+		
 		<file src="output\Nest\net46\Nest.dll" target="lib\net46"/>
 		<file src="output\Nest\net46\Nest.XML" target="lib\net46"/>
 


Mpdreamz:
Hey @gokussx4 thanks for the PR :smile:

We removed `net45` support in between `6.1.0` and `6.2.0` as per:

https://github.com/elastic/elasticsearch-net/pull/3266

The bug was that we left the `net45` dependencyGroup in the `nuspec` file:

https://github.com/elastic/elasticsearch-net/blob/6.2.0/build/NEST.nuspec#L19-L21

We are not building `net45` output on [`6.2.0`](https://github.com/elastic/elasticsearch-net/blob/6.2.0/src/Nest/Nest.csproj) to be included as you propose. We did this very much on purpose as only `4.5.2` is still supported by MS. For now we target the lower end equivalent of` netstandard 2.0` but we might even jump that to `net47` so that we can start using the latest greatest c# features such as ValueTuple without needing additional nuget packages. We also want to include `netcoreapp2.<latest>` as explicit TFM so that we can start to make use of the improvements to HttpClient (SocketsHandler) which should bring some stability to the client performance on linux (see #3236).  

Moving the needle on our lower end of supported TFM's is never an easy decision but in the long run we feel its for the best. E.g we [removed `4.0` support](https://github.com/elastic/elasticsearch-net/issues/1719) nearly 2 years ago and [`3.5` 6 years ago](https://github.com/elastic/elasticsearch-net/issues/48).

We can not republish the `6.2.0` package sadly. 

I will unlist the `6.2.0` packages tomorrow which is the best we can do at this point. cc @codebrain @russcam.

PR #3438, opened 2018-10-01
Processors can now tested in isolation and guarded in isolation with …

Mpdreamz:
…a SkipVersion


src/Tests/Tests/Ingest/ProcessorAssertions.cs
```diff
@@ -0,0 +1,345 @@
...
+
+		public static IPromise<IList<IProcessor>> Fluent(ProcessorsDescriptor d)
+		{
+			foreach (var a in All) a.Fluent(d);
+			return d;
+		}
+
+		public static IProcessor[] Initializer => All.Select(a => a.Initializer).ToArray();
+
+		public class Append : ProcessorAssertion
```
russcam: `AppendProcessorAssertion`, etc. ?

Mpdreamz: I didn't bother since these are nested and will show up as 
`parent+nested.method` in test runners

PR #3439, opened 2018-10-01
Add KeyValueProcessor tests and add support for trimming properties

Mpdreamz:
pending #3438


src/Nest/Ingest/ProcessorsDescriptor.cs
```diff
@@ -124,9 +124,14 @@ public class ProcessorsDescriptor : DescriptorPromiseBase<ProcessorsDescriptor,
 		public ProcessorsDescriptor UserAgent<T>(Func<UserAgentProcessorDescriptor<T>, IUserAgentProcessor> selector) where T : class  =>
 			Assign(a => a.AddIfNotNull(selector?.Invoke(new UserAgentProcessorDescriptor<T>())));
 
+		[Obsolete("This method takes the wrong descriptor please use Kv")]
```
russcam: ++

PR #3440, opened 2018-10-01
add support for the bytes processor

Mpdreamz:
pending #3438


src/Nest/Ingest/Processors/BytesProcessor.cs
```diff
@@ -0,0 +1,74 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.Linq.Expressions;
+
+namespace Nest
+{
+	/// <summary>
+	/// URL-decodes a string
+	/// </summary>
```
russcam: Incorrect docs. Think this should be

```
Converts a human readable byte value (e.g. 1kb) to its value in bytes (e.g. 1024).

Supported human readable units are "b", "kb", "mb", "gb", "tb", "pb" case insensitive. An error will occur if the field is not a supported format or resultant value exceeds 2^63.
```

PR #3447, opened 2018-10-01
Feature/alias is write index

Mpdreamz:


src/Nest/CommonAbstractions/Response/DictionaryResponseBase.cs
```diff
@@ -14,7 +14,9 @@ public interface IDictionaryResponse<TKey, TValue> : IResponse
 	public abstract class DictionaryResponseBase<TKey, TValue> : ResponseBase, IDictionaryResponse<TKey, TValue>
 	{
 		protected IDictionaryResponse<TKey, TValue> Self => this;
-		IReadOnlyDictionary<TKey, TValue> IDictionaryResponse<TKey, TValue>.BackingDictionary { get; set; }
+
+		IReadOnlyDictionary<TKey, TValue> IDictionaryResponse<TKey, TValue>.BackingDictionary { get; set; } =
+			EmptyReadOnly<TKey, TValue>.Dictionary;
```
russcam: Does this need to be initialized with `EmptyReadOnly<TKey, TValue>.Dictionary`? Looks like this is initialized in the `JsonConverter`

https://github.com/elastic/elasticsearch-net/blob/1db9c481b4cf5c3c51366fe34ca3afd1b1f3b52a/src/Nest/CommonAbstractions/Response/DictionaryResponseBase.cs#L62-L64

Mpdreamz: Perhaps not :+1: if we want a codestandards test we might want to keep the duplication.

PR #3465, opened 2018-10-29
Allow null refresh_interval to be passed

### LINKED ISSUE #3402: Can't set refresh_interval to default value (null)
6.3.0

**Elasticsearch version**:
6.4.0

`refresh_interval` setting cannot be changed to default (`null`) because `null` values are not sent.
I also don't see an option to just send `null`'s for a single request.

Expected: `refresh_interval` set to `null`
Actual: `refresh_interval` doesn't update to `null`

**Steps to reproduce**:
 1. Create Index using `CreateIndexAsync`
 2. Update Index settings to set `refresh_interval` to -1
```
UpdateIndexSettingsDescriptor GetUpdateDescriptor(Time refreshInterval)
{
   return new UpdateIndexSettingsDescriptor()
      .Index("my-index")
      .IndexSettings(s => s
         .RefreshInterval(refreshInterval)
      );
}

await _client.UpdateIndexSettingsAsync(GetUpdateDescriptor(-1));
```
 3. Reset `refresh_interval` to default by providing `null`
```
await _client.UpdateIndexSettingsAsync(GetUpdateDescriptor(null));
```
4. Retrieve settings of Index, `refresh_interval` is still -1, `GET my-index/_settings` returns:
```
{
  "my-index": {
    "settings": {
      "index": {
        "refresh_interval": "-1"
      }
    }
  }
}
```

`new ConnectionSettings(node).DefaultTypeName("_doc")`

N/A

russcam:
This commit allows a null value to be passed for refresh_interval, which
allows a user to remove a transient value to fallback to the configured
persistent value.

A couple of different approaches were considered

1. Introducing an internal field on concrete implementations of IDynamicIndexSettings
that could be checked in IndexSettingsConverter to determine whether to serialize
the refresh_interval value, allowing a user to pass null. This has the downside of
needing to cast to concrete implementations in IndexSettingsConverter, rendering the
interface less useful.
2. Introducing a Time 'null' value with a Null Object pattern implementation, and
setting a Time.Null value in the event that null is passed to the property setter.
This would mean that IndexSettingsConverter would not need to change, but now introduces
much more complexity within the Time type to handle a new special value for equality and
comparisons.

In the end, the least intrusive approach of setting the key and value in the underlying
dictionary, and handling whether the key/value should be skipped in IndexSettingsConverter
was considered the cleanest.

Closes #3402

### DIFF
--- a/src/Nest/IndexModules/IndexSettings/Settings/DynamicIndexSettings.cs
+++ b/src/Nest/IndexModules/IndexSettings/Settings/DynamicIndexSettings.cs
@@ -5,6 +5,9 @@ using Newtonsoft.Json;
 
 namespace Nest
 {
+	/// <summary>
+	/// Dynamic index settings
+	/// </summary>
 	[ContractJsonConverter(typeof(IndexSettingsConverter))]
 	public interface IDynamicIndexSettings : IIsADictionary<string, object>
 	{
@@ -98,8 +101,10 @@ namespace Nest
 		ISimilarities Similarity { get; set; }
 	}
 
+	/// <inheritdoc />
 	public class DynamicIndexSettings : IsADictionaryBase<string, object>, IDynamicIndexSettings
 	{
+		private Time _refreshInterval;
 		public DynamicIndexSettings() { }
 
 		public DynamicIndexSettings(IDictionary<string, object> container) : base(container) { }
@@ -135,7 +140,15 @@ namespace Nest
 		public Union<int, RecoveryInitialShards> RecoveryInitialShards { get; set; }
 
 		/// <inheritdoc/>
-		public Time RefreshInterval { get; set; }
+		public Time RefreshInterval
+		{
+			get => _refreshInterval;
+			set
+			{
+				this.BackingDictionary[UpdatableIndexSettings.RefreshInterval] = value;
+				_refreshInterval = value;
+			}
+		}
 
 		/// <inheritdoc/>
 		public int? RoutingAllocationTotalShardsPerNode { get; set; }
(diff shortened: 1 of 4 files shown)


russcam:
We may consider simplifying `IndexSettingsConverter` in future by taking a similar approach for all known settings i.e. getting/setting values in the underlying dictionary within properties.

src/Tests/Tests/Document/Multiple/Bulk/BulkApiTests.cs
```diff
@@ -38,7 +38,7 @@ protected override void IntegrationSetup(IElasticClient client, CallUniqueValues
 			);
 
 			if (!pipelineResponse.IsValid)
-				throw new Exception("Failed to set up pipeline required for bulk");
+				throw new Exception("Failed to set up pipeline named 'default-pipeline' required for bulk");
```
Mpdreamz: ❤️

src/Nest/IndexModules/IndexSettings/Settings/DynamicIndexSettings.cs
```diff
@@ -158,88 +171,89 @@ public class DynamicIndexSettings : IsADictionaryBase<string, object>, IDynamicI
...
 		protected DynamicIndexSettingsDescriptorBase(TIndexSettings instance) : base(instance) { }
 
-		/// <summary>
-		/// Add any setting to the index
-		/// </summary>
+		/// <inheritdoc cref="DynamicIndexSettings.Add"/>
 		public TDescriptor Setting(string setting, object value)
 		{
-			this.PromisedValue.Add(setting, value);
+			this.PromisedValue[setting] = value;
```
Mpdreamz: 👍

Mpdreamz:
backported to `6.x` and `5.x`

PR #3466, opened 2018-10-29
try/catch when attempting to deserialize ServerError from response

### LINKED ISSUE #3378: StringResponse.TryGetServerError parse error on 401 response
**NEST/Elasticsearch.Net version**: 6.2

**Elasticsearch version**: (any)

**Description of the problem including expected versus actual behavior**: when StringResponse is used and server returns 401 (requires HTTP auth) 'Invalid JSON string' exception is thrown because [ServerError class](https://github.com/elastic/elasticsearch-net/blob/master/src/Elasticsearch.Net/Responses/ServerException/ServerError.cs) tries to parse response body as JSON. SkipDeserializationForStatusCodes doesn't help in this case because it is ignored for types listed in ResponseBuilder.SetSpecialTypes.

**Steps to reproduce**:

> var client = new ElasticLowLevelClient(new ConnectionConfiguration(new Uri(anyUrlThatRequiresHttpAuth)));
> var res = client.Search&lt;StringResponse&gt;("someIndex",PostData.String(anyQueryJson));
> // that's all, 'SerializationException: Invalid JSON string' is thrown
> 

In fact this issue can be reproduced for 'BytesResponse' too. 
Wrapping [StringResponse.TryGetServerError](https://github.com/elastic/elasticsearch-net/blob/master/src/Elasticsearch.Net/Responses/Special/StringResponse.cs) code with try-catch might solve the issue.

russcam:
This commit adds a try/catch block around attempting to deserialize a ServerError
from the body, since IElasticsearchResponse types withn Elasticsearch.Net
set/attempt to set the response body, irrespective of the response status code.

Move BytesResponseTests and StringResponseTests from Tests.Reproduce into
main Tests project.
Allow byte array to be passed to FixedResponseClient.

Fixes #3378


src/Elasticsearch.Net/Responses/Special/BytesResponse.cs
```diff
@@ -12,9 +13,16 @@ public bool TryGetServerError(out ServerError serverError)
 		{
 			serverError = null;
 			if (this.Body == null || this.Body.Length == 0) return false;
-			using(var stream = new MemoryStream(this.Body))
-				serverError = ServerError.Create(stream);
-			return true;
+			try
```
Mpdreamz: Should the try catch happen inside `ServerError.Create()`? Can we use the response mimetype as a hint as well here? e.g in the nginx proxy case we already know what we got back from the server is not json.

russcam: ++ to using the MIME type.

> Should the try catch happen inside ServerError.Create()

I think the try/catch should be in the `Try*` method, with the potential for `ServerError.Create()` to throw an exception if a `ServerError` cannot be created from the `Stream`. To my mind, this aligns with BCL `Try*` method semantics.

Mpdreamz: TryX pattern would be even better :+1:

Mpdreamz: Its more about the case where the MIME type informs us the response is not `JSON` we can omit the `TryCreate()` all together.

russcam: So I understand correctly, are you suggesting 

1. if the MIME type is "application/json", call `ServerError.Create();` inside `BytesResponse`/`StringResponse`
2. if the MIME type is anything else, use `ServerError.TryCreate()`
?

Mpdreamz: 1. Call trycreate
2. Omit all together? A bad proxy could break this but a bad proxy is broken anyways.

russcam: Have now added that check. Would you mind taking another look?

russcam:
backported to 6.x in 42414ec116f193881894d4565e3f5bae0a52e053

russcam:
does not need to be ported to 5.x as 5.x has a different implementation

PR #3481, opened 2018-11-01
Cleanup/master

Mpdreamz:


Mpdreamz:
Closing this and opening another.

- These code settings are indenting multiline xmldocs which looks bad. 

```
<summary>
         Text Here
</summary>
```

- Prefer no braces for single line is not smart enough to deal with complex bodies 

```
using(disposable)
     foreach(var in y)
          do()
                .Something();
````

This is now fixed using a resharper specific editorconfig settings. 

@codebrain @russcam will leave the ordering as is on this second pass. The case I found where ordering statics caused a run time error in the test framework was because it used `public static new` which is hacky to begin with.

PR #3486, opened 2018-11-06
Fix #3485

### LINKED ISSUE #3485: Response DebugInformation property exposes login credentials
**NEST/Elasticsearch.Net version**: NEST 5.5.0 (and possibly others)

**Elasticsearch version**: 6.4.0 (and possibly others)

The DebugInformation property on the query result object shows useful information about the result of the query. This includes the full connection string to the ElasticSearch machine.

However, it may occur that this connection string contains username and password. It could be argued that it should not and authentication should be done in a different way. But nevertheless, the DebugInformation property should filter this out. Otherwise, logfiles or other places where the value is used could be contaminated with login data. This is a potential security issue.

**Steps to reproduce**:
 1. Run a NEST query
 2. Log the DebugInformation (string) property from the result object
 3. Observe the log message to find the connectionstring including username and password

**ConnectionSettings**:
https://[redacted_username]:[redacted_password]@[redacted_url]:[redacted_port]

Mpdreamz:
Passing username/pass in the `Node` Uri directly bleeds into `DebugInformation`.

If you use `BasicAuthentication(username, password)` on `ConnectionSettings` which is the recommended route this does not happen.

### DIFF
--- a/src/Tests/Tests/ClientConcepts/Troubleshooting/DebugInformation.doc.cs
+++ b/src/Tests/Tests/ClientConcepts/Troubleshooting/DebugInformation.doc.cs
@@ -29,10 +29,10 @@ namespace Tests.ClientConcepts.Troubleshooting
 	{
 		public DebugInformation(ReadOnlyCluster cluster) : base(cluster) {}
 
-        [I] public void DefaultDebug()
-        {
-            // hide
-            var client = this.Client;
+		[I] public void DefaultDebug()
+		{
+			// hide
+			var client = this.Client;
 
 			var response = client.Search<Project>(s => s
 				.Query(q => q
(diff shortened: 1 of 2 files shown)


src/Tests/Tests/ClientConcepts/Troubleshooting/DebugInformation.doc.cs
```diff
@@ -41,6 +41,41 @@ public void DefaultDebug()
             );
 
             response.DebugInformation.Should().Contain("Valid NEST response");
+        }
+		//hide
+        [U] public void PasswordIsNotExposedInDebugInformation()
```
Mpdreamz: Notice we have spaces here, since `*.doc.cs` have `formatter off` this was not fixed as part of our recent automated code cleanup commits. We really need a `order file off` mechanism but I have not found a way to do that in file comments or editorconfig setting. Will tweet @ jetBrainsRider to see if I missed anything and will fix this on my plane ride home for the active integration branches manually.

Mpdreamz: This should now be fixed on `master" and this PR

PR #3488, opened 2018-11-12
Adding net461 targets to Elasticsearch.Net and Nest.JsonNetSerializer to stay inline with Nest

megakid:
This avoids pulling in all sorts of .NET standard DLLs when referencing NEST from a full framework project.


elasticcla:
Hi @megakid, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/3488.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

megakid:
I see there's more too it than this due to explicit .nuspecs etc

PR #3491, opened 2018-11-19
Include Format property on Field for Doc Values fields

### LINKED ISSUE #3490: Add a format option to docvalue_fields
elastic/elasticsearch#29639

russcam:
This commit adds a format property on the Field type
to use with Doc Values fields, and updates serialization
components to handle the presence of a Format property value.

Closes #3490

### DIFF
--- a/src/Nest/CommonAbstractions/Infer/Field/Field.cs
+++ b/src/Nest/CommonAbstractions/Infer/Field/Field.cs
@@ -4,54 +4,82 @@ using System.Globalization;
 using System.Linq.Expressions;
 using System.Reflection;
 using Elasticsearch.Net;
+using Newtonsoft.Json;
 
 namespace Nest
 {
-	[ContractJsonConverter(typeof(FieldJsonConverter))]
+	/// <summary>
+	/// A field within Elasticsearch
+	/// </summary>
+	[JsonConverter(typeof(FieldJsonConverter))]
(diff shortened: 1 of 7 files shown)


russcam:
backported to 6.x in c6652ac3886a664970b815425d4ca0015a900391

src/Nest/CommonAbstractions/Infer/Field/Field.cs
```diff
@@ -4,54 +4,82 @@
...
+	/// </summary>
+	[JsonConverter(typeof(FieldJsonConverter))]
 	[DebuggerDisplay("{DebugDisplay,nq}")]
 	public class Field : IEquatable<Field>, IUrlParameter
 	{
 		private readonly object _comparisonValue;
 		private readonly Type _type;
 
-		public Field(string name, double? boost = null)
+		public Field(string name, double? boost = null, string format = null)
```
filpen: Isn't this an undocumented breaking change? Just had a MissingMethodException in one of my projects. See https://stackoverflow.com/a/23517490/1282778

Mpdreamz: It absolutely is! This one is totally on me for not running: https://github.com/nullean/AssemblyDiffer before doing the last release as we normally would do. Will see if we can rectify this in the next release. Thank you for bringing this to our attention @filpen

PR #3498, opened 2018-11-26
remove Tests.Profiling

Mpdreamz:
As discussed gets rid of `Tests.Profiling` which was not often used.


Mpdreamz:
ported to `6.x` and `5.x`

PR #3499, opened 2018-11-26
Move differ target over to use AssemblyDiffer

Mpdreamz:
This now lives as more full featured dotnet CLI tool under:

https://github.com/nullean/AssemblyDiffer

@russcam I've also moved the `Differ` project from `elasticsearch-net-abstractions` to the aforementioned repos. The nice part about the `diff` target in our build is that it sets `NUGET` env variable automatically.


Mpdreamz:
ported to `6.x` and `5.x`

PR #3532, opened 2019-01-04
update "Searching" DSL

JamieLybo:
the old "Searching" syntax for ES 6.* is wrong, so I commit the right one

### DIFF
--- a/docs/client-concepts/low-level/getting-started.asciidoc
+++ b/docs/client-concepts/low-level/getting-started.asciidoc
@@ -177,8 +177,7 @@ var searchResponse = lowlevelClient.Search<StringResponse>("people", "person", P
     {
         match = new
         {
-            field = "firstName",
-            query = "Martijn"
+            firstName = "Martijn"
         }
     }
 }));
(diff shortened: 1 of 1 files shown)


JamieLybo:
update "Searching"

maydayzjj:
I have a similar problem. I haven't solved it yet. I tried your method and it really solved it.

russcam:
Thanks for the PR @JamieLybo. The documentation page is generated from 

https://github.com/elastic/elasticsearch-net/blob/6.x/src/Tests/Tests/ClientConcepts/LowLevel/GettingStarted.doc.cs#L188-L192

Please could you update the PR to make the change to that source file? The documentation can then be regenerated from there.

docs/client-concepts/low-level/getting-started.asciidoc
```diff
@@ -99,10 +99,10 @@ var person = new Person
     LastName = "Laarman"
 };
 
-var indexResponse = lowlevelClient.Index<BytesResponse>("people", "person", "1", PostData.Serializable(person)); <1>
+var indexResponse = lowlevelClient.Index<BytesResponse>("people", "person", "1", PostData.Serializable(person)); <1>
```
russcam: Would you be able to update the PR to make these changes in the original source file? The source file is:

https://github.com/elastic/elasticsearch-net/tree/6.x/src/Tests/Tests/ClientConcepts/LowLevel/GettingStarted.doc.cs

russcam: I've made these changes

russcam:
Thanks for the PR, @JamieLybo 👍

PR #3533, opened 2019-01-07
Cross Cluster Replication support

Mpdreamz:
This adds support for all the CCR related API's

- ccr.follow.json
- ccr.follow_stats.json
- ccr.get_auto_follow_pattern.json
- ccr.pause_follow.json
- ccr.put_auto_follow_pattern.json
- ccr.resume_follow.json
- ccr.stats.json
- ccr.unfollow.json

This also introduces `CoordinatedIntegrationTestBase<>`  as a means to execute integration tests in a forced sequence for all api overloads.

Further more it introduces `RemoteClusterConfiguration` to simplify specifying remote clusters as part of the cluster settings API for use with Cross Cluster replication and searches.

You can now also enable soft deletes on index settings in a strongly typed fashion.


src/Tests/Tests/XPack/CrossClusterReplication/CrossClusterReplicationFollowTests.cs
```diff
@@ -0,0 +1,312 @@
...
+using Tests.Core.ManagedElasticsearch.NodeSeeders;
+using Tests.Domain;
+using Tests.Framework;
+using Tests.Framework.EndpointTests.TestState;
+using Tests.Framework.Integration;
+
+namespace Tests.XPack.CrossClusterReplication
+{
+	[SkipVersion("<6.5.0", "")]
+	public class CrossClusterReplicationFollowTests : CoordinatedIntegrationTestBase<WritableCluster>
```
russcam: There's a lot going on in this integration test! I think it may be useful to add a `<summary>` that provides a brief overview of what is being tested.

russcam: As I understand it, each step is dependent on the "setup" of the previous one, and is asserted in its own integration test method in the class. 

If my understanding is correct, maybe we could do something similar here as to what exists for ML job API integration tests - a base class with setup steps that can be called in `IntegrationSetup` method for each deriving integration test, and a derived test class for each separate API?

src/Tests/Tests/XPack/CrossClusterReplication/CrossClusterReplicationAutoFollowTests .cs
```diff
@@ -0,0 +1,117 @@
...
+using Tests.Core.ManagedElasticsearch.NodeSeeders;
+using Tests.Domain;
+using Tests.Framework;
+using Tests.Framework.EndpointTests.TestState;
+using Tests.Framework.Integration;
+
+namespace Tests.XPack.CrossClusterReplication
+{
+	[SkipVersion("<6.5.0", "")]
+	public class CrossClusterReplicationAutoFollowTests : CoordinatedIntegrationTestBase<WritableCluster>
```
russcam: Similar to other comment:

As I understand it, each step is dependent on the "setup" of the previous one, and is asserted in its own integration test method in the class.

If my understanding is correct, maybe we could do something similar here as to what exists for ML job API integration tests - a base class with setup steps that can be called in IntegrationSetup method for each deriving integration test, and a derived test class for each separate API?

Mpdreamz: Aye, The ML approach is valid since most operate on a job so their `IntegrationSetup` is relatively straightforward and inheritable.

The setup in ML run constantly though for each `ApiTest` test class.

CCR is more a chain of API's to call, where the same approach would leave a lot of unfinished setup chains e.g `CoodinatedIntegrationTest` allows us to express this dependency chain:

`A1 => A2 => A3 => A4 = > A5 => A6`

Which will be only executed once for each permutation of `[async, sync]` and `[fluent, ois]`

if we take the object graph approach to test each `AN` in isolation

We end up with:

```
A1
A1 => A2
A1 => A2 => A3 
A1 => A2 => A3 => A4 
A1 => A2 => A3 => A4 = > A5 
A1 => A2 => A3 => A4 = > A5 => A6
```

Foreach permutation of `[async, sync]` and `[fluent, ois]`.

Mpdreamz:
@russcam thank you for the review, I've updated the xmldocs to be less `This API` heavy 😸 

`CoordinatedIntegrationTestBase` is an approach I definitely want to keep for complicated family of API's that do not exist in isolation. E.g `ccr` `autofollow` but also `application privileges` (still working on that in a branch)) and the `ILM` API's from 6.6 will benefit from this less rigid approach to integration tests.

PR #3536, opened 2019-01-09
add support for the reload secure settings API

Mpdreamz:


src/Nest/Cluster/ReloadSecureSettings/ElasticClient-ReloadSecureSettings.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Elasticsearch.Net;
+
+namespace Nest
+{
+	public partial interface IElasticClient
+	{
+		/// <summary>
+		/// This API will decrypt and re-read the entire keystore, on every cluster node, but only the reloadable secure settings will
```
russcam: I think this would flow better starting with "Decrypts and re-reads the entire keystore..."

Mpdreamz: I took this from the elasticsearch docs, inclined to leave it as. Want to at some point have descriptions as part of the json spec.

russcam: I should have been more clear; I think the description is fine, but dislike that it starts with "This API...". This feels superfluous when reading XML docs

PR #3537, opened 2019-01-09
add support for the get ssl certificates API

Mpdreamz:


src/Nest/Nest.csproj
```diff
@@ -12,4 +12,7 @@
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="12.0.1" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="XPack\Ssl" />
+  </ItemGroup>
```
russcam: Superfluous, I think

src/Nest/XPack/Ssl/GetCertificates/GetCertificatesResponse.cs
```diff
@@ -0,0 +1,44 @@
...
+	{
+		public IReadOnlyCollection<ClusterCertificateInformation> Certificates { get; internal set; } =
+			EmptyReadOnly<ClusterCertificateInformation>.Collection;
+	}
+
+	public class ClusterCertificateInformation
+	{
+
+		[JsonProperty("path")]
+		public string Path { get; set; }
```
russcam: `internal set` properties on `ClusterCertificateInformation`

src/Nest/Nest.csproj
Mpdreamz: yeah, annoying not sure what IDE action causes these to pop up yet.

PR #3539, opened 2019-01-09
Update readme.md

afzaal-ahmad-zeeshan:
Added an await expression for usage.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -155,6 +155,9 @@ All the calls have async variants:
 
 ```csharp
 var response = client.IndexAsync(tweet, idx => idx.Index("mytweetindex")); // returns a Task<IndexResponse>
+
+// Or, in an async-context
+var response = await client.IndexAsync(tweet, idx => idx.Index("mytweetindex")); // unwraps a Task<IndexResponse>
 ```
 
 ### Getting a document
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -155,6 +155,9 @@ All the calls have async variants:
 
 ```csharp
 var response = client.IndexAsync(tweet, idx => idx.Index("mytweetindex")); // returns a Task<IndexResponse>
+
+// Or, in an async-context
+var response = await client.IndexAsync(tweet, idx => idx.Index("mytweetindex")); // unwraps a Task<IndexResponse>
```
Mpdreamz: not sure I like the `unwrap` here.  That term is also used in TPL to go from `Task<Task<T>>` to just `Task<T>`. 
maybe change `unwrap` with `await`.

```suggestion
var response = await client.IndexAsync(tweet, idx => idx.Index("mytweetindex")); // awaits a Task<IndexResponse>
```

Mpdreamz:
Thanks for the PR @afzaal-ahmad-zeeshan, I've made one suggested edit.

readme.md
afzaal-ahmad-zeeshan: Thanks, yep, you are right, `await` sounds suitable here.

Mpdreamz:
Thanks @afzaal-ahmad-zeeshan for updating the PR with my nitpick 😄

PR #3545, opened 2019-01-15
fix #3544 JsonNetSerializer now also implements IPropertyMappingProvi…

### LINKED ISSUE #3544: JsonPropertyAttribute not working with FluentSearch
**Elasticsearch version** (`bin/elasticsearch --version`): 6.5.1
**OS version** (`uname -a` if on a Unix-like system): Windows 10 X64, Build 1803

I've got the following classes:
```
public abstract class EventBase {
   ...
   [JsonProperty("source")]
   public SourceModel Source {get;set;}
   ...
}

public class SourceModel {
   ...
   [JsonProperty("geo")]
   public GeoModel Geo {get;set;}
   ...
}

public class GeoModel{
   ...
   [JsonProperty("country_iso_code")]
   public string CountryIsoCode {get;set;}
   ...
}
```

I create a NEST client:
```
var settings = new ConnectionSettings(new SIngleNodeConnectionPool(<elasticuri>), JsonNetSerialize.Default);
```

Next, I want to search ES and aggregate by country:
```
...
var response = await client.SearchAsync<MyModel>(s => s
   .Index("index-name")
   .Size(0)
   .Query(q => q
      .Bool(b => b
         .Filter(bf => bf
            .DateRange(dr => dr
               .Field(f => f.Timestamp)
               .GreaterThanOrEquals(DateTime.Parse(startDate))
               .LessThanOrEquals(DateTime.Parse(endDate))
            )
         )
      )
   )
   .Aggregations(agg => agg
      .Terms("per_country", ta => ta
         .Size(1000)
         .Field(f => f.Source.Geo.CountryIsoCode))
      )
   .MatchAll()
);
...
```
This query does not give me any results, although the index is correct. When I run the query without the aggregation, with size(10) it returns 10 documents.
I used Fiddler to check the data that is being send, and see the following:
```
{
  "size":0,
  "aggs":
  {
    "per_country":
    {
      "terms":
      {
        "field":"source.geo.countryIsoCode",
        "size":1000
      }
    }
  }
  ,"query":
  {
    "match_all":{}
  }
}
```
As you can see the field that is being used in `source.geo.countryIsoCode`, while I expect the field to 'listen' to the JsonAttributes and be `source.geo.country_iso_code`
Adding [JsonObject] on the class level does not fix this issue.

Is my assumption right that this call should respect the JsonProperty attributes on the fields? 
If so, is there something I am doing wrong that this does not work?

Thanks,
Marco

Mpdreamz:
…der making it easier to bootstrap

### DIFF
--- a/src/Tests/Tests/ClientConcepts/HighLevel/Inference/FieldInference.doc.cs
+++ b/src/Tests/Tests/ClientConcepts/HighLevel/Inference/FieldInference.doc.cs
@@ -596,6 +596,8 @@ namespace Tests.ClientConcepts.HighLevel.Inference
 			{
 				"gexo",
 			}).AsPropertiesOf(new SourceModel { Geo = new GeoModel { CountryIsoCode = "nl" } });
+
+			usingSettings.Expect("gexo.country_iso_code").ForField(Field<SourceModel>(p=>p.Geo.CountryIsoCode));
 		}
 
 
--- a/src/Tests/Tests/ClientConcepts/HighLevel/Inference/FieldInference.doc.cs
+++ b/src/Tests/Tests/ClientConcepts/HighLevel/Inference/FieldInference.doc.cs
@@ -8,6 +8,7 @@ using Elastic.Xunit.XunitPlumbing;
 using Elasticsearch.Net;
 using FluentAssertions;
 using Nest;
+using Nest.JsonNetSerializer;
 using Newtonsoft.Json;
 using Tests.Core.Client;
 using Tests.Core.Client.Settings;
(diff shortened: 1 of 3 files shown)


src/Tests/Tests/ClientConcepts/HighLevel/Inference/FieldInference.doc.cs
```diff
@@ -563,5 +564,40 @@ public void CodeBasedConfigurationInherits()
...
+			{
+				"country_iso_code",
+			}).AsPropertiesOf(new GeoModel { CountryIsoCode = "nl" });
+			usingSettings.Expect(new []
+			{
+				"gexo",
+			}).AsPropertiesOf(new SourceModel { Geo = new GeoModel { CountryIsoCode = "nl" } });
+		}
+
+
```
russcam: I think a test for `Field<SourceModel>(p=>p.Geo.CountryIsoCode)` may also be useful here, for completeness?

Mpdreamz: 👍 expanded the unit test to include an assertion for Field inference as well.

PR #3546, opened 2019-01-15
fix #3505 daterange default DateTime should not serialize

### LINKED ISSUE #3505: Incorrect query after using DateTime.MinValue in the DateRange filter
**NEST/Elasticsearch.Net version**: 6.4.0

**Elasticsearch version**: 6.5.1

I am using filter for counting documents like this

```
private static Func<QueryContainerDescriptor<ISearchableNotification>, QueryContainer> IndexedDateFromFilter(SearchQuery query)
{
    if (query.IndexedDateFrom.HasValue)
        return filter => filter.DateRange(rf => rf.Field(nm => nm.IndexedDateTime)
                                                  .GreaterThanOrEquals(query.IndexedDateFrom));
     return null;
}
```

IndexedDateFrom's type is **[DateTime?]**. When **IndexedDateFrom** equals **DateTime.MinValue** everything crashes with exception

`org.elasticsearch.ElasticsearchParseException: failed to parse date field [] with format [strict_date_optional_time||epoch_millis]`

 because NEST translates C# code into this (part of the query I got from DebugInformation)
```
{
                "range": {
                  "indexedDateTime": {
                    "gte": ""
                  }
                }
}
```

If I use **DateTime.MinValue.AddDays(1)** instead of **DateTime.MinValue** everything works fine. It looks like a bug.

Mpdreamz:
Pending the merge of `6.5` branch and all related PR's to that target branch.

### DIFF
--- a/src/Nest/CommonOptions/DateMath/DateMath.cs
+++ b/src/Nest/CommonOptions/DateMath/DateMath.cs
@@ -79,10 +79,11 @@ namespace Nest
 			return math;
 		}
 
+		internal bool IsValid => Self.Anchor.Match(d => d != default, s => !s.IsNullOrEmpty());
+
 		public override string ToString()
 		{
-			var isValid = Self.Anchor.Match(d => d != default(DateTime), s => !s.IsNullOrEmpty());
-			if (!isValid) return string.Empty;
+			if (!IsValid) return string.Empty;
 
 			var separator = Self.Round.HasValue || Self.Ranges.HasAny() ? "||" : string.Empty;
 
--- a/src/Nest/CommonAbstractions/Infer/Field/Field.cs
+++ b/src/Nest/CommonAbstractions/Infer/Field/Field.cs
@@ -18,7 +18,9 @@ namespace Nest
 		private readonly object _comparisonValue;
 		private readonly Type _type;
 
-		public Field(string name, double? boost = null, string format = null)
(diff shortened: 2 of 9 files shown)


Mpdreamz:
The build is failing on travis because we need to be explicit about langversion.

Something [already fixed on `6.5`](https://github.com/elastic/elasticsearch-net/commit/c7f462c740f6a7deefffd6457b8fb7f9e679c8d6), will retarget this PR against `6.5`.

src/Nest/CommonOptions/DateMath/DateMath.cs
```diff
@@ -79,10 +79,11 @@ public static DateMath FromString(string dateMath)
 			return math;
 		}
 
+		internal bool IsValid => Self.Anchor.Match(d => d != default, s => !s.IsNullOrEmpty());
+
 		public override string ToString()
 		{
-			var isValid = Self.Anchor.Match(d => d != default(DateTime), s => !s.IsNullOrEmpty());
-			if (!isValid) return string.Empty;
+			if (!IsValid) return string.Empty;
```
russcam: Should we have a separate `DateMath` test for this?

Mpdreamz: We should! Attached a unit test for this.

PR #3557, opened 2019-01-30
Use the version number present in the global.json file when generating documentation

codebrain:
This pull request addresses an issue whereby the documentation can be generated for a release with the incorrect version number. The version number is now calculated from the `global.json` file, thereby limiting the chances of the incorrect version number being used on a release.

Regular expressions and basic character manipulation has been used to avoid the need for JSON parsing and Semver libraries.

Backport to `6.x` on merge.


russcam:
ported to 6.x: 3b9cde872900d1ab76015fa92066f70fe3f2d171
ported to 5.x: 4e080471b2aeb923874371e8098c5faef6a2391a

PR #3571, opened 2019-02-12
Implement ML Info Api

codebrain:
Implemented referencing the YAML tests and server code.

Tests are asserted against the integration test configured cluster.

Requires porting to `master`.


Mpdreamz:
@codebrain mind retargetting this against `6.6` ?

src/Nest/XPack/MachineLearning/MlInfo/ElasticClient-MlInfo.cs
```diff
@@ -0,0 +1,58 @@
...
+using Elasticsearch.Net;
+
+namespace Nest
+{
+	public partial interface IElasticClient
+	{
+		/// <summary>
+		/// Retrieves machine learning job results for one or more categories.
+		/// </summary>
+		IMlInfoResponse MlInfo(Func<MlInfoDescriptor, IMlInfoRequest> selector = null);
```
russcam: My vote would be to use `MachineLearningInfo` here

codebrain: I had also thought the same thing, however all of the other ML APIs use the `ML` prefix, so we would be breaking consistency. My vote would be to keep as is, can we get a tie breaking vote from @mpdreamz?

russcam: The ML prefix is only in the specs though, not in the client source?

Mpdreamz: We do have vague plans to move all of the ml api's behind a `MachineLearning` property in 7.0 as well.

I'd prefer `MachineLearningInfo()` for 6.x too.

codebrain: Done, renamed as voted! Thanks!

codebrain:
Ported to `master`

PR #3579, opened 2019-02-28
Avoid doing string concat when not needed

Henr1k80:
I noticed this OOM exception and saw string concat in the call stack.
No need to do concat when we are putting the result into a JsonWriter anyway.

I can see that you have improved the serialization in other branches, but I guess that is not coming in the near future.

> Inner exception System.OutOfMemoryException handled at Elasticsearch.Net.Transport`1+<RequestAsync>d__15`1.MoveNext:
>    at System.String.Concat (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
>    at Nest.BulkRequestJsonConverter.WriteJson (Nest, Version=6.0.0.0, Culture=neutral, PublicKeyToken=96c599bbe3e70f5d)

### DIFF
--- a/src/Nest/Document/Multiple/Bulk/BulkRequestJsonConverter.cs
+++ b/src/Nest/Document/Multiple/Bulk/BulkRequestJsonConverter.cs
@@ -29,7 +29,11 @@ namespace Nest
 				op.Routing = op.GetRoutingForOperation(settings.Inferrer);
 
 				var opJson = requestResponseSerializer.SerializeToString(op, SerializationFormatting.None);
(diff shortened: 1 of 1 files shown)


Mpdreamz:
Thanks! Yes we are working on making major improvements but this a super quick win. 

The serialization improvements should land soon but for now target 7.x only.

Henr1k80:
This will not make into 6.6, right?
Do you have a release plan or is it just on demand when there is enough meaningful changes?

PR #3580, opened 2019-02-28
Avoid doing multiple dictionary lookups

Henr1k80:
Dictionary lookup involves get key hash, find bucket for hash and iterate bucket for key, so look the value up once, instead of checking if the key exists and then using the indexer to look up the value.
It also reduces the amount of duplicate hardcoded keys in the code.

Could not find any documentation for running the benchmark, so I haven't checked if it actually makes any difference. Haven't analyzed how often the code is executed.

Some places ContainsKey is used before Remove. That is not needed for Dictionary<>, as it will just return false for key not found, but I am uncertain of the specification for IDictionary<> void Remove, so I left it in place.

### DIFF
--- a/src/Elasticsearch.Net/Responses/Sniff/SniffResponse.cs
+++ b/src/Elasticsearch.Net/Responses/Sniff/SniffResponse.cs
@@ -78,8 +78,8 @@ namespace Elasticsearch.Net
 		{
 			get
 			{
-				if (settings != null && settings.ContainsKey("http.enabled"))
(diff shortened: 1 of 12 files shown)


russcam:
Thanks @Henr1k80 👍

russcam:
ported to 5.x: 641df53f5fa646648bc9cfb8d14d5ef11041f30a
ported to master: c159ddffe0592bcc4e38fe8270b11905a6da9c64

PR #3589, opened 2019-03-04
fix #3578 Add support for disabling id inference

### LINKED ISSUE #3578: Option to disable Id Inference
We're working on heavily optimizing our data ingest process to connect Loupe to Elasticsearch for our customers.  One problem we've run into is that we're getting a performance hit because virtually every object we have has a property called "Id" which is being automapped by NEST to be a document Id in Elastic and in turn meaning Elastic is evaluating these for uniqueness as we insert.

Don't get me wrong, in a world where all things are free we wouldn't care but a bit of experimentation has shown that this may be significantly slowing down ingestion rates when you get to the 30-50 million document size.  We'd like to disable this because we just don't need or want Elastic to treat this like a key but we haven't been able to come up with any way that allows us to have this property on our documents but *not* get it auto-mapped to __id.  

You could argue that we could make new objects that have this field renamed to bypass the mapper but we're all about that performance and taking objects and recreating new objects to put them in ES, then getting them out of ES and making new objects to convert them back, feels like a *lot* of burn and churn to bypass some automagic in the .NET client.

What we'd like is an option to disable the auto inference of this - ideally either on the .Settings() of CreateIndexDescriptor or the Map<T>() function.

Mpdreamz:


russcam:
@Mpdreamz if you're happy with the changes, I'll merge this one in

Mpdreamz:
Agree with the changes 👍

PR #3591, opened 2019-03-05
Delete forecast API

codebrain:


codebrain:
Also raised an issue here: https://github.com/elastic/elasticsearch/issues/39673

src/CodeGeneration/ApiGenerator/RestSpecification/XPack/MachineLearning/xpack.ml.delete_forecast.patch.json
```diff
@@ -0,0 +1,16 @@
...
+  "xpack.ml.delete_forecast": {
+    "url": {
+      "path": "/_xpack/ml/anomaly_detectors/{job_id}/_forecast/{forecast_id}",
+      "paths": [
+        "/_xpack/ml/anomaly_detectors/{job_id}/_forecast/{forecast_id}"
+      ],
+      "parts": {
+        "forecast_id": {
+          "required": true,
+          "description": "The ID of the forecast to delete, can be comma delimited list or `_all`"
```
russcam: Since this can be a comma delimited list, this should maybe map to `Ids` or another, new type?

codebrain: Good point, have implemented a custom type, see: https://github.com/elastic/elasticsearch-net/pull/3591/commits/06474b14a41ed4923adfbf36d496863adb522469

src/Nest/CommonAbstractions/Infer/ForecastIds/ForecastIds.cs
```diff
@@ -0,0 +1,49 @@
...
+		string IUrlParameter.GetString(IConnectionConfigurationValues settings) => string.Join(",", _forecastIds);
+
+		public static implicit operator ForecastIds(string forecastIds) =>
+			forecastIds.IsNullOrEmptyCommaSeparatedList(out var list) ? null : new ForecastIds(list);
+
+		public static implicit operator ForecastIds(string[] forecastIds) => forecastIds.IsEmpty() ? null : new ForecastIds(forecastIds);
+
+		public override bool Equals(object obj) => obj is ForecastIds other && Equals(other);
+
+		public override int GetHashCode() => _forecastIds.GetHashCode();
```
russcam: should hashcode be calculated from the sorted contents of `_forecastIds`?

src/Nest/CommonAbstractions/Infer/ForecastIds/ForecastIds.cs
```diff
@@ -0,0 +1,49 @@
...
+
+			return Ids.Count == other.Ids.Count && !Ids.Except(other.Ids).Any();
+		}
+
+		string IUrlParameter.GetString(IConnectionConfigurationValues settings) => string.Join(",", _forecastIds);
+
+		public static implicit operator ForecastIds(string forecastIds) =>
+			forecastIds.IsNullOrEmptyCommaSeparatedList(out var list) ? null : new ForecastIds(list);
+
+		public static implicit operator ForecastIds(string[] forecastIds) => forecastIds.IsEmpty() ? null : new ForecastIds(forecastIds);
```
russcam: This should have the same semantics as using the constructor. With an empty `forecastIds`, this returns `null`, but the constructor assigns a `new List<string>()` to the backing field. Should both ways of constructing an instance of `ForecastIds` result in the same outcome?

src/Nest/CommonAbstractions/Infer/ForecastIds/ForecastIds.cs
```diff
@@ -0,0 +1,49 @@
...
+		{
+			if (Ids == null && other.Ids == null) return true;
+			if (Ids == null || other.Ids == null) return false;
+
+			return Ids.Count == other.Ids.Count && !Ids.Except(other.Ids).Any();
+		}
+
+		string IUrlParameter.GetString(IConnectionConfigurationValues settings) => string.Join(",", _forecastIds);
+
+		public static implicit operator ForecastIds(string forecastIds) =>
```
russcam: This should have the same semantics as using the constructor. With a null or empty `forecastIds`, this returns `null`, but the constructor assigns a `new List<string>()` to the backing field. Should both ways of constructing an instance of `ForecastIds` result in the same outcome?

src/Nest/CommonAbstractions/Infer/ForecastIds/ForecastIds.cs
```diff
@@ -0,0 +1,49 @@
...
+		public static implicit operator ForecastIds(string[] forecastIds) => forecastIds.IsEmpty() ? null : new ForecastIds(forecastIds);
+
+		public override bool Equals(object obj) => obj is ForecastIds other && Equals(other);
+
+		public override int GetHashCode() => _forecastIds.GetHashCode();
+
+		public static bool operator ==(ForecastIds left, ForecastIds right) => Equals(left, right);
+
+		public static bool operator !=(ForecastIds left, ForecastIds right) => !Equals(left, right);
+	}
```
russcam: Can we add a static `All` property that returns `"_all"`?

src/Nest/CommonAbstractions/Infer/ForecastIds/ForecastIds.cs
```diff
@@ -0,0 +1,49 @@
...
+
+		public ForecastIds(IEnumerable<string> forecastIds) => _forecastIds = forecastIds?.ToList() ?? new List<string>();
+
+		public ForecastIds(string forecastIds) => _forecastIds = forecastIds.IsNullOrEmpty()
+			? new List<string>()
+			: forecastIds.Split(new[] { "," }, StringSplitOptions.RemoveEmptyEntries)
+				.Select(s => s.Trim())
+				.ToList();
+
+		internal IReadOnlyList<string> Ids => _forecastIds;
```
russcam: I don't think this is necessary. But if it is, I think making the property type `List<string>` would be preferable to accessing through the interface for an internal property.

src/Nest/CommonAbstractions/Infer/ForecastIds/ForecastIds.cs
codebrain: Yep, good catch, will sort first to ensure repeatability.

src/Nest/CommonAbstractions/Infer/ForecastIds/ForecastIds.cs
codebrain: Very good question. This implementation was copy-pasta'd from the `ActionIds` type. My suspicion is that all "null or empty" initialisation should result in an empty list, as per the constructor implementation.

@Mpdreamz - is this assumption correct? If no, I can fix here and submit a patch for `ActionIds`

src/Nest/CommonAbstractions/Infer/ForecastIds/ForecastIds.cs
codebrain: Also applies to the `ActionIds` type.

src/Nest/CommonAbstractions/Infer/ForecastIds/ForecastIds.cs
russcam: I think `ActionIds` should be updated too.

src/Nest/CommonAbstractions/Infer/ForecastIds/ForecastIds.cs
codebrain: ++

src/Nest/CommonAbstractions/Infer/ForecastIds/ForecastIds.cs
codebrain: Required for the equality comparison.

src/Nest/CommonAbstractions/Infer/ForecastIds/ForecastIds.cs
Mpdreamz: +1 on also updating `ActionIds`

codebrain: Will change to use `new List<string>()` instead of `null` values.

russcam:
I've updated the `ForecastIds` implementation to:

- implement `GetHashCode()` based on the sorted contents
- remove internal `IReadOnlyList<string>` property
- `All` static property initializer
- Remove the conditional operators in implicit operator conversions into the ctors

PR #3583, opened 2019-03-04
Replace internal serializer with utf8json

russcam:
This PR replaces both the SimpleJson (Elasticsearch.Net) and JSON.Net (in NEST) based internal serializers with a serializer based on [utf8json](https://github.com/neuecc/Utf8Json).

The serializer source has been included in Elasticsearch.Net to allow 
- further modification for the specific need of the client
- mark all serializer types as `internal`

Research as part of #3493 demonstrates an immediate 2x performance improvement in serialization, and it is expected that this could be further improved by

- using `Span<T>`, `Memory<T>` where possible
- reusing `ArrayPool<T>` 
- cleaning and simplifying the interface between serializer and clients, now that the serializer source is included.

All unit tests pass in this PR. One test, `ExtendingNestTypes`, is currently skipped until a general serialization routine is in place. This PR is opened now because there are further changes coming related to 7.x, so the preference is to merge this work in now and fix the remaining failing integration tests which at point of opening are:

```
Failed collections:
 - ReadOnly: Tests.Aggregations.Bucket.Composite.CompositeAggregationUsageTests
 - ReadOnly: Tests.Aggregations.Metric.TopHits.TopHitsAggregationUsageTests
 - ReadOnly: Tests.Search.Request.NestedSortUsageTests
 - ReadOnly: Tests.Search.Request.SearchAfterUsageTests
 - ReadOnly: Tests.Search.Request.SortUsageTests
 - TimeSeries: Tests.XPack.Rollup.RollupJobCrudTests
 - XPack: Tests.XPack.License.GetTrialLicenseStatus.GetTrialLicenseStatusApiTests
 - XPack: Tests.XPack.License.StartTrialLicense.StartTrialLicenseApiTests
 - XPack: Tests.XPack.Migration.MigrationUpgrade.InvalidMigrationUpgradeApiTests
---Reproduce: -----
build.bat seed:51817 integrate 6.4.2 "timeseries,xpack,readonly" "compositeaggregation,tophitsaggregation,nestedsort,searchafter,sort,gettriallicensestatus,starttriallicense,invalidmigrationupgrade,rollupjobcrud"
--------
=== TEST EXECUTION SUMMARY ===
   Tests  Total: 1458, Errors: 0, Failed: 29, Skipped: 13, Time: 689.411s
```

PR #3592, opened 2019-03-11
Add fuzzy auto length support for NEST

SukharevAndrey:
As of now fuzziness AUTO parameter does not support custom length behaviour added in https://github.com/elastic/elasticsearch/issues/25528 and described on https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#fuzziness

This PR contains simple implementation for that to NEST. It could be easily ported to the new JSON serialization implementation: https://github.com/elastic/elasticsearch-net/pull/3583

### DIFF
--- a/src/Nest/CommonOptions/Fuzziness/Fuzziness.cs
+++ b/src/Nest/CommonOptions/Fuzziness/Fuzziness.cs
@@ -3,12 +3,22 @@
 	public class Fuzziness : IFuzziness
 	{
 		private bool _auto;
+		private int? _low;
+		private int? _high;
 		private int? _editDistance;
 		private double? _ratio;
 
 		public static Fuzziness Auto => new Fuzziness { _auto = true };
+		public static Fuzziness AutoLength(int low, int high) => new Fuzziness
+		{
+			_auto = true,
+			_low = low,
+			_high = high
+		};
 
 		bool IFuzziness.Auto => _auto;
+		int? IFuzziness.Low => _low;
+		int? IFuzziness.High => _high;
 		int? IFuzziness.EditDistance => _editDistance;
 		double? IFuzziness.Ratio => _ratio;
 
--- a/src/Nest/CommonOptions/Fuzziness/Fuzziness.cs
+++ b/src/Nest/CommonOptions/Fuzziness/Fuzziness.cs
@@ -8,7 +8,19 @@
 		private int? _editDistance;
 		private double? _ratio;
 
+		/// <summary>
+		/// Generates an edit distance based on the length of the term.
+		/// </summary>
+		/// <remarks>
+		/// <para>Equivalent to <see cref="AutoLength"/> with parameters 3 and 6.</para>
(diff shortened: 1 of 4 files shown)


src/Nest/CommonOptions/Fuzziness/FuzzinessJsonConverter.cs
```diff
@@ -1,19 +1,28 @@
 ﻿using System;
+using System.Text.RegularExpressions;
 using Newtonsoft.Json;
 
 namespace Nest
 {
 	internal class FuzzinessJsonConverter : JsonConverter
 	{
+		private static readonly Regex AutoLengthRegex = new Regex(@"^AUTO:(?<low>\+?\d+),(?<high>\+?\d+)$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
```
russcam: I suspect looking for the index of `:` and `,` will be faster than a regex here.

SukharevAndrey:
Thank you for your feedback, @russcam.
Changed the implementation, and it is now more than 6 times faster :)
Also added docs for parameters based on official documentation and included usage of AutoLength in  MatchUsageTests.

src/Nest/CommonOptions/Fuzziness/FuzzinessJsonConverter.cs
SukharevAndrey: Changed. it is now more simple, but may throw an exception when source string is malformed (for example when ```,``` is before ```;```

src/Nest/CommonOptions/Fuzziness/Fuzziness.cs
```diff
@@ -3,15 +3,41 @@
...
+		/// <para>Equivalent to <see cref="AutoLength"/> with parameters 3 and 6.</para>
+		/// <para><see cref="Auto"/> should generally be the preferred value for <see cref="Fuzziness"/></para>
+		/// </remarks>
 		public static Fuzziness Auto => new Fuzziness { _auto = true };
+		/// <summary>
+		/// Generates an edit distance based on the length of the term.
+		/// </summary>
+		/// <param name="low">Must match exactly for terms with less length</param>
+		/// <param name="high">Two edits allowed for terms with greater length</param>
+		public static Fuzziness AutoLength(int low, int high) => new Fuzziness
```
russcam: Documentation 💯

src/Nest/CommonOptions/Fuzziness/FuzzinessJsonConverter.cs
russcam: I think this is acceptable here, since Elasticsearch constrains the format of the response:

https://github.com/elastic/elasticsearch/blob/3b8c2e4d8e221677324753cd3265cc111d0faa95/server/src/main/java/org/elasticsearch/common/unit/Fuzziness.java#L69-L77

russcam:
ported to 6.x: 167fa29e34b50d853d11f709e517a1630b84662c

PR #3596, opened 2019-03-13
 Throw a parsing exception when boost is set in span_or query,

codebrain:
Relates to elastic/elasticsearch#28390 and elastic/elasticsearch#34112


src/Nest/QueryDsl/Abstractions/Query/QueryBase.cs
```diff
@@ -9,6 +9,7 @@ public interface IQuery
 		/// Provides a boost to this query to influence its relevance score.
 		/// For example, a query with a boost of 2 is twice as important as a query with a boost of 1,
 		/// although the actual boost value that is applied undergoes normalization and internal optimization.
+		/// Setting a boost with a span_or query will throw a parsing exception on the server.
```
russcam: Should we use `<see ref="ISpanQuery" />` instead of `span_or` query?

PR #3598, opened 2019-03-13
Expose BulkResponseItems on BulkAllResponse

### LINKED ISSUE #3487: BulkAll Return Response with document _id's
**NEST/Elasticsearch.Net version**: 6.3.1

**Elasticsearch version**: 6.3

**Describe the feature**:

I am using the BulkAll interface to index a group of documents. I would like Elastic to Create the _id field for these documents so I do not specify it in my request. I would like to know the _id of the documents indexed, like the response of the index API or the [standard bulk API](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html). Could the response from Elastic be included in the [BulkAllResponse](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Document/Multiple/BulkAll/BulkAllResponse.cs)?

I am thinking that all that needs be done is modify that class to include a response. Then include the successful response from Elastic [here](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Document/Multiple/BulkAll/BulkAllObservable.cs#L152). When onNext is called in the subscriber the response for items included in that bulk update will be shown.

I could be completely off base with a solution / should I just write my own version of BulkAll with the functionality that I desire?

russcam:
This commit exposes the bulk response items on the BulkAllResponse.

Closes #3487

### DIFF
--- a/src/Nest/Document/Multiple/BulkAll/BulkAllObservable.cs
+++ b/src/Nest/Document/Multiple/BulkAll/BulkAllObservable.cs
@@ -112,18 +112,18 @@ namespace Nest
 		{
 			_compositeCancelToken.ThrowIfCancellationRequested();
 
-			var r = _partitionedBulkRequest;
+			var request = _partitionedBulkRequest;
 			var response = await _client.BulkAsync(s =>
 				{
-					s.Index(r.Index).Type(r.Type);
-					if (r.BufferToBulk != null) r.BufferToBulk(s, buffer);
+					s.Index(request.Index).Type(request.Type);
+					if (request.BufferToBulk != null) request.BufferToBulk(s, buffer);
 					else s.IndexMany(buffer);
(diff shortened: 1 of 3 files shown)


src/Nest/Document/Multiple/BulkAll/BulkAllObservable.cs
```diff
@@ -185,7 +194,7 @@ private async Task<IBulkAllResponse> HandleBulkRequest(IList<T> buffer, long pag
 					throw ThrowOnBadBulk(response,
 						$"BulkAll halted after {nameof(PipelineFailure)}{failureReason.GetStringValue()} from _bulk");
 				default:
-					return await RetryDocuments(page, ++backOffRetries, buffer);
+					return await RetryDocuments(page, ++backOffRetries, buffer).ConfigureAwait(false);
```
codebrain: 👍

PR #3603, opened 2019-03-14
Add node count to ML usage response

codebrain:
Relates to: https://github.com/elastic/elasticsearch/pull/33863


src/Nest/XPack/Info/XPackUsage/XPackUsageResponse.cs
```diff
@@ -164,6 +164,9 @@ public class MonitoringUsage : XPackUsage
 
 	public class MachineLearningUsage : XPackUsage
 	{
+		[JsonProperty("node_count")]
+		public int NodeCount { get; internal set; }
```
russcam: Can we add `<remarks>Valid only for Elasticsearch 6.5.0+</remarks>`?

PR #3609, opened 2019-03-18
Add Parent Aggregation

russcam:
Relates: elastic/elasticsearch#34210

This commit adds Parent Aggregation to the high level client.

### DIFF
--- a/src/Nest/Aggregations/Visitor/AggregationVisitor.cs
+++ b/src/Nest/Aggregations/Visitor/AggregationVisitor.cs
@@ -66,6 +66,8 @@
 
 		void Visit(INestedAggregation aggregation);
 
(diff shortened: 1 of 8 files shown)


src/Nest/Aggregations/Visitor/AggregationVisitor.cs
```diff
@@ -149,6 +151,7 @@ public class AggregationVisitor : IAggregationVisitor
 
 		public virtual void Visit(ITermsAggregation aggregation) { }
 
+		// TODO: make virtual
```
codebrain: TODO?

russcam: yep, can't change it in 6.x as it would break BWC

PR #3616, opened 2019-03-19
Avoid copying byte content to new array. Use buffer directly

Henr1k80:


elasticcla:
Hi @Henr1k80, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-net/pull/3616.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

Henr1k80:
Integration tests seems to be failing even without my changes

russcam:
Hey @Henr1k80, thanks for another PR! 

I think @Mpdreamz is working on the Azure Pipelines config, so don't worry about them for now. Rest assured, we'll run them before merging in 😃

Henr1k80:
I have considered obsoleting
`public static bool TryDeserializeObject(string json, out object obj)`
and other string methods in favor of the char[] methods, but they are public/internal and I do not know where they could be used and how intrusive that change is.

Mpdreamz:
It's been a while since I wrote those routines but was aware of `GetBuffer()` back then too, I had to reread a bit why I didn't use this technique everywhere to begin with. 

The buffer does not always start at position `0` (more true on our current `7.x` branch now we moved to utf8json then `6.x`) and `GetBuffer()` can throw if the [`MemoryStream` was constructed in such a way that `_exposable` is false](https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/IO/MemoryStream.cs#L40). 

At the time of writing most of these routines we also accepted streams directly so we had little control over this. 

I think most of these qualms no longer apply but will need some more time to further validate this.

Mpdreamz:
Hey @Henr1k80 this almost LGTM 👍 

I would like to make sure the utility methods are more defensive against the cases I just listed. Took a stab at it here:

https://github.com/elastic/elasticsearch-net/commit/660971429e87a0fdad24ecc840369f9e98724bc9

Mind incorporating these changes into your PR as well?

Henr1k80:
Ah, didn't know that it could fail. Of course your changes should be incorporated.

PR #3624, opened 2019-03-25
V7.x/integration

Mpdreamz:
Initial stab at getting the integration tests to pass against 7.0.0-beta1

Those tests that warrant a deeper investigation are muted as part of this PR.

They are either marked as:

`SkipNonStructuralChange` if the failure was not related to a structural json change in request and response but due to a functional change in Elasticsearch. Examples might be painless script changes. 

`BlockedByVersion(url)` this means the test can not run because of a presumed issue with Elasticsearch 7.0.0-beta1 that we've reported and waiting to hear back on.

Both attributes will need to be handled before we can release.


Mpdreamz:
travis is red due to SDK bootstrapping issues. We're moving away from travis though

PR #3629, opened 2019-03-28
Set the timeout on the bulk call to use the request timeout

codebrain:


codebrain:
Will fix in 6.7 branch instead.

PR #3631, opened 2019-03-29
Remove delegate allocations in Fluent.Assign

russcam:
This commit removes delegate allocations when
using Fluent.Assign() by introducing a TValue parameter
and changing the action delegate to also take TValue. In doing so,
no closure over the assigned value is created, removing the
construction of a new delegate in the process.

Some methods such as on BulkDescriptor.CreateMany<T>() no longer
use Fluent.Assign() to avoid the allocation and because they do not fit
a generic Fluent.Assign signature without introducing a closure.


russcam:
There's a few places I've missed in this refactoring. Adding those

PR #3636, opened 2019-04-01
Refactor/build scripts

Mpdreamz:
This PR refactors the way we run build scripts.

Our build script project is now a regular fsharp .NET core project that compiles with the rest of the projects and can be run with the dotnet CLI. 

#### This is a move away from FAKE as a runner

* FAKE5 needs a global CLI tool which is not a nice way to do reproducible builds
* FAKE relies on scripts being run through `Yaaf.FSharp.Scripting`, our build is sufficiently complex the compiler errors from here are not always great. 
* FAKE as a runner is also a tad slow since it goes through `Yaaf.FSharp.Scripting` 

FAKE5 is highly modular we still take advantage of many modules available here!

#### This is also a move away from Paket:

* I love Paket, I strongly believe in its defaults and design philosophy and feature set.
* It's dotnet core integration is a bit hit and miss (anecdotal I know)
* We only used Paket for FAKE dependencies and given the choice to move to all in Paket or all in Nuget the path of least resistence is nuget in the new SDK style project world.

#### This introduces BullsEye

* Simple library that declares targets with dependencies that does just enough to be useful, gets out of the way and does the right things OOTB.

#### This removes all checked in deps

This removes all checked in dependencies and pulls everything in through nuget. This was not possible for a long time but now Nuget4.9 makes the path to nuget packages on disk available allowing us to make use of packages distributing artifacts under `tools` again.

Our `assembly-differ` and `assembly-rewriter` are now distributed as `DotNetCoreCliTools` until .NET 3.0 supports project local tool installation again.

#### Refactor build scripts

The build scripts have been refactored quite a bit a long the way as well. Now that they are compiled as `.fs` files in a regular `netcoreapp2.2` application. I am sure many places can be refactored further but best leave that to more isolated PR's.


build.bat
```diff
@@ -1,6 +1,2 @@
 @echo off
-
-.paket\paket.bootstrapper.exe 
-IF EXIST paket.lock (.paket\paket.exe restore)
-IF NOT EXIST paket.lock (.paket\paket.exe install)
-"packages\build\FAKE\tools\Fake.exe" "build\\scripts\\Targets.fsx" "cmdline=%*"
+dotnet run --project build/scripts -- %*
```
russcam: `-c Release` mode?

build.sh
```diff
@@ -1,8 +1,3 @@
 #!/usr/bin/env bash
-FAKE="packages/build/FAKE/tools/FAKE.exe"
-BUILDSCRIPT="build/scripts/Targets.fsx"
-
-mono .paket/paket.bootstrapper.exe
-if [[ -f paket.lock ]]; then mono .paket/paket.exe restore; fi
-if [[ ! -f paket.lock ]]; then mono .paket/paket.exe install; fi
-mono $FAKE $BUILDSCRIPT "cmdline=$*" --fsiargs -d:MONO 
+set -euo pipefail
+dotnet run --project build/scripts -- "$@"
```
russcam: `-c Release` mode?

build.bat
adamralph: I guess you could do that, but I've never seen anyone do it. After all, the `build/scripts` project will never be "released". It's effectively a script.

build/scripts/scripts.fsproj
```diff
@@ -1,26 +1,90 @@
...
   </ItemGroup>
   <ItemGroup>
-    <Content Include="..\..\paket.bat"><Link>paket.bat</Link></Content>
-    <Content Include="..\..\paket.dependencies"><Link>paket.bat</Link></Content>
-    <Content Include="..\..\build.sh"><Link>paket.bat</Link></Content>
-    <Content Include="..\..\build.bat"><Link>paket.bat</Link></Content>
+    <Content Include="..\..\build.sh"><Link>build.sh</Link></Content>
+    <Content Include="..\..\build.bat"><Link>build.bat</Link></Content>
   </ItemGroup>
+  <!-- Type providers currently can't run inside the .NET Core 2.0 hosted compiler, see https://github.com/Microsoft/visualfsharp/pull/3658#issuecomment-334773415 -->
```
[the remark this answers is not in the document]
Mpdreamz: FWIW if you have the absolute lastest mono and .NET SDK type providers work like a charm on ubuntu/osx. Pulling these in on azure pipelines took 16 minutes 😢

PR #3637, opened 2019-04-03
Adds the x-pack REST API downloads into the client generator.

codebrain:


src/CodeGeneration/ApiGenerator/ApiGenerator.csproj
```diff
@@ -18,4 +18,8 @@
     <PackageReference Include="RazorLight.Unofficial" Version="2.0.0-beta1.1" />
     <!--<PackageReference Include="RazorLight" Version="2.0.0-beta1" />-->
   </ItemGroup>
+  <ItemGroup>
```
russcam: will be implicitly added - these can be removed

src/CodeGeneration/ApiGenerator/Program.cs
```diff
@@ -39,8 +39,23 @@ private static void Main(string[] args)
...
+				"Indices",
+				"Info",
+				"License",
+				"MachineLearning",
+				"Migration",
+				"Monitoring",
+				"Rollup",
+				"Security",
+				"Sql",
+				"Sql",
```
russcam: one too many "Sql"

russcam:
#3638 should be merged in after this as it fixes some of the issues introduced in the new REST API specs.

PR #3644, opened 2019-04-04
Change to StaticConnectionPool for Multiple Hosts

huzefakagdi:
Change Example for Connection to Multiple Elastic Sever Hosts, which are known already, to use StaticConnectionPool instead of SniffingConnectionPool. Using SniffingConnectionPool was giving Null Pointer Exception in the method RequestPipeline.CreateClientException(). I was not able to pinpoint the parameter which was null.

### DIFF
--- a/docs/client-concepts/low-level/getting-started.asciidoc
+++ b/docs/client-concepts/low-level/getting-started.asciidoc
@@ -55,7 +55,7 @@ There are many other <<configuration-options,configuration options>> on `Connect
 
 `ConnectionConfiguration` is not restricted to being passed a single address for Elasticsearch. There are several different
 types of <<connection-pooling,Connection pool>> available, each with different characteristics that can be used to
-configure the client. The following example uses a <<sniffing-connection-pool,SniffingConnectionPool>> seeded with the addresses
+configure the client. The following example uses a <<static-connection-pool,StaticConnectionPool>> seeded with the addresses
 of three Elasticsearch nodes in the cluster, and the client will use this type of pool to maintain a list of available nodes within the
 cluster to which it can send requests in a round-robin fashion.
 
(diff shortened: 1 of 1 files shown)


Mpdreamz:
@huzefakagdi I much rather solve the `NRE` do you have a stack trace when this happens?

huzefakagdi:
Hi Martijn,

The relevant stacktrace for the NRE when we use StickyConnectionPool to create ElasticLowLevelClient is as below:

`   at Elasticsearch.Net.RequestPipeline.CreateClientException[TResponse](TResponse response, IApiCallDetails callDetails, RequestData data, List`1 pipelineExceptions)
   at Elasticsearch.Net.Transport`1.FinalizeResponse[TResponse](RequestData requestData, IRequestPipeline pipeline, List`1 seenExceptions, TResponse response)
   at Elasticsearch.Net.Transport`1.Request[TResponse](HttpMethod method, String path, PostData data, IRequestParameters requestParameters)
   at Elasticsearch.Net.ElasticLowLevelClient.DoRequest[TResponse](HttpMethod method, String path, PostData data, IRequestParameters requestParameters)
   at Elasticsearch.Net.ElasticLowLevelClient.Index[TResponse](String index, String type, PostData body, IndexRequestParameters requestParameters)`

The code that we are using is similar to the example in the documentation. The only difference is that we are calling the BasicAuthentication on the ConnectionConfiguration that is created.
The only change in code required to get rid of the exception is to change the connection pool to StaticConnectionPool.

We are using the Elasticsearch.Net version 6.0.0.

Mpdreamz:
If you update to the latest version of NEST/Elasticsearch.NET `6.6.0` is the problem still there?

huzefakagdi:
I updated to the latest version on my local machine and we do not get the error for SniffingConnectionPool. It seems that there were issues with the older package. But unfortunately we cannot update to v6.6 in our actual code as we will need to upgrade to .NET Framework v4.6.1 for that.

Thanks for your help. You may reject the PR.

Mpdreamz:
https://www.nuget.org/packages/NEST/6.2.0 was our latest version on 4.5 hopefully that contains the fix for it as well @huzefakagdi thank you for bringing it to our attention :+1:

PR #3649, opened 2019-04-08
Refactor/test version config

Mpdreamz:
- Simplifies `ITestConfiguration` and its implementations. 
- Default version resolutions is now driven by the yaml file rather than also having baked in fallback in code.

### DIFF
--- a/build/scripts/Testing.fs
+++ b/build/scripts/Testing.fs
@@ -24,7 +24,6 @@ module Tests =
         env "NEST_INTEGRATION_CLUSTER" clusterFilter
         env "NEST_TEST_FILTER" testFilter
         env "NEST_TEST_SEED" (Some <| args.Seed)
-        env "NEST_COMMAND_LINE_BUILD" <| Some "1"
 
         for random in args.RandomArguments do 
             let tokens = random.Split [|':'|]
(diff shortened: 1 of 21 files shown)


build/scripts/Testing.fs
```diff
@@ -54,7 +53,10 @@ module Tests =
             | (true) -> [ "--logger"; "trx"; "--collect"; "\"Code Coverage\""; "-v"; "m"] |> List.append command
             | _  -> command
             
-        Tooling.DotNet.ExecInWithTimeout "src/Tests/Tests" commandWithCodeCoverage (TimeSpan.FromMinutes 30.) 
+        // using std redirection since that foces vstest not to redirect Console.Write from Elastic.Xunit somehow
```
russcam: Should this comment be updated with how vstest and xunit communicate?

src/Tests/Tests.Configuration/TestConfigurationBase.cs
```diff
@@ -1,22 +1,70 @@
...
+		/// <summary> whether the seed was provided externally to be fixated </summary>
+		public bool SeedProvidedExternally { get; private set; }
+
+		/// <summary>
+		/// This is fixed for now, specifying false leads to flaky tests, warrants deeper investigation
+		/// in our abstractions project
+		/// </summary>
+		public bool ShowElasticsearchOutputAfterStarted { get; } = true;
+
+		/// <summary> When specified will only run one overload in API tests, helpful when debugging locally </summary>
```
russcam: comments 💯

build/scripts/Testing.fs
Mpdreamz: reverting this since `Proc` is not happy on linux

Mpdreamz:
ported to master

PR #3651, opened 2019-04-08
Docs: Fix the delimiter on a code block

nik9000:
Fixes the delimiter on a code block that Asciidoctor is not a fan of.
This should let us convert all of these docs from the unmaintained
AsciiDoc project to the actively maintained Asciidoctor project.


nik9000:
This only hits the 1.x branch. All of the others are great.

PR #3653, opened 2019-04-10
Use ContractJsonConverterAttribute where possible

russcam:
Use ContractJsonConverterAttribute where possible, and short-circuit `JsonContract` creation when a custom converter is found.

### DIFF
--- a/src/Nest/CrossPlatform/TypeExtensions.cs
+++ b/src/Nest/CrossPlatform/TypeExtensions.cs
@@ -8,7 +8,6 @@ namespace Nest
 {
 	internal static class DotNetCoreTypeExtensions
 	{
-		[MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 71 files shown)


src/Nest/CrossPlatform/TypeExtensions.cs
```diff
@@ -16,6 +18,7 @@ internal static bool IsGeneric(this Type type)
 #endif
 		}
 
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
```
Mpdreamz: Curious what triggered this change?

russcam: I should have looked at the IL before making this change....it appears to make no difference. Will revert 👍

russcam: reverted

PR #3660, opened 2019-04-11
Serialize fractional numbers with trailing decimal point and value

### LINKED ISSUE #3657: Double values should always be serialized with decimal point
With the move to utf8json in #3493, floating point and decimal numbers that do not have fractional values are not serialized with decimal point by utf8json. This saves a few bytes, but would break Elasticsearch type mapping inference. Floating point and decimals should always be sent with a decimal point.

russcam:
This commit updates

- the DoubleToStringConverter from utf8json to
always emit a trailing decimal point and trailing decimal value when
serializing floating point numbers to JSON.
- the DecimalFormatter from utf8json to always emit a trailing decimal
point and trailing decimal value when serializing floating point numbers to JSON.

Update unit tests to reflect this change

Fixes #3657


src/Elasticsearch.Net/Utf8Json/Internal/DoubleConversion/DoubleToStringConverter.cs
```diff
@@ -214,7 +215,7 @@ enum Flags
         //const int max_leading_padding_zeroes_in_precision_mode_;
         //const int max_trailing_padding_zeroes_in_precision_mode_;
 
-        static readonly Flags flags_ = Flags.UNIQUE_ZERO | Flags.EMIT_POSITIVE_EXPONENT_SIGN;
+        static readonly Flags flags_ = Flags.UNIQUE_ZERO | Flags.EMIT_POSITIVE_EXPONENT_SIGN | Flags.EMIT_TRAILING_DECIMAL_POINT | Flags.EMIT_TRAILING_ZERO_AFTER_POINT;
```
Mpdreamz: So happy these flags exist and problem is solved at the root!

PR #3661, opened 2019-04-11
Use float? for suggester fractional properties

russcam:
This commit updates TermSuggester to use float? instead of decimal? for
properties that take a float value.

Add XML documentation to suggesters


src/Nest/Mapping/Types/Specialized/Attachment/Attachment.cs
```diff
@@ -5,6 +5,13 @@
 
 namespace Nest
 {
+	/// <summary>
```
Mpdreamz: `An indexed Attachement` feels like a SandCastle comment :cat: 

`<remarks>` should be used as extra documentation  about `why` something exists or a particular implementation inside the class was chosen. `<summary>` shows up in intellisense. Here I feel the remarks should be hoisted into the summary.

src/Nest/Search/Suggesters/CompletionSuggester/CompletionField.cs
```diff
@@ -4,16 +4,29 @@
...
 	/// <summary>
-	/// Convenience class for use when indexing completion fields.
+	/// A completion field
 	/// </summary>
+	/// <remarks>
+	/// Convenience class for use when indexing completion fields.
+	/// </remarks>
 	public class CompletionField
 	{
+		/// <summary>
```
Mpdreamz: Hate the hate on adding xmldocs but not fan of these style of SandCastle'esque summaries that feel generated. I rather have none instead. 

Further more to continue my nitpick for small summaries can we have a single line

```chsarp
//// <summary> The summary </summary>
```

src/Nest/Search/Suggesters/CompletionSuggester/CompletionField.cs
```diff
@@ -4,16 +4,29 @@
 namespace Nest
 {
 	/// <summary>
-	/// Convenience class for use when indexing completion fields.
+	/// A completion field
 	/// </summary>
+	/// <remarks>
```
Mpdreamz: Merge remarks and summary

src/Nest/Search/Suggesters/CompletionSuggester/FuzzySuggest.cs
```diff
@@ -3,35 +3,61 @@
 
 namespace Nest
 {
+	/// <summary>
```
Mpdreamz: `FuzzySugester` is a bad name now that I am rereading it maybe `SuggesterFuzziness`?

Maybe link to https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters-completion.html#fuzzy ?

The comment itself as it is now adds little IMO.

russcam: ++ to `SuggesterFuzziness`; it better fits what it is.

I share some of the sentiment of GhostDoc/Sandcastle autogenerated XML comments. Do you think we should aim for consistency in applying `<summary>`?

src/Nest/Search/Suggesters/CompletionSuggester/CompletionField.cs
russcam: > for small summaries can we have a single line

I can do this. We'll also need to tweak our formatting configuration which always seems to format on multiple lines

src/Nest/Search/Suggesters/CompletionSuggester/FuzzySuggest.cs
russcam: Looking at the naming convention of other types, `ISuggestFuzziness` might be more appropriate

russcam:
I'm merging this in, to look to get a release out

PR #3662, opened 2019-04-11
Low level client exception structure should match SimpleJson strategy

### LINKED ISSUE #3656: Low level client exception structure should match SimpleJson
With the move to utf8json in #3493, we should ensure that the serialized structured of an exception in the low level client matches that of SimpleJson, the preceding low level client serializer.

This may also be an opportunity to create a better exception structure, to support indexing into Elasticsearch and other logging systems.

russcam:
This commit introduces an ExceptionFormatters that serializes exceptions to the same structure as that used in 6.x

Fixes #3656

### DIFF
--- a/src/Tests/Tests/Framework/SerializationTests/ExceptionSerializationTests.cs
+++ b/src/Tests/Tests/Framework/SerializationTests/ExceptionSerializationTests.cs
@@ -1,0 +1,71 @@
+using System;
+using Elastic.Xunit.XunitPlumbing;
+using Elasticsearch.Net;
+using FluentAssertions;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Tests.Framework.SerializationTests
+{
+	public class ExceptionSerializationTests
+	{
+		private readonly IElasticsearchSerializer _elasticsearchNetSerializer;
+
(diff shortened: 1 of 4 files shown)


src/Tests/Tests/Framework/SerializationTests/ExceptionSerializationTests.cs
```diff
@@ -0,0 +1,71 @@
...
+			{
+				CreateException(typeof(Exception), "outer_exception", 0),
+				CreateException(typeof(InnerException), "inner_exception", 1),
+				CreateException(typeof(InnerInnerException), "inner_inner_exception", 2),
+			};
+
+			var jArray = JArray.Parse(serialized);
+			var jArray2 = JArray.Parse(JsonConvert.SerializeObject(simpleJsonException));
+
+			JToken.DeepEquals(jArray, jArray2).Should().BeTrue();
```
Mpdreamz: Could utilize `SerializationTester.Serializes` which does the same but in case of a mismatch sorts the json and generates a nice diff exception message.

russcam: I did look at `SerializationTester`. It is heavily tied to serializing/deserializing with the client serializer, whereas this test needs to serialize with the low level client serializer

PR #3665, opened 2019-04-12
Remove superfluous pragma directives

russcam:
This PR removes superfluous pragma directives for `DOTNETCORE` and tidies up some superfluous method calls and unused methods.

With the move to NetStandard 2.0, many of the directives that were needed for NetStandard 1.3 are no longer required, because the API surface in NetStandard 2.0 is much larger.


russcam:
porting to 7.x and master

PR #3667, opened 2019-04-15
Deserialize relation on TotalHits

russcam:
This PR deserializes the relation and renames `HitsTotal` to `TotalHits`, line with Java source.

`TotalHitsFormatter` handles both object and numeric representations of `TotalHits`. When the value is numeric, no relation is present, and there is no simple way to determine what the relation should be. To handle this, `Relation` is of type `TotalHitsRelation?`


src/Nest/Search/Search/Hits/TotalHits.cs
```diff
@@ -0,0 +1,99 @@
...
+						if (propertyName.EqualsBytes(ValueField))
+							value = reader.ReadInt64();
+						else if (propertyName.EqualsBytes(RelationField))
+							relation = RelationFormatter.Deserialize(ref reader, formatterResolver);
+						else
+							reader.ReadNextBlock();
+					}
+
+					return new TotalHits { Value = value, Relation = relation };
+				case JsonToken.Number:
```
codebrain: Mixed format response :(

russcam: This is to handle [`rest_total_hits_as_int=true`](https://www.elastic.co/guide/en/elastic-stack/7.0/elasticsearch-breaking-changes.html#_hits_total_is_now_an_object_in_the_search_response) request parameter

PR #3670, opened 2019-04-15
7x/remove response interface

Mpdreamz:
API Calls now return concrete response implementations.

DoRequestAsync no longer needs to do perform tricks to go from `Task<Response>` to `Task<IResponse>`.

Interfaces are kept for responses where they benefit from covariance. 
This change now also makes `ISearchRequest<T>` completely covariant 🎉


Mpdreamz:
Reviewed with the team, using my admin powers to merge unreviewed.

Mpdreamz:
ported to master

PR #3676, opened 2019-04-16
Docs: Drop subtitle in Asciidoctor

nik9000:
When we use Asciidoctor to build the docs for the elasticsearch-net repo
it sees the `:` in the title of the book as a delimiter for a subtitle.
This is probably a nice feature, but we actually intend the title of the
book to have a `:` in it. The simple work around is to use a different
character for the delimiter. This change does that.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -1,3 +1,5 @@
+:title-separator: |
+
 [[elasticsearch-net-reference]]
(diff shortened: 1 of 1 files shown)


russcam:
I'll pull in now to the relevant branches

nik9000:
Thanks!

PR #3677, opened 2019-04-17
Refactor ServerError, Error and ErrorCause

russcam:
With both the low level client and high level client now using the same internal serializer, the deserialization of `ServerError`, `Error` and `ErrorCause` can be simplified. 

`ErrorCauseMetadata` is removed from `ErrorCause`, and the properties it contains are hoisted onto `ErrorCause`. It is possible that the JSON for and error cause contains properties that are not mapped on `ErrorCause`. Currently, these will be ignored. If we feel it is worthwhile, we could also capture these too, by implementing `ErrorCause` as a dictionary with a formatter to deserialize and set known values


Mpdreamz:
>  we could also capture these too, by implementing ErrorCause as a dictionary with a formatter to deserialize and set known values

Relates to https://github.com/elastic/elasticsearch/issues/27672

I'd be +1 to rework `ErrorCause` into a dictionary as a long term plan. Let's track this in a separate issue though.

src/Tests/Tests/Document/Multiple/DeleteByQuery/DeleteByQueryApiTests.cs
```diff
@@ -209,7 +209,7 @@ protected override void ExpectResponse(IDeleteByQueryResponse response)
 			failure.Id.Should().NotBeNullOrWhiteSpace();
 
 			failure.Cause.Should().NotBeNull();
-			failure.Cause.IndexUniqueId.Should().NotBeNullOrWhiteSpace();
+			failure.Cause.IndexUUID.Should().NotBeNullOrWhiteSpace();
```
Mpdreamz: 👍

PR #3697, opened 2019-04-26
Fix MoreLikeThisQuery conditionless behaviour

### LINKED ISSUE #3696: More Like This Query Conditionless behaviour
Relates: https://stackoverflow.com/questions/55750953/more-like-this-query-not-getting-serialized-nest

More Like This Query is determined to be conditionless when `Fields` is conditionless, which is incorrect. The query should be conditionless only when `Fields` is conditionless and `Like` is conditionless.

russcam:
Fixes #3696


russcam:
ported to 7.x: e58c56a5a32ba7bae4ec9c3917953e4e1f654d73

PR #3702, opened 2019-04-29
Support deprecate Name property on ElasticsearchTypeAttribute

### LINKED ISSUE #3681: 7.0.0 ElasticsearchTypeAttribute should continue to support Name property
7.0.0-alpha1
**Elasticsearch version**:
7.0.0

**Actual**
Even though mapping types have been removed in Elasticsearch 7.x, option to set name of the class for serialization is still useful when using fluent/strongly-type interfaces.

1. When defining `Join` in index mapping:
```
_client.CreateIndexAsync(IndexName,
    index => index.Map<MyDocument>(
        tmd => tmd.AutoMap()
        .Properties(
            props => props.Join(
                j => j.Name(p => p.Relationship)
                .Relations(
                    r => r.Join<ParentType, ChildType>()
                    )
                )
            )
        )
    );
```

2. When linking instances of child types to parent types:
```
var parent = new ParentType { Id = 1 };
var parentDoc = new MyDocument
{
    Parent = parent,
    Relationship = JoinField.Root<ParentType>()
}

var childDoc = new MyDocument
{
    Child = new ChildType(),
    Relationship = JoinField.Link<ChildType>(new Id(parent.Id))
}
```
3. When using `JoinAttribute` on `JoinField` property:
```
public class MyDocument
{
    public ParentType Parent { get; set; }

    public ChildType Child { get; set; }

    [Join(typeof(ParentType), typeof(ChildType))]
    public JoinField Relationship { get; set; }
}
```

**Expected**
Therefore, possibility to override class name for serialization should still be supported. In the past `ElasticsearchTypeAttribute`'s `Name` property was used for this.

russcam:
This commit reintroduces the Name property on ElasticsearchTypeAttribute, which gets/sets the RelationName property. The reintroduction is marked as Obsolete as it will be removed in 8.x

Closes #3681

### DIFF
--- a/src/Nest/Mapping/AttributeBased/ElasticsearchTypeAttribute.cs
+++ b/src/Nest/Mapping/AttributeBased/ElasticsearchTypeAttribute.cs
@@ -5,22 +5,43 @@ using System.Reflection;
 
 namespace Nest
 {
+	/// <summary>
+	/// Applied to a CLR type to override the name of a CLR type and the property from which an _id is inferred
+	/// when serializing an instance of the type.
+	/// </summary>
 	[AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
 	public class ElasticsearchTypeAttribute : Attribute
 	{
 		private static readonly ConcurrentDictionary<Type, ElasticsearchTypeAttribute> CachedTypeLookups =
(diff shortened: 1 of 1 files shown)


russcam:
I'm going to pull this in for the 7.0.0-alpha2

PR #3714, opened 2019-05-03
Add SequenceNumber and PrimaryTerm to Hits, Get and MulitGetHit

### LINKED ISSUE #3708: SequenceNumber and PrimaryTerm missing from IMultiGetHit and IHit
**NEST/Elasticsearch.Net version**: 7.0.0-alpha1

**Elasticsearch version**: 7.0.0

When using the `seq_no_primary_term` option on search requests it returns the `_seq_no` and `_primary_term`, but those fields are not present in the `IHit` models in the C# client.

russcam:
This commit adds the SequenceNumber and PrimaryTerm properties to the Hit<T>, MultiGetHit<T> and GetResponse.

Closes #3708

### DIFF
--- a/src/Nest/Document/Single/Get/GetResponse.cs
+++ b/src/Nest/Document/Single/Get/GetResponse.cs
@@ -1,13 +1,14 @@
-﻿using System;
-using System.Runtime.Serialization;
+﻿using System.Runtime.Serialization;
 using Elasticsearch.Net;
 
 namespace Nest
 {
+	// TODO: Looks like this can be removed?
 	public interface IGetResponse<out TDocument> : IResponse where TDocument : class
 	{
 		TDocument Source { get; }
 	}
+
 	public class GetResponse<TDocument> : ResponseBase, IGetResponse<TDocument> where TDocument : class
 	{
 		[DataMember(Name = "fields")]
@@ -22,8 +23,14 @@ namespace Nest
 		[DataMember(Name = "_index")]
 		public string Index { get; internal set; }
 
+		[DataMember(Name = "_primary_term")]
+		public long? PrimaryTerm { get; internal set; }
+
 		[DataMember(Name = "_routing")]
 		public string Routing { get; internal set; }
+
+		[DataMember(Name = "_seq_no")]
+		public long? SequenceNumber { get; internal set; }
 
 		[DataMember(Name = "_source")]
 		[JsonFormatter(typeof(SourceFormatter<>))]
--- a/src/CodeGeneration/ApiGenerator/Overrides/GlobalOverrides.cs
+++ b/src/CodeGeneration/ApiGenerator/Overrides/GlobalOverrides.cs
@@ -38,6 +38,8 @@ namespace ApiGenerator.Overrides
 			{ "s", "sort_by_columns" },
 			{ "v", "verbose" },
 			{ "ts", "include_timestamp" },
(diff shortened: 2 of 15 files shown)


russcam:
We should consider standardising on the naming for Sequence Number, Primary Term. I've added as `SequenceNumber` and `PrimaryTerm` but notice that generated methods use `SeqNoPrimaryTerm`, etc.

src/Nest/Document/Single/Get/GetResponse.cs
```diff
@@ -1,13 +1,14 @@
-﻿using System;
-using System.Runtime.Serialization;
+﻿using System.Runtime.Serialization;
 using Elasticsearch.Net;
 
 namespace Nest
 {
+	// TODO: Looks like this can be removed?
```
codebrain: TODO?

ejsmith:
These need to be added to `IBulkOperation` as well.

Also, it seems a bit bad that there isn't any shared interface that contains `SequenceNumber` and `PrimaryTerm`. I currently need to check for these different types `GetResponse` `IHit` `IndexResponse` `IMultiGetHit` to get the information I need. Seems like it would be nice to have them all implement an interface.

src/Nest/Document/Single/Get/GetResponse.cs
russcam: Related to the removal of response interfaces that @Mpdreamz was working on.

Added to discuss before GA, but not needed to address to merge this PR.

russcam:
@ejsmith Are there scenarios where you think having an interface for the two properties would help? I'm struggling to think of them, because aren't you always dealing with a known response or type? Do you have an example of how an interface would help, to help me see?

ejsmith:
@russcam I have scenarios where I am doing versioning on documents for all sorts of use cases and that requires me to map those properties to and from my own models. So currently, I need to have extension methods for all of the different response and request models that have the properties instead of just a single extension method if there was a shared contract for these properties.

Also, just want to confirm that you saw that these properties are missing on `IBulkOperation` as I am currently blocked waiting for these.

russcam:
@ejsmith Ok, we can introduce an interface for both properties. Suggestions for a name, `ISequenceNumberPrimaryTerm`?

> Also, just want to confirm that you saw that these properties are missing on IBulkOperation as I am currently blocked waiting for these.

They're on [bulk index](https://github.com/elastic/elasticsearch-net/blob/f374afcc30a2aae597faa91709cc89f8e5cd34a4/src/Nest/Document/Multiple/Bulk/BulkOperation/BulkIndex.cs#L19-L23) and [delete](https://github.com/elastic/elasticsearch-net/blob/f374afcc30a2aae597faa91709cc89f8e5cd34a4/src/Nest/Document/Multiple/Bulk/BulkOperation/BulkDelete.cs#L13-L17) operations [per the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-optimistic-concurrency-control), are we missing a place?

russcam:
Looking into an interface for SequenceNumber and PrimaryTerm, I don't think it'd be possible to unify types that have these properties to one interface because they may not have values on some types, depending on whether `SeqNoPrimaryTerm` is passed in the request. For this reason, they're modelled with `long?` on these types, and modelled as `long` in cases where they will always have values.

In order to get an alpha2 out, I'm going to merge this PR in, and we can open another issue to discuss an interface.

PR #3755, opened 2019-05-20
Clean up ApiGenerator

Mpdreamz:
This PR cleans up the ApiGenerator a tad. One of the oldest parts of the codebase dating back to `1.0`

The razor views can now be split using partials and they use more dedicated models for each individual component. In the past everything was driven by `CsharpMethod` now request/interfaces/et all have specific domain models. 

The generated files are now also automatically indented so we need to fiddle a whole lot less with whitespace in the razor views.

We no longer generate alternative http methods on the low level client so things such as `client.SearchGet()` are a thing of the past.

This PR also serves as jumping point for introducing namespaces and the domain model should help massively when we want to reimplement the yaml test runner next.


src/CodeGeneration/ApiGenerator/ApiEndpointFactory.cs
```diff
@@ -0,0 +1,94 @@
...
+			var (name, endpoint) = officialJsonSpec.ToObject<Dictionary<string, ApiEndpoint>>().First();
+			
+			endpoint.FileName = Path.GetFileName(jsonFile);
+			endpoint.Name = name;
+			var tokens = name.Split(".");
+			
+			endpoint.MethodName = tokens.Last();
+			if (tokens.Length > 1)
+				endpoint.Namespace = tokens[0];
+			//todo side effect
```
codebrain: Stray TODO?

src/CodeGeneration/ApiGenerator/ApiEndpointFactory.cs
```diff
@@ -0,0 +1,94 @@
...
+			var patchedJson = JObject.Parse(File.ReadAllText(patchFile));
+
+			var pathsOverride = patchedJson.SelectToken("*.url.paths");
+
+			original.Merge(patchedJson, new JsonMergeSettings
+			{
+				MergeArrayHandling = MergeArrayHandling.Union
+			});
+
+			if (pathsOverride != null) original.SelectToken("*.url.paths").Replace(pathsOverride);
```
codebrain: :+1: simplifies the `.replace.json` files...

src/CodeGeneration/ApiGenerator/Views/Descriptors/Descriptor.cshtml
```diff
@@ -0,0 +1,85 @@
...
+	{
+<text>
+		///<summary>@param.Description</summary>
+		public @Raw(type) @param.ClsName@(Raw(typed ? "" : "<T>"))(@Raw("Expression<Func<"+ g +", object>>") field) @Raw(typed ? "" : "where " + g + " : class") => Qs("@original", (Field)field);
+</text>
+	}
+}
+	@if (names.DescriptorNotFoundInCodebase)
+	{<text>
+		//TODO THIS METHOD IS UNMAPPED!
```
codebrain: I've always wondered if we ought to introduce a deliberate compiler error here instead, to prompt us to do the right thing and fix it.... thoughts?

src/CodeGeneration/ApiGenerator/Views/Requests/Requests.cshtml
```diff
@@ -0,0 +1,66 @@
...
+	if (param.RenderPartial && (r.HasBody))
+	{
+		continue;
+	}
+	var doc = param.DescriptionHighLevel.ToArray();
+<text>		@Raw(param.InitializerGenerator(param.TypeHighLevel, param.ClsName, original, param.SetterHighLevel, doc))
+</text>
+}
+@if (names.DescriptorNotFoundInCodebase)
+{<text>		//TODO THIS METHOD IS UNMAPPED! Expected to find @names.DescriptorName and @names.RequestName in a file called @(names.RequestName).cs in NEST's codebase
```
codebrain: Introduce a deliberate compiler error here instead?

src/CodeGeneration/ApiGenerator/Views/Descriptors/Descriptor.cshtml
Mpdreamz: ++ :100: I want to generate these into `Unmapped.Requests|Response.<MethodName>.cs` where we generate stubs for `Request` and `Response` with obsoletes with compiler errors on them.

This makes adding a new API easier and makes sure we won't release a stub.

PR #3757, opened 2019-05-21
Include stored_scripts response in cluster metadata response.

### LINKED ISSUE #3746: Stored Scripts are not serialized when requesting cluster state Metadata

6.5.0

**Elasticsearch version**:

6.2.4

**Describe the feature**:

Currently requesting the cluster state metadata (`client.ClusterState(s => s.Metric(ClusterStateMetric.Metadata))`) does not serialize the `stored_scripts` portion of the response. 

If I were to take a swing at implementing this, the sub field on the state response would provide something like a `Dictionary<Id, IStoredScript>`.  The (simplified) response json from the above request:

```json
{
    "metadata": {
        "stored_scripts": {
            "my-script-id": {
                "lang":"painless",
                "source":"return 1;",
                "options":{}
            },
            "my-other-script-id": {
                "lang":"painless",
                "source":"return 0;",
                "options":{}
            },
        }
    }
}

```

I also notice the response includes an `options` that isn't currently being serialized. I'm not sure what this is useful for since we don't utilize it, but it is missing (maybe intentionally?)

codebrain:
Fixes #3746.

Backport to `7.x` and `master` after approval and merge.

### DIFF
--- a/src/Nest/Cluster/ClusterState/StoredScriptMapping.cs
+++ b/src/Nest/Cluster/ClusterState/StoredScriptMapping.cs
@@ -6,7 +6,7 @@ namespace Nest
 	public class StoredScriptMapping
 	{
 		[JsonProperty("lang")]
-		public string Lang { get; internal set; }
+		public string Language { get; internal set; }
 
 		[JsonProperty("source")]
 		public string Source { get; internal set; }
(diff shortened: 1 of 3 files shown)


src/Nest/Cluster/ClusterState/StoredScriptMapping.cs
```diff
@@ -0,0 +1,17 @@
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Nest
+{
+	public class StoredScriptMapping
+	{
+		[JsonProperty("lang")]
+		public string Lang { get; internal set; }
```
russcam:
```suggestion
		public string Language { get; internal set; }
```

codebrain:
Ported to `7.x`

PR #3762, opened 2019-05-21
7x/lowlevel namespace

Mpdreamz:
This introduces namespace properties on `IElasticLowLevelClient`

Now you can do `client.Indices.PutIndex()` rather than `client.IndicesPutIndex()` the difference is subtle. This combined with the changes in the target PR that clean up the methods names really will make the lowlevel client much more intuitive to use.

This PR also further separates the domain modeling of the Request/Response naming with the name of the Method, these are no longer tightly coupled.


src/CodeGeneration/ApiGenerator/Domain/Code/CsharpNames.cs
```diff
@@ -1,26 +1,56 @@
...
-			Namespace = endpointNamespace.ToPascalCase();
+			Namespace = CreateCSharpNamespace(endpointNamespace);
 			if (CodeConfiguration.ApiNameMapping.TryGetValue(name, out var mapsApiMethodName))
-				MethodName = mapsApiMethodName;
-			else MethodName = $"{Namespace}{endpointMethodName.ToPascalCase()}";
+				ApiName = mapsApiMethodName;
+			else ApiName = endpointMethodName.ToPascalCase();
+
+			//if the api name starts with the namespace do not repeat it in the method name
+			MethodName = Regex.Replace(ApiName, $"^{Namespace}(\\w+)$", "$1");
```
codebrain: ++ also need to use a dictionary for pluralisation (Indices / Index) or abbreviations (ILM / Index Lifecycle Management)

src/CodeGeneration/ApiGenerator/Domain/Code/CsharpNames.cs
```diff
@@ -1,26 +1,56 @@
...
+		
+		public string RequestName => $"{ApiName}Request";
+		public string InterfaceName => $"I{RequestName}";
+		public string ParametersName => $"{RequestName}Parameters";
+		public string DescriptorName => $"{ApiName}Descriptor";
+
+		public const string RootNamespace = "NoNamespace";
+		private static string CreateCSharpNamespace(string endpointNamespace)
+		{
+			switch (endpointNamespace)
```
codebrain: ++

src/CodeGeneration/ApiGenerator/Domain/Specification/ApiEndpoint.cs
```diff
@@ -104,45 +104,27 @@ public IReadOnlyCollection<CsharpMethod> CsharpMethods
...
-						ReturnType = "Task<TResponse>",
-						PerPathMethodName = methodName + "Async",
+						Arguments = string.Join(", ", args),
+						CsharpNames = CsharpNames,
+						PerPathMethodName = methodName,
 						HttpMethod = httpMethod,
 						OfficialDocumentationLink = OfficialDocumentationLink,
-						ObsoleteMethodVersion = null, //TODO
-						Arguments = string.Join(", ", args),
+						DeprecatedPath = null, //TODO
```
codebrain: Wondering if we just delete it in the client for now...

src/CodeGeneration/ApiGenerator/Domain/Code/CsharpNames.cs
codebrain: (just seen abbreviations below!)

PR #3763, opened 2019-05-23
Attribute SuggestBucket for serialization

russcam:
This commit adds JsonProperty attributes to SuggestBucket to ensure
that properties are correctly serialized when a DefaultFieldNameInferrer delegate is passed.

These are required because the concrete SuggestBucket is not passed and serialized as the
ISuggestBucket interface in the VerbatimDictionaryKeysJsonConverter WriteJson method.

Fixes issue found by https://discuss.elastic.co/t/nest-suggester-errors/179634/5


codebrain:
Ported to `7.x`

PR #3764, opened 2019-05-23
 XPack info and usage improvements.

codebrain:
Add integration tests for XPack info and usage endpoints.
Additional properties on request objects.


codebrain:
Ported to `7.x`

Issue #3765, opened 2019-05-23
7.0.0-alpha2 - Clean up GET /_cluster/stats


codebrain:
Response model has changed and some properties can be cleaned up.

codebrain:
Since there are no guarantees that cluster metadata adheres to semver versioning it might make sense to remove the response model and introduce helper methods to retrieve values.

codebrain:
Linked to this PR: https://github.com/elastic/elasticsearch-net/pull/3780

codebrain:
Addressed in https://github.com/elastic/elasticsearch-net/pull/3780

PR #3766, opened 2019-05-23
Add pretty names (and other properties) to the cluster stats response.

codebrain:
Do not port to `master` or `7.x` as responses need to be cleaned up in those branches as per: https://github.com/elastic/elasticsearch-net/issues/3765

### DIFF
--- a/src/Nest/CommonOptions/Stats/PluginStats.cs
+++ b/src/Nest/CommonOptions/Stats/PluginStats.cs
@@ -11,6 +11,9 @@ namespace Nest
 		[JsonProperty("description")]
 		public string Description { get; set; }
 
+		[JsonProperty("elasticsearch_version")]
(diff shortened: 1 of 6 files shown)


src/Nest/CommonOptions/Stats/PluginStats.cs
```diff
@@ -23,6 +26,12 @@ public class PluginStats
 		[JsonProperty("site")]
 		public bool Site { get; set; }
 
+		[JsonProperty("java_version")]
+		public string JavaVersion { get; set; }
+
+		[JsonProperty("has_native_controller")]
+		public bool HasNativeController { get; set; }
```
Mpdreamz: Is this returned from `6.0` onwards? if not should this be `bool?`?

PR #3767, opened 2019-05-24
Add doc’s sequence number + primary term to get responses

codebrain:
Addresses: https://github.com/elastic/elasticsearch/pull/36680

Backport to `master` and `7.x` requires care due to naming conventions,

### DIFF
--- a/src/Nest/Document/Single/Get/GetResponse.cs
+++ b/src/Nest/Document/Single/Get/GetResponse.cs
@@ -33,6 +33,12 @@ namespace Nest
 
 		[JsonProperty("_version")]
 		long Version { get; }
+
+		[JsonProperty("_primary_term")]
+		long PrimaryTerm { get; }
+
+		[JsonProperty("_seq_no")]
+		long SequenceNumber { get; }
 	}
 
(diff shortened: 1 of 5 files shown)


codebrain:
Also need to check other Single document method responses (Create / Index...)

src/Nest/Document/Single/Get/GetResponse.cs
```diff
@@ -33,6 +33,12 @@ public interface IGetResponse<out TDocument> : IResponse where TDocument : class
 
 		[JsonProperty("_version")]
 		long Version { get; }
+
+		[JsonProperty("_primary_term")]
```
Mpdreamz: should this be `long?` on `6.x` ? Not returned on older versions.

codebrain: I've looked into this, there are other areas of the code that have implemented both properties as `long` and not `long?`...

russcam: They are implemented as `long` in places where a value is always returned, and as nullable in places where a value is returned only if a request parameter is provided that indicates to return it

PR #3773, opened 2019-05-27
7x/highlevel namespae methods

Mpdreamz:
Simplifies namespaced methods.


Mpdreamz:
Cherry picked the last two commits to `7.x`

PR #3776, opened 2019-05-28
Use a custom convertor for GeoOrientation to tolerate alternative options in Elasticsearch

codebrain:
The documentation (https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-shape.html) states that there are multiple values possible; 1. Right-hand rule: `right`, `ccw`, `counterclockwise`, 2. Left-hand rule: `left`, `cw`, `clockwise`.

It looks like NEST only supported `cw` and `ccw`.

Cluster metadata exposed `cw` as `LEFT` in the responses.

### DIFF
--- a/src/Nest/Mapping/Types/Geo/GeoShape/GeoOrientation.cs
+++ b/src/Nest/Mapping/Types/Geo/GeoShape/GeoOrientation.cs
@@ -1,16 +1,45 @@
-﻿using System.Runtime.Serialization;
+﻿using System;
 using Newtonsoft.Json;
-using Newtonsoft.Json.Converters;
 
 namespace Nest
 {
-	[JsonConverter(typeof(StringEnumConverter))]
+	[JsonConverter(typeof(GeoOrientationConverter))]
 	public enum GeoOrientation
 	{
-		[EnumMember(Value = "cw")]
 		ClockWise,
-
-		[EnumMember(Value = "ccw")]
 		CounterClockWise
+	}
+
+	public class GeoOrientationConverter : JsonConverter
+	{
+		public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+		{
+			var geoOrientation = (GeoOrientation)value;
+			switch (geoOrientation)
+			{
+				case GeoOrientation.ClockWise:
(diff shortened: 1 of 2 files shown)


src/Nest/Mapping/Types/Geo/GeoShape/GeoOrientation.cs
```diff
@@ -1,16 +1,45 @@
...
 	public enum GeoOrientation
 	{
-		[EnumMember(Value = "cw")]
 		ClockWise,
-
-		[EnumMember(Value = "ccw")]
 		CounterClockWise
 	}
+
+	public class GeoOrientationConverter : JsonConverter
```
Mpdreamz:
```
[xUnit.net 00:00:12.5484924]     Tests.CodeStandards.Serialization.Converters.CustomConvertersShouldBeInternal [FAIL]
Failed   Tests.CodeStandards.Serialization.Converters.CustomConvertersShouldBeInternal
Error Message:
 Expected collection to be empty, but found {"GeoOrientationConverter"}.
Stack Trace:
````

yay code standards tests :medal_sports: :smile_cat:

**Comment by codebrain** [comment]
[geo_shape_error.txt](https://github.com/elastic/elasticsearch-net/files/3230640/geo_shape_error.txt)

Including example cluster metadata state with the transformed `LEFT` value.

File: src/Nest/Mapping/Types/Geo/GeoShape/GeoOrientation.cs [hunk 2]
```diff
@@ -1,16 +1,45 @@
...
+				case GeoOrientation.CounterClockWise:
+					writer.WriteValue("ccw");
+					break;
+			}
+		}
+
+		public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+		{
+			var enumString = (string)reader.Value;
+			switch (enumString.ToLower())
```
russcam: think this should use `ToLowerInvariant()`, rather than use the current culture

codebrain: Didn't make the cut for the merge, but will fix for `7.x`

PR #3779, opened 2019-05-30
Use a custom convertor for GeoOrientation to tolerate alternative options in Elasticsearch #3776

codebrain:
Opening to start a conversation on how to implement a `JsonFormatter` for an `enum`.

OOTB it looks like UTF8JSON doesnt support placing the attribute on enum values.

Any idea where to go with this? @russcam

### DIFF
--- a/src/Nest/Mapping/Types/Geo/GeoShape/GeoOrientation.cs
+++ b/src/Nest/Mapping/Types/Geo/GeoShape/GeoOrientation.cs
@@ -1,5 +1,4 @@
-﻿using System.Runtime.Serialization;
-using Elasticsearch.Net;
+﻿using Elasticsearch.Net;
 
 namespace Nest
 {
@@ -27,11 +26,11 @@ namespace Nest
 		public GeoOrientation Deserialize(ref JsonReader reader, IJsonFormatterResolver formatterResolver)
 		{
 			var enumString = reader.ReadString();
-			switch (enumString.ToLowerInvariant())
+			switch (enumString)
 			{
-				case "left":
-				case "cw":
-				case "clockwise":
+				case "LEFT":
+				case "CW":
+				case "CLOCKWISE":
 					return GeoOrientation.ClockWise;
 			}
 			// Default, complies with the OGC standard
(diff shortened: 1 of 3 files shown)


src/Nest/Mapping/Types/Geo/GeoShape/GeoOrientation.cs
```diff
@@ -1,16 +1,41 @@
...
+				case GeoOrientation.CounterClockWise:
+					writer.WriteString("ccw");
+					break;
+			}
+		}
+
+		public GeoOrientation Deserialize(ref JsonReader reader, IJsonFormatterResolver formatterResolver)
+		{
+			var enumString = reader.ReadString();
+			switch (enumString.ToLowerInvariant())
```
russcam: Is `ToLowerInvariant()` necessary?

codebrain: Yes, ES capitalises.

russcam: would it be better to check against captialized names, and avoid the `ToLowerInvariant()` call?

codebrain: ...perhaps, but its not a very defensive approach.

russcam: But it avoids the string allocation in a scenario where we know what to expect from the server, avoiding unneccessary work

src/Tests/Tests/Mapping/Types/Geo/GeoShape/GeoShapeClusterMetadataApiTests.cs
```diff
@@ -0,0 +1,87 @@
...
+			(client, f) => client.Map(f),
+			(client, f) => client.MapAsync(f),
+			(client, r) => client.Map(r),
+			(client, r) => client.MapAsync(r)
+		);
+
+		protected override void ExpectResponse(PutMappingResponse response)
+		{
+			// Ensure metadata can be deserialised
+			var metadata = Client.Cluster.State(null, r => r.Metric(ClusterStateMetric.Metadata));
```
russcam: maybe limit to a given index from `IntegrationSetup`?

PR #3780, opened 2019-05-30
Add pretty names (and other properties) to the cluster stats response. #3766

codebrain:
Removes some of the redundant properties.

### DIFF
--- a/src/Nest/CommonOptions/Stats/SegmentsStats.cs
+++ b/src/Nest/CommonOptions/Stats/SegmentsStats.cs
@@ -8,68 +8,38 @@ namespace Nest
 		[DataMember(Name ="count")]
 		public long Count { get; set; }
 
-		[DataMember(Name ="doc_values_memory")]
-		public string DocValuesMemory { get; set; }
-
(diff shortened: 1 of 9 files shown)


src/Nest/CommonOptions/Stats/SegmentsStats.cs
```diff
@@ -8,70 +9,43 @@ public class SegmentsStats
...
 		public long TermVectorsMemoryInBytes { get; set; }
 
-		[DataMember(Name ="version_map_memory")]
-		public string VersionMapMemory { get; set; }
-
 		[DataMember(Name ="version_map_memory_in_bytes")]
 		public long VersionMapMemoryInBytes { get; set; }
+		
+		[DataMember(Name ="file_sizes")]
+		public IReadOnlyDictionary<string, ShardFileSizeInfo> FileSizes { get; internal set; } = EmptyReadOnly<string, ShardFileSizeInfo>.Dictionary;
```
russcam: does this need to be initialized with empty? Typically, we've only initialized the top level dictionary properties

codebrain: ++

PR #3784, opened 2019-05-31
Add User-Agent string to API calls to Elasticsearch

russcam:
This commit adds a User-Agent to all API calls to Elasticsearch, that includes the client name and version, along with OS, Framework and assembly.

Where System.Runtime.InteropServices.RuntimeInformation is available, use the properties of this type. For .NET 4.6.1 which does not support this type, use native Windows functions to determine the version of Windows, and Environment.OSVersion.VersionString for other platforms. The latter is not used for Windows because it can report incorrect version information depending on runtime version and service packs installed, and the context in which calls are executed i.e. it is not reliable.

### DIFF
--- a/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
+++ b/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
@@ -41,7 +41,7 @@ namespace Elasticsearch.Net
 
 		/// <summary>
 		/// The default connection limit for both Elasticsearch.Net and Nest. Defaults to <c>80</c> except for
-		/// <see cref="HttpClientHandler"/> implementations based on curl, which defaults to
+		/// HttpClientHandler implementations based on curl, which defaults to
 		/// <see cref="Environment.ProcessorCount"/>
 		/// </summary>
 		public static readonly int DefaultConnectionLimit = IsCurlHandler ? Environment.ProcessorCount : 80;
--- a/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
+++ b/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
@@ -18,52 +18,85 @@ using System.Runtime.InteropServices;
 namespace Elasticsearch.Net
 {
 	/// <summary>
-	/// ConnectionConfiguration allows you to control how ElasticLowLevelClient behaves and where/how it connects
-	/// to elasticsearch
+	/// Allows you to control how <see cref="ElasticLowLevelClient"/> behaves and where/how it connects to Elasticsearch
 	/// </summary>
 	public class ConnectionConfiguration : ConnectionConfiguration<ConnectionConfiguration>
 	{
+		/// <summary>
+		/// The default ping timeout. Defaults to 2 seconds
+		/// </summary>
 		public static readonly TimeSpan DefaultPingTimeout = TimeSpan.FromSeconds(2);
-		public static readonly TimeSpan DefaultPingTimeoutOnSSL = TimeSpan.FromSeconds(5);
-		public static readonly TimeSpan DefaultTimeout = TimeSpan.FromMinutes(1);
-		public static readonly int DefaultConnectionLimit = IsCurlHandler ? Environment.ProcessorCount : 80;
-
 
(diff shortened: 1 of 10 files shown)


russcam:
Opening to discuss

src/Elasticsearch.Net/Configuration/IConnectionConfigurationValues.cs
```diff
@@ -213,5 +214,11 @@ public interface IConnectionConfigurationValues : IDisposable
 		bool ThrowExceptions { get; }
 
 		ElasticsearchUrlFormatter UrlFormatter { get; }
+
+		/// <summary>
+		/// The user agent string to send with requests. Useful for debugging purposes to understand client and framework
+		/// versions that initiate requests to Elasticsearch
+		/// </summary>
+		string UserAgent { get; }
```
codebrain: Might be worth including a few examples?

russcam: What would those examples look like? Does providing examples encourage its usage? If it would, I don't think we should provide any.

There's the [RFC 2616 for Hypertext Transfer Protocol -- HTTP/1.1, that includes User-Agent](https://tools.ietf.org/html/rfc2616#page-145), which Elasticsearch may or may not adhere to (would need to dig into). 

The value I see in providing a way to set your own value is only really for the case where the default value that the client would now send messing with the environment in some way e.g. a proxy in between the client and Elasticsearch that doesn't like the default value.

codebrain: I was just thinking from the perspective of the user who can read the code, see an example of a user agent string and say to themselves "Oh, I get it".

Its a `{ get; }` anyway.

src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
```diff
@@ -502,6 +507,12 @@ public T EnableDebugMode(Action<IApiCallDetails> onRequestCompleted = null)
 		public T SkipDeserializationForStatusCodes(params int[] statusCodes) =>
 			Assign(new ReadOnlyCollection<int>(statusCodes), (a, v) => a._skipDeserializationForStatusCodes = v);
 
+		/// <summary>
+		/// The user agent string to send with requests. Useful for debugging purposes to understand client and framework
+		/// versions that initiate requests to Elasticsearch
+		/// </summary>
+		public T UserAgent(string userAgent) => Assign(userAgent, (a, v) => a._userAgent = v);
```
Mpdreamz: Should we allow the full userAgent to be configured or just the moniker?

Bonus would be that `ConnectionSettings` and `ConnectionConfiguration` could ship different default `monikers` but rely on the same full agent string building.

src/Elasticsearch.Net/Configuration/IConnectionConfigurationValues.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Net.Security;
+using System.Reflection;
```
Mpdreamz: unused?

src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
Mpdreamz: The client identification is part of the metadata. Perhaps this public method should only set the application name (where we put repos name).

Mpdreamz:
LGTM, totally happy to ship with metadata that is best effort on some platforms as long as that metadata is static and not costly to compute.

src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
```diff
@@ -139,6 +140,9 @@ public abstract class ConnectionConfiguration<T> : IConnectionConfigurationValue
 
 		private readonly ElasticsearchUrlFormatter _urlFormatter;
 
+		private string _userAgent =
```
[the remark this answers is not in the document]
russcam: introduced static readonly on `ConnectionConfiguration`

PR #3785, opened 2019-05-31
Ensure CatHelp API calls are always deserialised using a specific routine to avoid the cat ( =^.^=) ASCII art

codebrain:


src/Nest/ElasticClient.cs
```diff
@@ -46,6 +47,28 @@ private CatResponse<TCatRecord> DeserializeCatResponse<TCatRecord>(IApiCallDetai
 			return catResponse;
 		}
 
+		private CatResponse<TCatRecord> DeserializeCatHelpResponse<TCatRecord>(IApiCallDetails response, Stream stream)
+			where TCatRecord : ICatRecord
```
russcam: does this need to be generic over `TCatRecord`, or can it simply be `CatHelpRecord`?

codebrain: That would involve removing that generic argument from the entire method invocation list, which is a much bigger job...

codebrain: Will replace with `CatHelpRecord`

PR #3795, opened 2019-06-04
make SignificantTermsBucket generic on key type

kevinphelps:
The problem is similar to #2336 except with `SignificantTermsBucket` instead of `KeyedBucket`. My use case is significant terms aggregation over numeric terms. But `SignificantTermsBucket.Key` is always null due to the `Key = (string)key` line in `AggregateFormatter`.

This solution is modeled after the solution for `KeyedBucket` in 91f1f8146d095bb7e72c3036a0d1f6250bb3797e.

I have based this PR on `master` but if this could also be targeted for 6.x, I would be very appreciative. I can open a separate PR targeting the 6.x branch if necessary.


src/Nest/Aggregations/AggregateDictionary.cs
```diff
@@ -121,30 +121,30 @@ public FiltersAggregate Filters(string key)
...
 					BgCount = bucket.BgCount,
 					DocCount = bucket.DocCount,
-					Buckets = bucket.Items.OfType<SignificantTermsBucket>().ToList(),
+					Buckets = GetSignificantTermsBuckets<TKey>(bucket.Items).ToList(),
 					Meta = bucket.Meta
 				};
 		}
 
-		public SignificantTermsAggregate SignificantText(string key)
+		public SignificantTermsAggregate<TKey> SignificantText<TKey>(string key)
```
Mpdreamz: I think having a non generic overload defaulting to `TKey` as string would be a nicer experience for the most common case.

src/Nest/Aggregations/AggregateFormatter.cs
```diff
@@ -889,9 +889,9 @@ private IBucket GetCompositeBucket(ref JsonReader reader, IJsonFormatterResolver
 			else
 				reader.ReadNext(); // }
 
-			var significantTermItem = new SignificantTermsBucket(subAggregates)
+			var significantTermItem = new SignificantTermsBucket<object>(subAggregates)
 			{
-				Key = (string)key,
```
Mpdreamz: key is known to be only `string` or `double` from where its called:

```csharp
		private IBucket GetKeyedBucket(ref JsonReader reader, IJsonFormatterResolver formatterResolver)
```


 perhaps we can use this to our advantage and instantiate either a `SignificantTermBucket<string>` or `SignificantTermBucket<double>` and then use this type information in 

```csharp
private IEnumerable<SignificantTermsBucket<TKey>> GetSignificantTermsBuckets<TKey>(IEnumerable<IBucket> items)	
```

To return the `items` (casted) directly instead.

src/Nest/Aggregations/AggregateDictionary.cs
kevinphelps: I can add that.

kevinphelps:
> Awesome awesome PR thanks for raising it @kevinphelps

Thanks! I'll try to make the requested changes tomorrow! :)

> Backporting to 6.x might be a tad harder as SignificantTermsAggregate SignificantTermsBucket are now generic, we could look into keeping both the non generic and generic versions of those in that branch though

That sounds like it could work, but if not it's alright. I've already forked 6.x internally, so it's no big deal for me at least.

src/Nest/Aggregations/AggregateFormatter.cs
kevinphelps: You could do the same thing in `GetKeyedBucket`. Should that one be refactored as well?

src/Nest/Aggregations/AggregateDictionary.cs
kevinphelps: This is done.

PR #3797, opened 2019-06-04
[WIP] Remove ToArray calls on MemoryStream

Mpdreamz:
and try to reuse .Buffer, new serializer can handle any padding at the end.

### DIFF
--- a/src/Elasticsearch.Net/Extensions/UtilExtensions.cs
+++ b/src/Elasticsearch.Net/Extensions/UtilExtensions.cs
@@ -27,6 +27,16 @@ namespace Elasticsearch.Net.Extensions
 
 			return Encoding.UTF8.GetString(buffer.Array, buffer.Offset, buffer.Count);
 		}
+		/// <summary> Attempts to return to underlying bytes buffer (no-copy) if possible </summary>
+		internal static byte[] ToArrayOrBuffer(this MemoryStream ms)
+		{
+			if (ms is null) return null;
+
+			if (!ms.TryGetBuffer(out var buffer) || buffer.Array is null)
+				return ms.ToArray();
+
+			return buffer.Array;
+		}
 
 		internal static byte[] Utf8Bytes(this string s) => s.IsNullOrEmpty() ? null : Encoding.UTF8.GetBytes(s);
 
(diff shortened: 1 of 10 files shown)


Mpdreamz:
Working on a continuation of this PR where I update our `RecyclableMemoryStream` with recent changes from https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream which actually implements `TryGetBuffer()` rather then falling back to its base `MemoryStream` implementation. 

Something in the implementation of `TryGetBuffer()` seems off, either it needs to account for `offset` or the length is not correct if we are reusing blocks. 

Some more investigation is needed

src/Elasticsearch.Net/Extensions/UtilExtensions.cs
```diff
@@ -27,6 +27,16 @@ internal static string Utf8String(this MemoryStream ms)
...
 		}
+		/// <summary> Attempts to return to underlying bytes buffer (no-copy) if possible </summary>
+		internal static byte[] ToArrayOrBuffer(this MemoryStream ms)
+		{
+			if (ms is null) return null;
+
+			if (!ms.TryGetBuffer(out var buffer) || buffer.Array is null)
+				return ms.ToArray();
+
+			return buffer.Array;
```
russcam: Is it safe to call `TryGetBuffer` on a `RecyclableMemoryStream`? It looks like this method is not overloaded.

Mpdreamz: It's safe in the sense that it works but see my comment on the PR, updating to a later version that actually implements it yields some problems under heavy usage. 

I have a feeling it sometimes writes at an offset which trygetbuffer does not take into account, not been able to trap it yesterday though

russcam:
Is this still waiting on some changes, @Mpdreamz?

Zyklop:
@Mpdreamz In a performance analysis, I found this to be a problem in our use-case. Do you need some help, or testing assistance? I think it could make a noticeable difference in performance.

Mpdreamz:
Hi @Zyklop we are meeting as a team next week in person and spending some time doing some performance deep dives.

I've opened https://github.com/elastic/elasticsearch-net/pull/4191 and https://github.com/elastic/elasticsearch-net/pull/4172 as continuations of this branch. 

It does not tackle everything we want to investigate but its a first step. 

If you like to share your analysis results feel free to email me them to `ml` at `elastic.co`, more data/visuals to look at the better. 

Any testing you can do is helpful and appreciated as always, keep an eye out next week.

Mpdreamz:
Closing this as its out of date.

PR #3805, opened 2019-06-10
Routing must be specified when doc used

russcam:
This commit updates the unit test for more like this query to ensure that routing
is set when an artificial document is specified.


russcam:
On further investigation, I don't think that routing must be specified when doc is used; it's only that the `Project` mapping is configured to always require routing that it must be there for this specific test. Will update this PR.

PR #3806, opened 2019-06-10
User SecureString for Basic and Proxy passwords

### LINKED ISSUE #2505: Consider adding SecureString support for basic auth credentials
Should/do we want to add support for passing basic auth credentials using [SecureString]()?

It seems like it's feasible- quickly taking a look, we would need to use [NetworkCredential](https://msdn.microsoft.com/en-us/library/dd783746(v=vs.110).aspx), which both `HttpClient` and `HttpWebRequest` support, rather than manually adding to the request headers as we do now.

The tricky bit would be supporting both `string` and `SecureString`, as I'd imagine we'd want to do, and also backporting without breaking backwards compatibility.

Thoughts @Mpdreamz @russcam ?

russcam:
This commit changes the BasicAuthenticationCredentials and ProxyPassword types
to use SecureString for holding the passwords in memory, as opposed to as a plain text string.

A plain text string is retrieved from SecureString only at the point at which it is required.
This prevents plain text passwords from being automatically emitted in cases where properties
may be enumerated and captured e.g. logging frameworks.

Closes #2505

### DIFF
--- a/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
+++ b/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Net.Http;
 using System.Net.Security;
+using System.Security;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using Elasticsearch.Net.CrossPlatform;
@@ -21,7 +22,7 @@ namespace Elasticsearch.Net
 	public class ConnectionConfiguration : ConnectionConfiguration<ConnectionConfiguration>
 	{
 		private static bool IsCurlHandler { get; } = typeof(HttpClientHandler).Assembly().GetType("System.Net.Http.CurlHandler") != null;
-		
+
 		public static readonly TimeSpan DefaultPingTimeout = TimeSpan.FromSeconds(2);
 		public static readonly TimeSpan DefaultPingTimeoutOnSSL = TimeSpan.FromSeconds(5);
 		public static readonly TimeSpan DefaultTimeout = TimeSpan.FromMinutes(1);
@@ -72,7 +73,7 @@ namespace Elasticsearch.Net
 		private readonly NameValueCollection _queryString = new NameValueCollection();
 		private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1, 1);
 		private readonly ElasticsearchUrlFormatter _urlFormatter;
-		
+
 		private BasicAuthenticationCredentials _basicAuthCredentials;
 		private X509CertificateCollection _clientCertificates;
 		private Action<IApiCallDetails> _completedRequestHandler = DefaultCompletedRequestHandler;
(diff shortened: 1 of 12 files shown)


Mpdreamz:
Return of SecureString :cake: we left it out for a reason for a long time no less because the `Secure` bit is debatable:

https://github.com/dotnet/platform-compat/blob/master/docs/DE0001.md was always a reason for us not to implement it. 

>A plain text string is retrieved from SecureString only at the point at which it is required.
This prevents plain text passwords from being automatically emitted in cases where properties
may be enumerated and captured e.g. logging frameworks.

Is reason enough to add it in though :+1:

RolfDeVries:
In some cases the createstring method returns a value that looks like chinese instead of the original string. I still have to investigate why. Is this a known issue?

russcam:
@RolfDeVries could you provide more detail about 

1. how you're instantiating `ConnectionSettings` and `ElasticClient`
2. what platform you're running on and framework version
3. the environment it's running e.g. full trust

There should be just a single instance of `SecureString` created, with decryption handled by 

https://github.com/elastic/elasticsearch-net/blob/0584f8df4b1ab75814217f43c796d2c3ba152e4c/src/Elasticsearch.Net/Connection/SecureStrings.cs#L15-L33

which allocates an unmanaged binary string each time and copies the contents of `SecureString`.

russcam:
A quick test with

```c#
var secureString = "password".CreateSecureString();
var count = 10000;
var tasks = new List<Task<string>>(count);

for (int i = 0; i < count; i++)
	tasks.Add(new Task<string>(() =>
	{
		var s = secureString.CreateString();
		Console.WriteLine(s);
		return s;
	}));

tasks.ForEach(t => t.Start());	
Task.WaitAll(tasks.ToArray());
```

returns the expected results, but I may be missing something.

RolfDeVries:
When i use 4 parallel thread to read the password 1000 times it will fail:

using System;
using System.Security;
using System.Threading;
using Elasticsearch.Net;

namespace SecureStringTest
{
    public class Program
    {
        private const string Password = "password";
        private static readonly SecureString SecureString = Password.CreateSecureString();

        public static void Main(string[] args)
        {
            Console.WriteLine(SecureString.CreateString());

            for (var i = 0; i < 4; i++)
            {
                new Thread(ThreadStart).Start();
            }

            Console.ReadLine();
        }

        private static void ThreadStart()
        {
            for (var i = 0; i < 100; i++)
            {
                var actual = SecureString.CreateString();
                Console.WriteLine(actual);
            }
        }
    }
}

Running on .net core 2.2 windows server 2016 with 6 virtual processors.

russcam:
Thanks @RolfDeVries, can replicate on .NET Core 2.2.

russcam:
Looks to be related to `Marshal.SecureStringToBSTR()` et. al. `Marshal.SecureStringToGlobalAllocUnicode()` works

```c#
public static string CreateString(this System.Security.SecureString secureString)
{
    var num = IntPtr.Zero;
    if (secureString != null && secureString.Length != 0)
    {
        try
        {
            num = Marshal.SecureStringToGlobalAllocUnicode(secureString);
            return Marshal.PtrToStringUni(num);
        }
        finally
        {
            if (num != IntPtr.Zero)
                Marshal.ZeroFreeGlobalAllocUnicode(num);
        }
    }

    return string.Empty;
}
```

RolfDeVries:
Looks like its working

russcam:
@RolfDeVries I've opened https://github.com/elastic/elasticsearch-net/pull/3825 to address

RolfDeVries:
Thanks very much for the quick response!

PR #3808, opened 2019-06-10
Remove default TypeNames and TypeName from IClrTypeMapping

russcam:
This commit removes the defaultTypeNames mappings on ConnectionSettings as they
are no longer used.

The TypeName on IClrTypeMapping is also removed as it too is no longer used.

### DIFF
--- a/src/Nest/CommonAbstractions/ConnectionSettings/ClrTypeDefaults.cs
+++ b/src/Nest/CommonAbstractions/ConnectionSettings/ClrTypeDefaults.cs
@@ -26,11 +26,6 @@ namespace Nest
 		/// </summary>
 		string RelationName { get; set; }
 
-		/// <summary>
-		/// The default Elasticsearch type name for the given <see cref="ClrType" />
-		/// </summary>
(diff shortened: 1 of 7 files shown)


russcam:
@Mpdreamz I remember you were talking about reintroducing deprecated paths last week which would mean reintroducing types. Is that still the plan?

Mpdreamz:
It is but only for the low level client so that we can run the YAML tests on that. 

We can continue to scrub the remaining type bits from NEST.

Mpdreamz:
LGTM, one small unit test failure though: https://dev.azure.com/azure3/elasticsearch-net/_build/results?buildId=752&view=ms.vss-test-web.build-test-results-tab&runId=6900&resultId=100110&paneView=debug

russcam:
> LGTM, one small unit test failure though: https://dev.azure.com/azure3/elasticsearch-net/_build/results?buildId=752&view=ms.vss-test-web.build-test-results-tab&runId=6900&resultId=100110&paneView=debug

This is fixed in #3805

PR #3809, opened 2019-06-11
Add overloads of SerializeToString and SerializeToBytes

### LINKED ISSUE #3771: 7.0.0-alpha2 breaking formatter backwards compatibility
```cs
ElasticLowLevelClient.Serializer.SerializeToString()
```
introduced a new `IMemoryStreamFactory` parameter in the middle of the parameter chain - this means that libraries compiled against ElasticSearch.Net 6.X don't work with references to 7.0; this includes libraries installed from NuGet. The AssemblyRedirects generated will try to call `SerializeToString()` with parameters (data, format) instead of (data, null, format) throwing an error.

It'd be nice getting a `SerializeToString()` implementation without the `IMemoryStreamFactory` parameter for backwards compatibility so that existing libraries don't break.

russcam:
This commit adds method overloads of SerializeToString and SerializeToBytes
that accept an IMemoryStreamFactory as a non-optional parameter, and reinstates
methods that do not take an IMemoryStreamFactory. This will make upgrading from
6.x to 7.x easier for those that make use of SerializeToString.

Closes #3771

### DIFF
--- a/src/Elasticsearch.Net/Serialization/ElasticsearchSerializerExtensions.cs
+++ b/src/Elasticsearch.Net/Serialization/ElasticsearchSerializerExtensions.cs
@@ -8,7 +8,13 @@ namespace Elasticsearch.Net
 		public static byte[] SerializeToBytes<T>(
 			this IElasticsearchSerializer serializer,
 			T data,
-			IMemoryStreamFactory memoryStreamFactory = null,
+			SerializationFormatting formatting = SerializationFormatting.Indented) =>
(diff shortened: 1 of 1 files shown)


src/Elasticsearch.Net/Serialization/ElasticsearchSerializerExtensions.cs
```diff
@@ -23,7 +29,13 @@ public static class ElasticsearchSerializerExtensions
 		public static string SerializeToString<T>(
```
Mpdreamz: Obsolete?

PR #3810, opened 2019-06-11
Clear up Name and Id parameters

codebrain:
We could consider consolidating further by using `Id` and `Ids`, but my concern here is that we would be allowing `string` types for ids that only accept `long` values, hence the specialisations.


Mpdreamz:
@codebrain I've added some commits that allow us to expose params as `string` `long`, `long?` directly to the user for the cases that don't benefit from `IUrlParameter` without having them either rely on implicit conversion to a `IUrlParameter` implementation.

Mpdreamz:
Taking it to its logical conclusion and removing `Name` amongst others reminded me why we made this wrapped

```csharp
  GetAliasUrlTests.cs(29, 43): [CS0121] The call is ambiguous between the following methods or properties: 'GetAliasRequest.GetAliasRequest(Names)' and 'GetAliasRequest.GetAliasRequest(Indices)'
```

In cases where an API has multiple route that take a single string you quickly turn into these kinds of compiler issues.

Reverted allowing `string` and `long` directly

PR #3811, opened 2019-06-11
Invert document generation so it is not run by default.

### LINKED ISSUE #3674: Changes to build scripts
- Invert the `skipdocs` default parameter, so doc generation is always skipped
- Remove the IL rewriting for `7.x`

codebrain:
To generate documentation include the "gendocs" argument in the build command line, e.g.

`build.bat gendocs`

Fixes #3674

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,7 +6,7 @@ environment:
     DOTNET_CLI_TELEMETRY_OPTOUT: true
     DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
 build_script:
-    - cmd: build.bat canary skipdocs
+    - cmd: build.bat canary
 
 nuget:
     disable_publish_on_pr: true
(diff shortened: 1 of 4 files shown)


codebrain:
> I think we should still run docs when running with the default target e.g just calling `./build.sh` or `build.bat` so that we can easily gen the docs as part of a PR.

Ah, I was thinking that we remove it altogether and then if you want to generate docs you write:

`build.bat gendocs`

Does that seem reasonable @Mpdreamz ?

Mpdreamz:
`build.sh` or `build.bat` should be the command we run before submitting a PR to validate the build beyond actual compilation.

I think its best if the docs gen come through then as well.

Mpdreamz:
Updated and rebased this PR and made `gendocs` enabled when using the default `build` target

russcam:
LGTM. Merging this in

PR #3817, opened 2019-06-13
Refine generated documentation comments

russcam:
This commit refines the code generated XML comments.

- For namespaces, split pascal cased names.
- Use `<para>` instead of `<pre>`
- Replace usage of elasticsearch with Elasticsearch

### DIFF
--- a/src/Elasticsearch.Net/ElasticLowLevelClient.XPack.cs
+++ b/src/Elasticsearch.Net/ElasticLowLevelClient.XPack.cs
@@ -32,7 +32,7 @@ using static Elasticsearch.Net.HttpMethod;
 namespace Elasticsearch.Net.Specification.XPackApi
 {
 	///<summary>
-	/// X Pack APIs.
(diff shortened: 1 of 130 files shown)


src/Elasticsearch.Net/ElasticLowLevelClient.XPack.cs
```diff
@@ -32,8 +32,8 @@
 namespace Elasticsearch.Net.Specification.XPackApi
 {
 	///<summary>
-	/// Logically groups all XPack API's together so that they may be discovered more naturally.
-	/// <para>Not intended to be instantiated directly please defer to the <see cref = "IElasticLowLevelClient.XPack"/> property
+	/// X Pack APIs.
```
Mpdreamz: `XPack` :)

russcam: will fix :)

PR #3824, opened 2019-06-17
Feature/7.x/cluster state dynamic

Mpdreamz:
As part of https://github.com/elastic/elasticsearch/pull/38413

`cluster.state` is now documented to treat its json as `key/value` pairs where values can also be `object` or `array`.

This commit reflects that and exposes only the state as `DynamicBody` which is `Elasticsearch.Net` fail safe way to traverse json.

Opening as DRAFT to discuss first, will amend with the removal of the types later.


src/Elasticsearch.Net/Responses/Dynamic/DynamicBody.cs
```diff
@@ -51,19 +52,46 @@ public bool IsReadOnly
 		}
 
 		/// <summary>
-		/// Gets or sets the <see cref="ElasticsearchDynamicValue" /> with the specified name.
+		/// Helper to to easily traverse the data using a path notation
```
russcam:
```suggestion
		/// Traverses data using path notation
```

src/Elasticsearch.Net/Responses/Dynamic/DynamicBody.cs
```diff
@@ -51,19 +52,46 @@ public bool IsReadOnly
...
 		/// </summary>
-		/// <value>A <see cref="ElasticsearchDynamicValue" /> instance containing a value.</value>
-		public dynamic this[string name]
+		/// <param name="path">path into the stored object, keys are seperated with a dot and the last key is returned as T</param>
+		/// <typeparam name="T"></typeparam>
+		/// <returns>T or default</returns>
+		public T Get<T>(string path)
+		{
+			var stack = new Stack<string>(path.Split('.'));
```
russcam: does it need a guard for `null` path?

Mpdreamz: ++ added

PR #3825, opened 2019-06-18
Use Marshal.SecureStringToGlobalAllocUnicode

russcam:
Relates: https://github.com/elastic/elasticsearch-net/pull/3806#issuecomment-502700127

This commit updates the SecureStrings implementation to use the cross platform
Marshal methods that deal with unmanaged unicode strings.

Implement IDisposable on BasicAuthenticationCredentials and dispose of SecureString
instances when ConnectionConfiguration is disposed.


russcam:
Added a test for `SecureStrings`

RolfDeVries:
Thank you very much, the nightly build works.

PR #3827, opened 2019-06-18
Add index name to ICreateIndexResponse

### LINKED ISSUE #3823: CreateIndex doesn't return index name
6.8.0

**Elasticsearch version**:
6.8.0

**Describe the feature**:
Currently `CreateIndexAsync` returns an `ICreateIndexResponse` which includes the `ShardsAcknowledged` and `Acknowledged` properties but not the index name as returned in the json response:

```
{
  "acknowledged" : true,
  "shards_acknowledged" : true,
  "index" : "my-index-2019.06.17-000001"
}
```

This is useful when indices are created using Date Math patterns like `%3Cmy-index-%7Bnow%2Fd%7D-000001%3E` so the date doesn't has to be calculated again.

codebrain:
Fixes #3823

Port to `7.x` requires attention to new serialiser attributes.

### DIFF
--- a/src/Tests/Tests/Indices/IndexManagement/CreateIndex/CreateIndexApiTests.cs
+++ b/src/Tests/Tests/Indices/IndexManagement/CreateIndex/CreateIndexApiTests.cs
@@ -205,9 +205,7 @@ namespace Tests.Indices.IndexManagement.CreateIndex
 			response.ShouldBeValid();
 			response.Acknowledged.Should().BeTrue();
 			response.ShardsAcknowledged.Should().BeTrue();
-
-			if (Cluster.ClusterConfiguration.Version >= "6.8.0")
-				response.Index.Should().Be(CallIsolatedValue);
+			response.Index.Should().Be(CallIsolatedValue);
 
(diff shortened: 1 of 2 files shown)


src/Tests/Tests/Indices/IndexManagement/CreateIndex/CreateIndexApiTests.cs
```diff
@@ -206,6 +206,9 @@ protected override void ExpectResponse(ICreateIndexResponse response)
 			response.Acknowledged.Should().BeTrue();
 			response.ShardsAcknowledged.Should().BeTrue();
 
+			if (Cluster.ClusterConfiguration.Version >= "6.8.0")
```
russcam: index is returned in Elasticsearch 6.0.0, so this condition is not needed

codebrain: ++ will remove

russcam:
@codebrain Does this need to be ported to 7.x?

codebrain:
@russcam Yes. Done.

PR #3831, opened 2019-06-19
Return of net461

Mpdreamz:
- Reintroduced `HttpWebRequestConnection`, this makes the sync paths on full framework not
  rely on `GetAwaiter()` which is no good in environments with synchronization context (asp.net on full framework)

- reintroduced `net461` as `TFM`
  also better for packaging, `net461` relies on auto-magically including a whole bunch of system packages which is annoying to end users

- tests now run `net461` tests on windows

- `net461` support through multi targeting pack
  meaning you can build develop on the `net461` targets from `linux`/`osx` as well


build/Elasticsearch.Net.nuspec
```diff
@@ -16,18 +16,17 @@
 		<copyright>2014-$year$ Elasticsearch BV</copyright>
 		<tags>elasticsearch,elastic,search,lucene,nest</tags>
 		<dependencies>
-			<group targetFramework=".NETStandard2.0">
-				<dependency id="System.Buffers" version="[4.5.0,)"/>
-				<dependency id="System.Reflection.Emit" version="[4.3.0,)"/>
-				<dependency id="System.Reflection.Emit.Lightweight" version="[4.3.0,)"/>
-				<dependency id="System.Threading.Tasks.Extensions" version="[4.4.0,)"/>
-				<dependency id="System.ValueTuple" version="[4.4.0,)"/>
-			</group>
```
russcam: should these be removed?

Mpdreamz: All of these are part of netstandard2.0 

https://blog.monstuff.com/archives/2017/03/valuetuple-availability.html

Mpdreamz: I am wrong :cat: will update and introduce a separate `net47` TFM to prevent this:

https://stackoverflow.com/questions/53407579/prefer-net-standard-2-0-library-reference-over-net-4-5-1-from-a-net-4-7-2-pro

PR #3815, opened 2019-06-13
Collection deserialisation needs to check for null tokens

codebrain:
Sql query results are composed of an array of values, with elements that may be null.

Fixes #3813

PR #3833, opened 2019-06-20
Advance reader when not expected JSON token

russcam:
This commit advances the JSON reader when a JSON token is one that is unexpected e.g. when the JSON token is null.


russcam:
There's some overlap with #3815, although there may be one or two things in that PR that are not in this one

PR #3846, opened 2019-06-20
Fix/7.x/deps

Mpdreamz:
Updates to a proc that supports `SIGINT` on linux


azure-pipelines.yml
```diff
@@ -31,13 +31,32 @@ jobs:
...
+  dependsOn: LinuxUnitTests
+  pool:
+      vmImage: 'ubuntu-16.04'
+  strategy:
+      maxParallel: 5
+      matrix:
+          es7:
+              esVersion: '7.0.0'
+  steps:
+      - script: 'build.bat integrate $(esVersion) "readonly,writable" skipdocs'
```
russcam:
```suggestion
      - script: './build.sh integrate $(esVersion) "readonly,writable" skipdocs'
```

build/scripts/Tooling.fs
```diff
@@ -11,6 +12,15 @@ module Tooling =
     
     let private defaultTimeout = TimeSpan.FromMinutes(5.)
     
+    let startRedirectedInWithTimeout timeout workinDir bin args = 
+        let startArgs = StartArguments(bin, args |> List.toArray)
+        if (Option.isSome workinDir) then
+            startArgs.WorkingDirectory <- Option.defaultValue "" workinDir
+        let result = Proc.StartRedirected(startArgs, timeout, LineHighlightWriter())
+        if not result.Completed then failwithf "process failed to complete within %O: %s" timeout bin
+        let exitCode = match result.ExitCode.HasValue with | false -> None | true -> Some result.ExitCode.Value
```
russcam:
```suggestion
        let exitCode = if result.ExitCode.HasValue then Some result.ExitCode.Value else None 
```

PR #3867, opened 2019-06-24
Fix/7.x/clean up type extensions

Mpdreamz:
Make sure unused code with quite some complexity is removed


russcam:
the failing tests don't look related to this change. Merging in

PR #3869, opened 2019-06-24
DocCount no longer nullable on matrix stats aggregation

Mpdreamz:


Mpdreamz:
Integration test failures are unrelated, merging this in.

PR #3871, opened 2019-06-25
Feature/7.x/warnings as errors

Mpdreamz:
We now finally treat warnings as errors in the solution. 

We now also add a `ConfigureAwait(false)` roslyn Analyzer, the one we evaluated in the past has been updated and now works a treat!

Fixed the places we weren't calling `ConfigureAwait(false)`


Mpdreamz:
@codebrain @russcam used my admin discretion to self review and merge this one in. Too useful and would like to have it in during my working hours :sunglasses:

PR #3882, opened 2019-06-27
Return default when no bytes read from response streaming-no-response…

russcam:
…-body

This commit returns default(T) when no bytes are read from the response stream. Move the stream null and seek check into JsonSerializer.

Update index exists integration test to not disable direct streaming in order to test.


src/Tests/Tests.Core/ManagedElasticsearch/Clusters/DirectStreamingReadOnlyCluster.cs
```diff
@@ -0,0 +1,15 @@
+using Elastic.Managed.Ephemeral.Plugins;
+using Nest;
+using Tests.Core.ManagedElasticsearch.NodeSeeders;
+
+namespace Tests.Core.ManagedElasticsearch.Clusters
+{
+	public class DirectStreamingReadOnlyCluster : ClientTestClusterBase
```
codebrain: Seems a *bit* heavy handed, can we just override in the test?

(Although I can see a counter argument here that we may want to keep around for other tests).

src/Elasticsearch.Net/Utf8Json/JsonSerializer.cs
```diff
@@ -254,7 +255,11 @@ public static T Deserialize<T>(byte[] bytes, int offset)
 
         public static T Deserialize<T>(byte[] bytes, int offset, IJsonFormatterResolver resolver)
         {
-            if (resolver == null) resolver = DefaultResolver;
+			if (bytes == null || bytes.Length == 0)
```
codebrain: ++

PR #3881, opened 2019-06-27
Fix SecurityApiKey and ApplicationPrivilegesApi integration tests

codebrain:

PR #3883, opened 2019-06-27
Fix CatHelp NodeHotThreads integ tests due to mimetype mismatch 

Mpdreamz:
Fixes cathelp and nodehotthreads now being unsuccesful in the low level client due to mimetype mismatch.

Also improves SSL generation which was failing on my windows partition due to warnings emitted about forbidden API usage.


Mpdreamz:
Failing windows tests will be fixed by #3881

PR #3887, opened 2019-06-27
ResponseBuilder mimeType check is too strict on net461 

Mpdreamz:
as mimetype includes encoding when using HttpWebRequest

PR #3888, opened 2019-06-27
Fix/7.x/interface searchresponse

Mpdreamz:
This brings back `ISearchResponse` on `client.Search()` and related API's

Going forward this is the only the response interface we maintain.


Mpdreamz:
Integration test passed on netcoreapp 2.1 but failed on net461 because #3887 was not merged yet.

PR #3891, opened 2019-06-27
Ephemeral Cluster for tests now uses a SHA has for its cached folder

Mpdreamz:
Relates: https://github.com/elastic/elasticsearch-net-abstractions/commit/f75189735cb6d23dd685a2ad0003dea68bafc6dd

The ephemeral cluster generates a cached folder name based on the options including the chosen plugins which can be a very long list and cause max file path issues down the line.


Mpdreamz:
Merging this in, wanted to see linux integration tests were still happy

PR #3912, opened 2019-07-03
Implement Api Key authorisation on requests

codebrain:
It is possible to set both the `Basic` auth and `ApiKey` auth together.

Clearly this doesn't make sense to do, and to me this feels like an exception.

Should we throw an exception in the client? (which would only surface at runtime), or do we apply some kind of precedence in which auth to apply if both are set?

I also left in the `string` override as I think this might be a little easier for someone to use, if they are unable to find our `.CreateSecureString()` extension method.

### DIFF
--- a/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
+++ b/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
@@ -116,6 +116,7 @@ namespace Elasticsearch.Net
 		private readonly ElasticsearchUrlFormatter _urlFormatter;
 
 		private BasicAuthenticationCredentials _basicAuthCredentials;
+		private ApiKeyAuthenticationCredentials _apiKeyAuthentication;
 		private X509CertificateCollection _clientCertificates;
 		private Action<IApiCallDetails> _completedRequestHandler = DefaultCompletedRequestHandler;
 		private int _connectionLimit;
@@ -169,6 +170,7 @@ namespace Elasticsearch.Net
 
 		protected IElasticsearchSerializer UseThisRequestResponseSerializer { get; set; }
 		BasicAuthenticationCredentials IConnectionConfigurationValues.BasicAuthenticationCredentials => _basicAuthCredentials;
+		ApiKeyAuthenticationCredentials IConnectionConfigurationValues.ApiKeyAuthenticationCredentials => _apiKeyAuthentication;
 		SemaphoreSlim IConnectionConfigurationValues.BootstrapLock => _semaphore;
 		X509CertificateCollection IConnectionConfigurationValues.ClientCertificates => _clientCertificates;
 		IConnection IConnectionConfigurationValues.Connection => _connection;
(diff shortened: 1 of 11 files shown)


src/Elasticsearch.Net/Configuration/Security/ApiKeyAuthenticationCredentials.cs
```diff
@@ -0,0 +1,39 @@
...
+	/// <summary>
+	/// Credentials for Api Key Authentication
+	/// </summary>
+	public class ApiKeyAuthenticationCredentials : IDisposable
+	{
+		public ApiKeyAuthenticationCredentials()
+		{
+		}
+
+		public ApiKeyAuthenticationCredentials(string id, SecureString apiKey)
```
Mpdreamz: Create an overload that takes the full base64 encoded apiKey as well I reckon. This is what folks'll typically copy paste.

src/Elasticsearch.Net/Connection/HttpWebRequestConnection.cs
```diff
@@ -258,6 +260,19 @@ protected virtual void SetBasicAuthenticationIfNeeded(HttpWebRequest request, Re
...
+		{
+			// ApiKey auth credentials take the following precedence (highest -> lowest):
+			// 1 - Specified on the request (highest precedence)
+			// 2 - Specified at the global IConnectionSettings level
+
+			string apiKey = null;
+			if (requestData.ApiKeyAuthenticationCredentials != null)
+				apiKey = $"{requestData.ApiKeyAuthenticationCredentials.Id}:{requestData.ApiKeyAuthenticationCredentials.ApiKey.CreateString()}";
+
+			if (!string.IsNullOrWhiteSpace(apiKey))
```
Mpdreamz: Move base64'ing into `ApiKeyAuthenticationCredentials`

src/Elasticsearch.Net/Configuration/Security/ApiKeyAuthenticationCredentials.cs
codebrain: Will set that as a `SecureString` type

src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
```diff
@@ -434,6 +436,18 @@ protected ConnectionConfiguration(IConnectionPool connectionPool, IConnection co
...
+		/// <summary>
+		/// Api Key to send with all requests to Elasticsearch
+		/// </summary>
+		public T ApiKeyAuthentication(string id, SecureString apiKey) =>
+			Assign(new ApiKeyAuthenticationCredentials(id, apiKey), (a, v) => a._apiKeyAuthentication = v);
+
+		/// <summary>
+		/// Api Key to send with all requests to Elasticsearch
+		/// </summary>
+		public T ApiKeyAuthentication(string id, string apiKey) =>
```
russcam: Kept string password for BasicAuthentication in 7.x to ease the move from 6.x to 7.x, but wonder if we should we only support `SecureString` for apikey going forward?

src/Elasticsearch.Net/Connection/HttpWebRequestConnection.cs
```diff
@@ -239,7 +240,16 @@ protected virtual void SetProxyIfNeeded(HttpWebRequest request, RequestData requ
...
+		private void SetAuthenticationIfNeeded(RequestData requestData, HttpWebRequest request)
+		{
+			// Api Key authentication takes precedence
+			var apiKeySet = SetApiKeyAuthenticationIfNeeded(request, requestData);
+
+			if (!apiKeySet)
+				SetBasicAuthenticationIfNeeded(request, requestData);
+		}
+
+		protected virtual bool SetBasicAuthenticationIfNeeded(HttpWebRequest request, RequestData requestData)
```
russcam: public API breaking change going from `void` to `bool`

src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
codebrain: Would prefer to keep supporting `string` for ease of use.

PR #3918, opened 2019-07-04
Obsolete AllField and IndexField

### LINKED ISSUE #3905: TypeMappingDescriptor<T> still supports setting AllField
7.0.0/7.0.0

**Elasticsearch version**:
7.2.0

TypeMappingDescriptor<T> still has the AllField method even though using it results in a 400 when creating an index. The method should maybe have been marked Obsolete in the 6.X version and removed in 7.X.
**Steps to reproduce**:
 1. Define a CreateIndexDescriptor and in the Map section set AllField to disabled
 2. Attempt to create the index.

```
            var testDescriptor = new CreateIndexDescriptor("test")
                .Map(m => m
                    .AllField(af => af.Enabled(false))
                    .AutoMap());
            var result = this.client.Indices.Create(testDescriptor);

```

Root mapping definition has unsupported parameters:  [_all : {enabled=false}]

russcam:
This commit obsoletes the AllField and IndexField. Both should have been removed in 7.x but were missed.
Remove the DataMemberAttribute and add IgnoreDataMemberAttribute so that any value is not
sent in a request.

Fixes #3905

### DIFF
--- a/src/Nest/Indices/MappingManagement/GetFieldMapping/FieldMappingFormatter.cs
+++ b/src/Nest/Indices/MappingManagement/GetFieldMapping/FieldMappingFormatter.cs
@@ -37,8 +37,10 @@ namespace Nest
 					switch (value)
 					{
 						case 0:
+#pragma warning disable 618
 							mapping = formatterResolver.GetFormatter<AllField>()
 								.Deserialize(ref reader, formatterResolver);
+#pragma warning restore 618
 							break;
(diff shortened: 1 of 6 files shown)


russcam:
rebased on top of master and force pushed.

russcam:
ported to 7.1: 8abfe065f774204ca2308531bb5ef0b84fa5e95f

PR #3925, opened 2019-07-05
Do not set HttpRequestMessage content when no PostData

### LINKED ISSUE #3907: Version 7.0.0 not working with os httphandlers
If I am using 
  AppContext.SetSwitch("System.Net.Http.UseSocketsHttpHandler", false);
with version 7.0.0 at least Indices api Exists doesnt work correctly. if I am using 7.0.0-beta1 everything is ok.

Here is code to reproduce
{
    AppContext.SetSwitch("System.Net.Http.UseSocketsHttpHandler", false);
    var defaultIndex = "nonexist_index";
    var pool = new SingleNodeConnectionPool(new Uri("http://localhost:9210"));

    var settings = new ConnectionSettings(pool)
        .DefaultIndex(defaultIndex)
        .DisableDirectStreaming()
        .PrettyJson();

    var client = new ElasticClient(settings);

    var response = client.Indices.Exists(defaultIndex);
    if (response.ApiCall.HttpStatusCode != (int)HttpStatusCode.NotFound)
       throw new Exception("Should be 404 status code.");
}

russcam:
This commit changes the HttpConnection to not set Content on HttpRequestMessage when there is no PostData. The check is performed inside of Request and RequestAsync to avoid creating any content instance and allocations.

Change length to default in TryComputeLength to match HttpContent implementation.

Pass RequestData through to _onStreamAvailable delegates so that HttpCompression, ConnectionSettings and PostData can be accessed on the passed instance.

Fixes #3907

### DIFF
--- a/src/Elasticsearch.Net/Connection/Content/RequestDataContent.cs
+++ b/src/Elasticsearch.Net/Connection/Content/RequestDataContent.cs
@@ -91,7 +91,7 @@ namespace Elasticsearch.Net
 		protected override bool TryComputeLength(out long length)
 		{
 			// We can't know the length of the content being pushed to the output stream.
-			length = default;
+			length = -1;
 			return false;
 		}
 
@@ -112,11 +112,7 @@ namespace Elasticsearch.Net
 				base.Dispose();
 			}
 
-			public override void Close()
-			{
-				_serializeToStreamTask.TrySetResult(true);
-				base.Close();
-			}
+			public override void Close() => _serializeToStreamTask.TrySetResult(true);
 		}
 
 		/// <summary>
--- a/src/Elasticsearch.Net/Connection/Content/RequestDataContent.cs
+++ b/src/Elasticsearch.Net/Connection/Content/RequestDataContent.cs
@@ -25,8 +25,7 @@ namespace Elasticsearch.Net
 	internal class RequestDataContent : HttpContent
 	{
 		private readonly RequestData _requestData;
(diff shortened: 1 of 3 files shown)


src/Elasticsearch.Net/Connection/Content/RequestDataContent.cs
```diff
@@ -195,6 +197,8 @@ protected override void Dispose(bool disposing)
 			public override void EndWrite(IAsyncResult asyncResult) => _innerStream.EndWrite(asyncResult);
 
 			public override void WriteByte(byte value) => _innerStream.WriteByte(value);
+
+			public override void Close() => _innerStream.Close();
```
Mpdreamz: https://github.com/dotnet/corefx/blob/master/src/Common/src/System/IO/DelegatingStream.cs

Does not implement this, I think its good to have though.

src/Tests/Tests.Reproduce/GithubIssue3907.cs
```diff
@@ -0,0 +1,30 @@
...
+	{
+		private readonly IntrusiveOperationCluster _cluster;
+
+		// use intrusive operation because we're changing the underlying http handler
+		public GithubIssue3907(IntrusiveOperationCluster cluster) => _cluster = cluster;
+
+		[I]
+		public void NotUsingSocketsHttpHandlerDoesNotCauseException()
+		{
+			AppContext.SetSwitch("System.Net.Http.UseSocketsHttpHandler", false);
```
Mpdreamz: Add comment this is safe to do because `IntrusiveOperationCluster` runs with max concurrency of 1

PR #3933, opened 2019-07-09
Date Nanos implementation

codebrain:
Add nanosecond field mapper

https://github.com/elastic/elasticsearch/issues/37755

### DIFF
--- a/src/Nest/Mapping/DynamicTemplate/SingleMapping.cs
+++ b/src/Nest/Mapping/DynamicTemplate/SingleMapping.cs
@@ -18,6 +18,9 @@ namespace Nest
 		public IProperty Date(Func<DatePropertyDescriptor<T>, IDateProperty> selector) =>
 			selector?.Invoke(new DatePropertyDescriptor<T>());
 
+		public IProperty DateNanos(Func<DateNanosPropertyDescriptor<T>, IDateNanosProperty> selector) =>
+			selector?.Invoke(new DateNanosPropertyDescriptor<T>());
+
 		public IProperty DateRange(Func<DateRangePropertyDescriptor<T>, IDateRangeProperty> selector) =>
 			selector?.Invoke(new DateRangePropertyDescriptor<T>());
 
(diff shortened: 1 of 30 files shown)


src/Nest/Mapping/Types/Core/DateNanos/DateNanosProperty.cs
```diff
@@ -0,0 +1,109 @@
...
+	public interface IDateNanosProperty : IDocValuesProperty
+	{
+		/// <summary>
+		/// Mapping field-level query time boosting. Accepts a floating point number, defaults to 1.0.
+		/// </summary>
+		[DataMember(Name = "boost")]
+		double? Boost { get; set; }
+
+		[DataMember(Name = "fielddata")]
+		INumericFielddata Fielddata { get; set; }
```
russcam: Looking at the [`DateFieldMapper`](https://github.com/elastic/elasticsearch/blob/3400483af46189435e6039806dbee6b8fd9a8532/server/src/main/java/org/elasticsearch/index/mapper/DateFieldMapper.java), it doesn't look like this is supported. 

It looks like `INumericFielddata` should also be removed (via deprecated in 7.x), see https://github.com/elastic/elasticsearch/commit/e50eeeaffb33d7eaa0f7c2ebc71f1208460f4fcd#diff-abb176481762ceeebb86fa4c02c13993L79

russcam:
@codebrain this needs some tests too, similar to https://github.com/elastic/elasticsearch-net/blob/9f0d285904d4c2dc523b8bbc835535662d83f547/src/Tests/Tests/Mapping/Types/Core/Date/DatePropertyTests.cs

src/Nest/Mapping/Types/Core/DateNanos/DateNanosProperty.cs
```diff
@@ -0,0 +1,107 @@
...
+		string IDateNanosProperty.Format { get; set; }
+
+		/// <inheritdoc cref="IDateNanosProperty.IgnoreMalformed"/>
+		bool? IDateNanosProperty.IgnoreMalformed { get; set; }
+
+		/// <inheritdoc cref="IDateNanosProperty.Index"/>
+		bool? IDateNanosProperty.Index { get; set; }
+
+		/// <inheritdoc cref="IDateNanosProperty.NullValue"/>
+		DateTime? IDateNanosProperty.NullValue { get; set; }
```
russcam: explicit interface member implementations don't need `/// <inheritdoc />` as they already have the doc 😃

src/Nest/Mapping/DynamicTemplate/SingleMapping.cs
```diff
@@ -247,6 +250,30 @@ public IProperty Object<TChild>(Func<ObjectTypeDescriptor<T, TChild>, IObjectPro
 		) =>
 			selector.InvokeOrDefault(new DatePropertyDescriptor<T>().Name(field));
 
+		public IProperty ScalarNanos(Expression<Func<T, DateTime>> field, Func<DateNanosPropertyDescriptor<T>, IDateNanosProperty> selector = null) =>
```
codebrain: @russcam - not sure if you saw this, but I had to rename the method to avoid conflicts with the implementation for `Date`. As `date_nanos` and `date` types both use `DateTime` objects in C# there are places where the two property types conflict when method resolution is attempted by the compiler.

russcam: saw it 🙂 unavoidable here, I think

PR #3958, opened 2019-07-17
Fix default chunked transfer encoding and expose as configurable

### LINKED ISSUE #3952: Transfer-Encoding: chunked is enabled by default
**NEST/Elasticsearch.Net version**: 7.0.1

**Elasticsearch version**: 7.0

Since upgrading from Elasticsearch.Net 7.0.0.alpha to the most current version, all requests made to Elasticsearch are sent with the header "Transfer-Encoding: chunked". Since our Elasticsearch server is behind a reverse proxy, it can no longer pass on the requests. It worked flawlessly with versions 6 and 7 alpha, since this header wasn't included in requests sent through those versions.

I have found no way of disabling this header. Not through the GlobalHeaders() method on the connection settings, nor through a custom HttpConnection. The header is re-added/overwritten at all times.

Is this by design? Is there a workaround available?

russcam:
This PR reverts the behaviour of the client to not use chunked tranfer encoding by default, and exposes chunked transfer encoding as a configuration value on IConnectionConfigurationValues and RequestConfiguration.

With the introduction of the pull based streaming model in RequestDataContent, a ContentLength value cannot be computed for what will be written to the request stream; a ContentLength value is attempted to be determined, and TryComputeLength() executed, _before_ PostData is serialized to the request stream, making it not possible to provide a ContentLength value in the implementation, and falling into the HttpClient behaviour of using chunked transfer encoding.

A potential fix could have been applied in RequestDataContent, to serialize PostData ahead of time and store in a byte array or stream field, and the logical place to do this would have been the ctor. It was determined however that this would be less than optimal to do because

1. It starts to overcomplicate the RequestDataContent implementation
2. An implementation in the RequestDataContent ctor would not have been able to use aynchronous methods for the RequestAsync<T> asynchronous path.

Instead, the serialization of PostData happens in sync and async paths in HttpConnection, using ByteArrayContent and the written request bytes when DisabledDirectStreaming is enabled to avoid keeping two copies of the request bytes around.

Additional integration tests added for HttpConnection and HttpWebRequestConnection to assert behaviour.

Fixes #3952

### DIFF
--- a/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
+++ b/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
@@ -145,6 +145,7 @@ namespace Elasticsearch.Net
 		private bool _sniffOnConnectionFault;
 		private bool _sniffOnStartup;
 		private bool _throwExceptions;
+		private bool _transferEncodingChunked;
 
 		private string _userAgent = ConnectionConfiguration.DefaultUserAgent;
 		private Func<HttpMethod, int, bool> _statusCodeToResponseSuccess;
@@ -213,6 +214,7 @@ namespace Elasticsearch.Net
 		ElasticsearchUrlFormatter IConnectionConfigurationValues.UrlFormatter => _urlFormatter;
 		string IConnectionConfigurationValues.UserAgent => _userAgent;
 		Func<HttpMethod, int, bool> IConnectionConfigurationValues.StatusCodeToResponseSuccess => _statusCodeToResponseSuccess;
+		bool IConnectionConfigurationValues.TransferEncodingChunked => _transferEncodingChunked;
 
 		void IDisposable.Dispose() => DisposeManagedResources();
 
@@ -529,6 +531,11 @@ namespace Elasticsearch.Net
 		/// versions that initiate requests to Elasticsearch
 		/// </summary>
 		public T UserAgent(string userAgent) => Assign(userAgent, (a, v) => a._userAgent = v);
+
+		/// <summary>
+		/// Whether the request should be sent with chunked Transfer-Encoding. Default is <c>true</c>
+		/// </summary>
+		public T TransferEncodingChunked(bool transferEncodingChunked = true) => Assign(transferEncodingChunked, (a, v) => a._transferEncodingChunked = v);
 
 		protected virtual void DisposeManagedResources()
 		{
--- a/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
+++ b/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
@@ -533,7 +533,7 @@ namespace Elasticsearch.Net
 		public T UserAgent(string userAgent) => Assign(userAgent, (a, v) => a._userAgent = v);
 
 		/// <summary>
-		/// Whether the request should be sent with chunked Transfer-Encoding. Default is <c>true</c>
+		/// Whether the request should be sent with chunked Transfer-Encoding. Default is <c>false</c>
 		/// </summary>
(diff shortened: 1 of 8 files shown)


src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
```diff
@@ -530,6 +532,11 @@ protected ConnectionConfiguration(IConnectionPool connectionPool, IConnection co
 		/// </summary>
 		public T UserAgent(string userAgent) => Assign(userAgent, (a, v) => a._userAgent = v);
 
+		/// <summary>
+		/// Whether the request should be sent with chunked Transfer-Encoding. Default is <c>true</c>
```
russcam:
```suggestion
		/// Whether the request should be sent with chunked Transfer-Encoding. Default is <c>false</c>
```

src/Tests/Tests/ClientConcepts/Connection/HttpWebRequestConnectionTests.cs
```diff
@@ -0,0 +1,106 @@
...
+			connection.Request<StringResponse>(requestData);
+			connection.LastRequest.SendChunked.Should().BeTrue();
+			connection.LastRequest.ContentLength.Should().Be(-1);
+
+			await connection.RequestAsync<StringResponse>(requestData, CancellationToken.None).ConfigureAwait(false);
+			connection.LastRequest.SendChunked.Should().BeTrue();
+			connection.LastRequest.ContentLength.Should().Be(-1);
+		}
+
+		[I] public async Task HttpClientSetsContentLengthWhenTransferEncodingChunkedFalse()
```
russcam:
```suggestion
		[I] public async Task HttpWebRequestSetsContentLengthWhenTransferEncodingChunkedFalse()
```

src/Tests/Tests/ClientConcepts/Connection/HttpWebRequestConnectionTests.cs
```diff
@@ -0,0 +1,106 @@
...
+				connectionSettings.Proxy(proxyAddress, null, (string)null);
+
+			var requestData = new RequestData(HttpMethod.POST, "/_search", "{ \"query\": { \"match_all\" : { } } }", connectionSettings,
+				new SearchRequestParameters(),
+				new MemoryStreamFactory()) { Node = node };
+
+			return requestData;
+		}
+
+		[I] public async Task HttpClientUseTransferEncodingChunkedWhenTransferEncodingChunkedTrue()
```
russcam:
```suggestion
		[I] public async Task HttpWebRequestUseTransferEncodingChunkedWhenTransferEncodingChunkedTrue()
```

src/Elasticsearch.Net/Connection/HttpConnection.cs
```diff
@@ -306,10 +307,71 @@ protected virtual HttpRequestMessage CreateRequestMessage(RequestData requestDat
...
+					message.Content = new ByteArrayContent(requestData.PostData.WrittenBytes);
+					stream.Dispose();
+				}
+				else
+				{
+					stream.Position = 0;
+					message.Content = new StreamContent(stream);
+				}
+
+				if (requestData.HttpCompression)
```
codebrain: There is a similar `if` on line `352` - can they be merged?

russcam: no they cannot 😃 this `if` sets `gzip` Content Encoding on `message.Content` headers, which is set on line 360 or 366, after line 352.

codebrain: I wonder if this needs a comment to stop the eager refactor! :)

russcam: If you refactor and then run tests, they will fail - is that reasonable discouragement enough?

PR #3975, opened 2019-07-30
Connection pool ping honours relative path

### LINKED ISSUE #3970: Bitnami image does not work without DisablePing

`7.1`

**Elasticsearch version**:

`7.0.0`

The problem exists when Elasticsearch endpoint is not exposed on the host root path `/` but under `/elasticsearch` like on Bitnami ELK image.

Invoking any client method, like `client.Indices.Exists`, throws an exception `Failed to ping the specified node.. Call: Status code 200 from HEAD /`. Beside status code `200` response mime type is `text/html` which indicated that client tries to ping host without path included and gets Bitnami default landing page.

By debugging through ELK http logs assumption was confirmed. GET call was fired to `http://HOST` instead of `http://HOST/elasticsearch/` as configured. 

**Steps to reproduce**:
 1. Bootstrap Bitnami ELK image
 2. Instantiate client w/o calling `DisablePing` method
 3. Call `client.Indices.Exists` method for any index (existing or non-existing)

*** Workaround ***

Call `DisablePing` on the client instance before any Api Call.

russcam:
This commit updates the RequestPipeline ping operation to use an empty string for the ping relative path. Using a relative
path that starts with "/" will overwrite any base path when combined with the node URI.

Fixes #3970

### DIFF
--- a/src/Elasticsearch.Net/Transport/Pipeline/RequestPipeline.cs
+++ b/src/Elasticsearch.Net/Transport/Pipeline/RequestPipeline.cs
@@ -561,7 +561,7 @@ namespace Elasticsearch.Net
 			IRequestParameters requestParameters = new RootNodeInfoRequestParameters();
 			requestParameters.RequestConfiguration = requestOverrides;
 
-			var data = new RequestData(HttpMethod.HEAD, "/", null, _settings, requestParameters, _memoryStreamFactory) { Node = node };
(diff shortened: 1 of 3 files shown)


russcam:
ported to 7.x in 097b8c2ddd165cf9d0e8db2acfde849a7800dad3

Issue #3931, opened 2019-07-08
Enhance documentation around compatibility and clarify client roadmap


andrewrimmer:
It would be useful to provide clarification on what versions of Elasticsearch are supported by which version of the .NET client.

The compatibility matrix is pretty clear that the .NET client major version should match the major version of the Elastic cluster. However it is not clear from the documentation what support there is across minor versions. 

My assumption has been that there would be very few breaking changes between minor releases of Elastic and as such version 7.x of the .NET client should work against any version of the Elastic cluster that is equal or greater than the Elastic client version. This assumption has proven to be incorrect. @codebrain clarified that the 7.0 .NET client is not recommended to be used against a 7.1 or 7.2 cluster.

This brings up many different questions.

1. Should the minor version of the .NET client always match the elastic cluster? i.e. 7.0.x should be used against 7.0.x clusters? (Maybe this is implied in the compatibility matrix, but perhaps could be made clearer?)
2. When does trunk development (and automated testing) move to the next Elastic version?
3. Will the next supported version of Elastic cluster be 7.1 or 7.2?
4. Is there a roadmap for the .NET client and when current versions of Elastic cluster will be supported?
5. Are the loose targets for how long the .NET client will lag behind the Elastic cluster?
6. How dangerous is using a .NET client that is behind the minor version of the elastic cluster
7. Would it be useful to have the automated integration tests run against the latest current release of Elastic? This could give developers a better idea of what should work should they choose (have) to run a .NET client against an unsupported version of Elastic.

## Current Elasticsearch 7 .net support?
<table>
    <tr>
        <th><b>Version<b></th>
        <th><b>Release Date<b></th>
        <th><b>.NET support?<b></th>
    </tr>
    <tr>
    	<td><code>7.0</code></td>
    	<td>10th April 2019</td>
    	<td>2nd July 2019</td>
    </tr>
    <tr>
    	<td><code>7.1</code></td>
    	<td>20th May 2019</td>
    	<td>:x:</td>
    </tr>
    <tr>
    	<td><code>7.2</code></td>
    	<td>25th June 2019</td>
    	<td>:x:</td>
    </tr>
</table>

I think if you were new to Elasticsearch and started development right now there would be a good chance you would install the latest and greatest Elastic (7.2) and then install 7.0 of NEST and assume everything would be fine.

[...]

PR #3978, opened 2019-07-31
Update homepage to give further information on compatibility questions

codebrain:
Issue #3931 raised some questions around compatibility between the client and server versions. This new section addresses those questions raised and also aims to quell similar future questions.

Requires porting.

Thanks to @lcawl for reviewing.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -3,33 +3,18 @@ Repository for both **NEST** and **Elasticsearch.Net**, the two official [Elasti
 ## Compatibility Matrix
 <table>
     <tr>
+        <th><b>.NET Clients<b></th>
         <th><b>Elasticsearch<b></th>
-        <th><b>Clients<b></th>
         <th><b>Supported<b></th>
         <th><b>Windows/Linux CI</b></th>
         <th><b>Tests<b></th>
     </tr>
     <tr>
-    	<td><code>0.x</code></td>
-    	<td><code>0.x</code></td>
+    	<td><code>0.x - 2.x</code></td>
+    	<td><code>0.x - 2.x</code></td>
     	<td>:x:</td>
     	<td>:heavy_minus_sign:</td>
     	<td>:heavy_minus_sign:</td>
-    </tr>
-    <tr>
-    	<td><code>1.x</code></td>
-    	<td><code>1.x</code></td>
-    	<td>:x:</td>
-    	<td>:heavy_minus_sign:</td>
-    	<td>:heavy_minus_sign:</td>
-    </tr>
-    <tr>
-    	<td><code>2.x</code></td>
-    	<td><code>2.x</code></td>
-    	<td>:x:</td>
-    	<td>:heavy_minus_sign:</td>
-    	<td>:heavy_minus_sign:</td>
-    	</td>  
     </tr>
     <tr>
     	<td><code>5.x</code></td>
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -61,6 +46,38 @@ Repository for both **NEST** and **Elasticsearch.Net**, the two official [Elasti
...
+
+*Using a `7.0.0` client against a `7.0.1` - `7.2.0` server*. There is client support for all comparable features that are available in the `7.0.0` server. If you want to use new features introduced in `7.2.0` with a lower version client, then you will have to use the low level client `DoRequest` method and perform the request/response (de)serialisation yourself.
+
+When we release a client we will run the unit and integration tests against the latest minor patch version of Elasticsearch that matches the major.minor of the client release, and then any other latest minors within that major version.
+
+Any incompatibilities between minor versions are documented against the release.
+
+#### I have a `6.0` server, what client should I use?
+
+Always use the latest minor version of the client within that major version, so in this instance version `6.8`. The reason being is that `6.8` will contain many bug fixes not present in the `6.0` version of the client.
```
russcam:
```suggestion
Always use the latest minor version of the client within that major version, so in this instance, at time of writing, this is version `6.8.0`. The reason being is that `6.8.0` will contain many bug fixes not present in the `6.0.0` version of the client.
```

readme.md
```diff
@@ -61,6 +46,38 @@ Repository for both **NEST** and **Elasticsearch.Net**, the two official [Elasti
...
+
+#### Branch Compatibility
+
+`master` reflects the latest server version (`8.x` snapshots)
+`7.x` reflects the latest development of the `7.x` client
+`7.1` is the integration branch for the next client release (`7.1.0`)
```
russcam: I would be inclined to keep ephemeral minor integration branches out of the README

readme.md
```diff
@@ -3,34 +3,19 @@ Repository for both **NEST** and **Elasticsearch.Net**, the two official [Elasti
...
-        <th><b>Clients<b></th>
         <th><b>Supported<b></th>
         <th><b>Windows/Linux CI</b></th>
         <th><b>Tests<b></th>
     </tr>
     <tr>
-    	<td><code>0.x</code></td>
-    	<td><code>0.x</code></td>
+    	<td><code>0.x - 2.x</code></td>
+    	<td><code>0.x - 2.x</code></td>
```
russcam: I would keep these as separate major versions, to illustrate that 1.x client -> 1.x ES, etc.

readme.md
codebrain: Could we clarify "at time of writing" as before?

readme.md
codebrain: I think its fairly obvious from the rest of the table?

russcam: I appreciate what you're saying, but is there harm in being completely explicit, aside from the extra table rows, since the aim is to provide guidance for other version -> version scenarios?

readme.md
russcam: This will require updating with each minor release - is that something we want to do?

Mpdreamz:
```
`master` reflects the latest server version, this is typically the `current latest major + 1`
`N.x` where N represents the major version component of the Elasticsearch server release its integrating with e,g `7.x`
`N.Y` where `N` is the major version and `Y` is  the minor component, typically opened as integration branch for a specific minor leaving `N.x` free to do bug fix releases.
```

This is the reverse to what Elasticsearch does though, propose we keep `N.x` as the latest `minor` integration branch and open a `N.Y` branch to only backport issues that warrant a bug fix release. We can delete any `< N.Y` branches when we create a new one.

readme.md
```diff
@@ -139,7 +156,7 @@ var client = new ElasticClient(settings);
 
 ### Indexing
 
-Indexing a document is as simple as (with 6.x):
+Indexing a document is as simple as (with 7.x):
```
Mpdreamz:
```suggestion
Indexing a document is as simple as:
```

readme.md
Mpdreamz: This might need duplicating in the `CONTRIBUTING.md`

PR #3987, opened 2019-08-02
Fix CI resolving to bundled JDK

Mpdreamz:
This works  currently works locally but not on azure devops where it falls back to the installed JRE8 which the snapshots build no longer accepts as a valid java version.

This PR updates abstractions to version that prints JAVA_HOME just before starting elasticsearch and forces it on the Process instance


src/CodeGeneration/ApiGenerator/Views/LowLevel/RequestParameters/RequestParameters.cshtml
```diff
@@ -26,12 +26,14 @@ namespace Elasticsearch.Net@(ns)
...
 	{
 		var r = endpoint.RequestParameterImplementation;
+		var supportsBody = endpoint.Body != null;
 		var names = r.CsharpNames;
 <text>
 	///<summary>Request options for @names.MethodName <para>@r.OfficialDocumentationLink</para></summary>
 	public class @names.ParametersName : RequestParameters<@names.ParametersName> 
 	{
 		public override HttpMethod DefaultHttpMethod => HttpMethod.@r.HttpMethod;
+		public override bool SupportsBody => @(supportsBody ? "true" : "false");
```
russcam: Can this just be `@supportsBody`?

Mpdreamz:
https://github.com/elastic/elasticsearch/issues/44902

This will now raise an error if a body was sent but was left unconsumed by the server which is a great addition.

`SupportsBody` now makes it way to our codegen to make sure we never sent a body if no body is documented.

src/CodeGeneration/ApiGenerator/Views/LowLevel/RequestParameters/RequestParameters.cshtml
Mpdreamz: It can will follow up :)

PR #3992, opened 2019-08-06
Expose external refreshes through the stats API

codebrain:
Expose external refreshes through the stats API

### DIFF
--- a/src/Nest/CommonOptions/Stats/RefreshStats.cs
+++ b/src/Nest/CommonOptions/Stats/RefreshStats.cs
@@ -13,5 +13,11 @@ namespace Nest
 
 		[DataMember(Name ="total_time_in_millis")]
 		public long TotalTimeInMilliseconds { get; set; }
+
+		[DataMember(Name ="external_total")]
+		public long ExternalTotal { get; set; }
+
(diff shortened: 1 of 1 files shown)


src/Nest/CommonOptions/Stats/RefreshStats.cs
```diff
@@ -13,5 +13,11 @@ public class RefreshStats
 
 		[DataMember(Name ="total_time_in_millis")]
 		public long TotalTimeInMilliseconds { get; set; }
+
+		[DataMember(Name ="external_total")]
+		public long ExternalTotal { get; set; }
```
russcam: minor nitpick: XML doc to indicate that this is only valid for Elasticsearch 7.2.0+

src/Nest/CommonOptions/Stats/RefreshStats.cs
```diff
@@ -13,5 +13,11 @@ public class RefreshStats
...
+
+		[DataMember(Name ="external_total_time_in_millis")]
+		public long ExternalTotalTimeInMilliseconds { get; set; }
```
russcam: minor nitpick: XML doc to indicate that this is only valid for Elasticsearch 7.2.0+

PR #3998, opened 2019-08-07
Add numeric_type sorting option.

codebrain:
For numeric fields it is also possible to cast the values from one type to another using this option. This can be useful for cross-index search if the sort field is mapped differently on some indices.

### DIFF
--- a/src/Tests/Tests/Search/Request/SortUsageTests.cs
+++ b/src/Tests/Tests/Search/Request/SortUsageTests.cs
@@ -333,7 +333,7 @@ namespace Tests.Search.Request
 	}
 
 	//hide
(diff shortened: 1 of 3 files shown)


src/Tests/Tests/Search/Request/SortUsageTests.cs
```diff
@@ -331,4 +331,38 @@ public class GeoDistanceIgnoreUnmappedUsageTests : SearchUsageTestBase
 				}
 			};
 	}
+
+	//hide
+	[SkipVersion("<7..0", "numeric_type added in 7.2.0")]
```
Mpdreamz:
```suggestion
	[SkipVersion("<7.2.0", "numeric_type added in 7.2.0")]
```

PR #4004, opened 2019-08-07
Feature/7.x/remove upgrade assistant

Mpdreamz:
This was a one off release. 

We need to decide if we want this removed with 7.2 or 7.3 though.


Mpdreamz:
Test failures:

### 7.0.0

```
build.bat seed:9368 integrate 7.0.0 "writable" "rolloverindex,putpipeline"
````

Rolloverindex test fails due to a mapping problem with the random data on this seed that needs looking into.

PutPipeline failed due to cluster `ProcessClusterEventTimeoutException` which I've seen happen a couple of times before with this test.


### 7.0.0-SNAPSHOT

```
Failed   Tests.Indices.IndexManagement.SplitIndex.SplitIndexApiTests.ReturnsExpectedResponse
Error Message:
 Tests.Framework.EndpointTests.ResponseAssertionException : Expected response.ShardsAcknowledged to be true, but found False.
Response Under Test:
Valid NEST response built from a successful (200) low level call on PUT: /nest-fluentasync-77a092ae/_split/nest-fluentasync-77a092ae-target?pretty=true&error_trace=true
 - [1] HealthyResponse: Node: http://localhost:9200/ Took: 00:00:30.5230768
# Request:
{"settings":{"index.number_of_shards":8}}
# Response:
{
  "acknowledged" : true,
  "shards_acknowledged" : false,
  "index" : "nest-fluentasync-77a092ae-target"
}
```

`shards_acknowledged: false` is not common here, need to understand when `acknowledged` and `shards_acknowledged` differ and what this means.

Will create new tickets for these.

Mpdreamz:
Merging this in on my own accord @codebrain @russcam

Mpdreamz:
This is now also cherry picked to `master`

PR #3954, opened 2019-07-17
Check backOffRetries before retrying in HandleBulkRequest

mb1337:
BulkAll will hang indefinitely, giving no error, if a bulk request keeps returning BadResponse. The bulk all request should throw an error after hitting the back off retry number.

PR #4014, opened 2019-08-09
BulkAll() improvements

Mpdreamz:
This PR fixes a bug and improves the run time behaviour of `BulkAll()`

* If the `_bulk` operation itself keeps returning an invalid status code the `BulkAll()` could potentially hang, As reported on #3954 This is rare since `_bulk` tends to return 200 even if all items fail but a problem nonetheless.

* `ForEachAsync()` should bail out after receiving a faulted task. This was preventing exceptions thrown from halting the `BulkAll()` in a controlled fashion.

* Added a callback for every `BulkResponse` received. `OnNext()` is for being notified when a batch succeeded but that might constitute multiple `Bulk()` calls.


codebrain:
Ported to `7.x` branch.

russcam:
ported to 6.x

PR #4017, opened 2019-08-12
Add XML docs for index inference on BulkAll

### LINKED ISSUE #3951: BulkAll requires DefaultIndex or DefaultMappingFor with index
From: https://discuss.elastic.co/t/nest-client-bulkall-requires-defaultindexfor-or-defaultindex-to-work/189766

The `BulkAll` method of `ElasticClient` requires 

- `DefaultMappingFor<T>()` with Index 

or 

- `DefaultIndex()`

or

- `.Index()` on the BulkAll call

to work. 

You'll get an `ArgumentException` otherwise : 

```
Index name is null for the given type and no default index is set. Map an index name using ConnectionSettings.DefaultMappingFor<TDocument>() or set a default index using ConnectionSettings.DefaultIndex()
```

Repro

```c#
{
	var connectionSettings = new ConnectionSettings(new Uri("http://localhost:9200"))
	.DisableDirectStreaming()
	.PrettyJson()
	.OnRequestCompleted(callDetails =>
	{
		if (callDetails.RequestBodyInBytes != null)
		{
			Console.WriteLine(
				$"{callDetails.HttpMethod} {callDetails.Uri} \n" +
				$"{Encoding.UTF8.GetString(callDetails.RequestBodyInBytes)}");
		}
		else
		{
			Console.WriteLine($"{callDetails.HttpMethod} {callDetails.Uri}");
		}

		Console.WriteLine();

		if (callDetails.ResponseBodyInBytes != null)
		{
			Console.WriteLine($"Status: {callDetails.HttpStatusCode}\n" +
					 $"{Encoding.UTF8.GetString(callDetails.ResponseBodyInBytes)}\n" +
					 $"{new string('-', 30)}\n");
		}
		else
		{
			Console.WriteLine($"Status: {callDetails.HttpStatusCode}\n" +
					 $"{new string('-', 30)}\n");
		}
	});
	//uncomment one of the following lines to crash
	//connectionSettings.DefaultMappingFor<Document>(d => d.IndexName("myindex"));
	//connectionSettings.DefaultIndex("mydefaultindex");

	ElasticClient client = new ElasticClient(connectionSettings);

	BulkAllObservable<Document> bulk = client.BulkAll(GetDocuments(), b => b
		.BufferToBulk((descriptor, list) =>
		{
			foreach (Document doc in list)
			{
				descriptor.Index<Document>(bi => bi
					.Index($"myindex-{doc.Date:yyyy.mm.dd}")
					.Document(doc)
				);
			}
		})
		.Size(2)
	);

	bulk.Subscribe(new BulkAllObserver(
		onNext: r => Console.WriteLine(string.Join(";", r.Items.Select(i => i.Id))),
		onError: e => Console.Error.WriteLine(e),
		onCompleted: () => Console.WriteLine("Done")
	));
	bulk.Wait(TimeSpan.FromSeconds(30), null);
}

private static IEnumerable<Document> GetDocuments()
{
	for (int i = 0; i < 10; i++)
	{
		yield return new Document { Date = DateTime.UtcNow.AddDays(i) };
	}
}

private class Document
{
	public DateTime Date { get; set; }
}
```

russcam:
This commit adds additional XML docs to indicate that Index on `BulkAllRequest<T>` will be inferred from `T`. When a default index has not been mapped for
`T` on `ConnectionSettings`, using `DefaultMappingFor<T>(m => m.IndexName("some_index"))`, an exception will be thrown.

Closes #3951

### DIFF
--- a/src/Nest/Document/Multiple/BulkAll/BulkAllRequest.cs
+++ b/src/Nest/Document/Multiple/BulkAll/BulkAllRequest.cs
@@ -47,7 +47,11 @@ namespace Nest
 		/// </summary>
 		Action<BulkResponseItemBase, T> DroppedDocumentCallback { get; set; }
 
-		///<summary>Default index for items which don't provide one</summary>
+		///<summary>The index to use for items that don't specify one. By default, will be inferred from <typeparamref name="T"/>.
+		/// If no default index has been mapped for <typeparamref name="T" />
+		/// using <see cref="ConnectionSettingsBase{TConnectionSettings}.DefaultMappingFor{TDocument}"/>
+		/// on <see cref="Nest.ConnectionSettings"/>, an exception will be thrown.
+		/// </summary>
 		IndexName Index { get; set; }
 
 		///<summary>The maximum number of bulk operations we want to have in flight at a time</summary>
@@ -85,7 +89,7 @@ namespace Nest
 		/// non-negative value less than or equal to the total number of copies for the shard (number of replicas + 1)
 		/// </summary>
 		int? WaitForActiveShards { get; set; }
-		
+
 		/// <summary>
 		/// Be notified every time a bulk response returns, this includes retries.
 		/// <see cref="IObserver{T}.OnNext"/> is only called for successful batches.
@@ -152,7 +156,7 @@ namespace Nest
 
 		/// <inheritdoc />
 		public int? WaitForActiveShards { get; set; }
-		
+
 		/// <inheritdoc />
 		public Action<BulkResponse> BulkResponseCallback { get; set; }
 	}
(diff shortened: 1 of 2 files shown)


src/Nest/Document/Multiple/BulkAll/BulkAllRequest.cs
```diff
@@ -47,7 +47,11 @@ public interface IBulkAllRequest<T> where T : class
 		/// </summary>
 		Action<BulkResponseItemBase, T> DroppedDocumentCallback { get; set; }
 
-		///<summary>Default index for items which don't provide one</summary>
+		///<summary>The index to use for items that don't specify one. By default, will be inferred from <typeparamref name="T"/>.
```
codebrain:
```suggestion
		///<summary>The index to use for documents that don't specify one. By default, will be inferred from <typeparamref name="T"/>.
```

russcam: I originally wrote documents :) I changed it to items because the `IBulkAllRequest<T>` sends bulk request items which might include documents in the case of index operations, but might not for others like delete operations, although bulking deletes would be an unusual usage of `BulkAll`. The other thought was that _documents don't specify the index_, but maybe we can come up with better wording for this?

codebrain: Ah, yes, I see the problem now :) In that case I think "items" is fine as a cover-all term.

If we really wanted to be pedantic perhaps "documents and bulk operations" - but I think this starts to become confusing too!

russcam: I'll leave as is for the moment. Thanks for looking 👍

russcam:
ported to 6.x and 7.x

PR #4019, opened 2019-08-12
Map node failures.

### LINKED ISSUE #3851: Map failures on NodeStatistics
Map failures on `NodeStatistics` response object.

codebrain:
Fixes #3851

### DIFF
--- a/src/Nest/Cluster/NodesResponseBase.cs
+++ b/src/Nest/Cluster/NodesResponseBase.cs
@@ -1,4 +1,6 @@
-﻿using System.Runtime.Serialization;
+﻿using System.Collections.Generic;
+using System.Runtime.Serialization;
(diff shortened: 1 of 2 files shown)


codebrain:
> LGTM 👍
> 
> Is there a simple integration test that we can perform to assert failures map?

That really means creating an integration test that creates a failure on the node and then asserts this is present. Whilst we could do this, I don't see that it provides much value over a unit test.

The class maps this object: https://github.com/elastic/elasticsearch/blob/a086c665a35af32e60e6d23ab6ca75fcf085b46f/client/rest-high-level/src/main/java/org/elasticsearch/client/NodesResponseHeader.java#L55

I can look to add something to `NodesUsageUnitTests` for this.

PR #4024, opened 2019-08-13
feat(date-range-agg): Add Missing method

fontourafernando:
Missing method was missing in DateRange Aggregation


codebrain:
This looks good @fontourafernando 

@russcam - I have marked as `6.8.2` - will look at merging shortly. I will also port into the `7.2` branch after merge.

PR #4027, opened 2019-08-15
Normalize exception from cancelling an observable coordinated operation

Mpdreamz:
Coordinated request observables when aborted could bubble out an `UnexpectedElasticsearchClientException` or an `OperationCancelledException`.

This normalizes what gets passed to `OnError` namely (`OperationCancelledException`)

This is a bug was never released and something that the early termination in `ForEach` that #4014 fixed highlighted.

### DIFF
--- a/src/Elasticsearch.Net/Transport/Transport.cs
+++ b/src/Elasticsearch.Net/Transport/Transport.cs
@@ -143,6 +143,9 @@ namespace Elasticsearch.Net
 					}
 					catch (Exception killerException)
 					{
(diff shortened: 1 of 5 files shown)


src/Elasticsearch.Net/Transport/Transport.cs
```diff
@@ -143,6 +143,9 @@ public TResponse Request<TResponse>(HttpMethod method, string path, PostData dat
 					}
 					catch (Exception killerException)
 					{
+						if (killerException is OperationCanceledException && cancellationToken.IsCancellationRequested)
```
codebrain: 👍

PR #4030, opened 2019-08-19
Allow ref version for doc generation

russcam:
This PR allows a reference version to be passed when generating documentation. The reference version is the version used in the link to the Elasticsearch reference documentation.

- Improve connection pooling docs
    Improve the connection pooling docs by hiding test assertions from being output in documentation, and linking to sniffing and pinging behaviour docs.
- Regen docs for master


russcam:
ported to 7.x and 6.x

PR #4032, opened 2019-08-19
fix #4029: Use enum item.Name instead of Enum.GetName to get enum value

### LINKED ISSUE #4029: "Key was already exists" on NEST indexing after upgrading from ES v6.8.1
**NEST/Elasticsearch.Net version**: 7.2.0/7.2.0 (and 7.1.0/7.1.0)

**Elasticsearch version**: 7.3.0

I upgraded my elasticsearch server from 6.8.1 to 7.3.0 and after this, the NEST indexing function (`Client.IndexDocument(document)`) throws with `Key was already exists. Key:MultipleChoices`. I can't see any properties in my document corresponding to "MultipleChoices", so I'm not sure where that is coming from.
Furthermore, by changing what index it posts to, I still get the same error (even on a new index), and the (new) index is never created, so this error gets thrown before indexing happens.

If I change the document type from my custom class to just `new`-ing up a `{ Test = "testing" }`, that indexes successfully without throwing an error.

Simply POSTing the document directly at the Elasticsearch API also works fine.

By rewriting my code to use the low-level client I made it work again:

`Client.LowLevel.Index<MyType>(index, documentAsJson);`

I searched a bit and found #3706, but that looks a lot more low-level than what I was doing, but maybe that error propagates up to the NEST index function? It is the case that my custom type was first created/indexed on ES v6.6.0, then used again on v6.8.1 and now in v7.3.0.

My custom class is as follows: 
```cs
[Serializable] // using Elasticsearch.Net;
public class HealthCheckResponse : IElasticsearchResponse
{
    // HealthStatus, HealthCheckProtocol and HttpStatusCode are enums
    public string Host;
    public HealthStatus Status;
    public string Message;
    public string Name;
    public HealthCheckProtocol Protocol;
    public DateTime Timestamp;
    public HttpStatusCode HttpStatusCode;
    public long PingRoundTripTime;
}
```

And the JSON I now send "manually" through the low-level client API is as follows (manually mapping the enums from ints to strings)
```json
{
  "host": "127.0.0.1",
  "status": "Success",
  "message": "Success",
  "name": "localhost",
  "protocol": "Ping",
  "timestamp": "2019-08-16T13:33:21.8036015+02:00",
  "httpStatusCode": 0,
  "pingRoundTripTime": 0
}
```

**Steps to reproduce**:
 1. Use NEST.Client.IndexDocument to index a document against an ES v6.6.0 server
 2. Upgrade the ES server to v7.3.0 (through 6.8.1)
 3. Try to use IndexDocument the same way against the upgraded server, and get an exception thrown.

Do tell if you would like me to debug some more or test some other cases, I'll do my best to help out debugging this :) And feel free to ask for more information if that's relevant.

sklirg:
This is the propsed fix from the original repo of the code: https://github.com/neuecc/Utf8Json/pull/129

Resolves #4029 

I've tested this locally in a .NET Framework project and it works, compared to previously where it'd throw with an exception as mentioned in #4029.

I'm not sure if this should target 7.x or the master branch.

### DIFF
--- a/src/Elasticsearch.Net/Utf8Json/Formatters/EnumFormatter.cs
+++ b/src/Elasticsearch.Net/Utf8Json/Formatters/EnumFormatter.cs
@@ -166,7 +166,7 @@ namespace Elasticsearch.Net.Utf8Json.Formatters
 			foreach (var item in type.GetFields().Where(fi => fi.FieldType == type))
 			{
 				var value = item.GetValue(null);
-				var name = Enum.GetName(type, value);
+				var name = item.Name;
 				var dataMember = item.GetCustomAttributes(typeof(DataMemberAttribute), true)
 					.OfType<DataMemberAttribute>()
 					.FirstOrDefault();


russcam:
Thanks for opening this, @sklirg 👍  Targeting master is best, we can backport to 7.x then.

russcam:
ported to 7.x: 7d974ac5b18022ad4edaadf1ab17950ae0140243

sklirg:
Is there a planned release that includes this fix in the near future? :-) Would love to have this in our testing environment to make sure our upgrade to 7.x works smoothly in production as well.

juteasl:
What's the status on implementing this PR?

PR #4036, opened 2019-08-20
Support StringEnumAttribute on properties

### LINKED ISSUE #4035: Support StringEnumAttribute on a property
The current usage of `StringEnumAttribute` in 7.x allows an `enum` type to be attributed to serialize all members as strings. 

This should be extended to allow an enum property on a type to be attributed with `StringEnumAttribute`, to allow that property to be serialized as a string.

russcam:
This commit supports attributing a an enum property on a type with StringEnumAttribute to serialize it as a string.

Closes #4035


russcam:
ported to 7.x

PR #4039, opened 2019-08-20
fix: Remove redundant word from exception message.

sklirg:
I think the exception message sounds better without the extra "was" in there 😊


russcam:
ported to 7.x. Thanks @sklirg 👍

Issue #4041, opened 2019-08-20
Serialization is locale-dependent for Distance type


rickardp:
While having the NEST code checked out for other reasons, I happened to notice that the Distance fail to serialize properly on my system (sv-SE locale).

I will try to create a failing unit test for it as an example.

russcam:
Thanks for reporting @rickardp. Distance should always serialize with `CultureInfo.InvariantCulture`

PR #4042, opened 2019-08-20
Serialize Distance and Percentiles aggregate using an invariant locale

rickardp:
See issue #4041

Added unit tests to show problem and added a fix using the invariant culture to serialize the type. I only saw the problem with fractional values but there can potentially also be problems with thousand separators in some locales.


src/Nest/CommonOptions/Geo/Distance.cs
```diff
@@ -7,7 +7,7 @@
 namespace Nest
 {
 	[JsonFormatter(typeof(DistanceFormatter))]
-	public class Distance
+	public class Distance : IFormattable
```
russcam: `IFormattable` shouldn't be implemented; the string representation of `Distance` should always use `CultureInfo.InvariantCulture`

src/Tests/Tests/CommonOptions/DateMath/DateMathExpressions.doc.cs
```diff
@@ -232,5 +232,21 @@ public class DateMathExpressions
 			(twelveMonths > fiftyTwoWeeks).Should().BeTrue();
 			(oneYear > fiftyTwoWeeks).Should().BeTrue();
 		}
+		[U]
+		public void DifferentLocales()
```
russcam: Can use `UseCultureAttribute` on test method, and separate out the three different culture tests into separate unit tests. Since these are in a `*.doc.cs` file, these should also be marked with `// hide` to not be emitted in generated documentation e.g.

```c#
// hide
[U]
[UseCulture("sv-SE")]
public void SerializesWithInvariantCulture()
{
   // ..implementation
}
```

src/Tests/Tests/CommonOptions/DistanceUnit/DistanceUnits.doc.cs
```diff
@@ -101,5 +102,36 @@ public void UsingDifferentUnits()
 			*/
 			Expect("45.5nmi").WhenSerializing(new Distance(45.5, Nest.DistanceUnit.NauticalMiles));
 		}
+		[U]
+		public void DifferentLocales()
```
russcam: Can use `UseCultureAttribute` on test method, and separate out into separate unit tests. Since these are in a `*.doc.cs` file, these should also be marked with `// hide` to not be emitted in generated documentation e.g.

```c#
// hide
[U]
[UseCulture("sv-SE")]
public void SerializesWithInvariantCulture()
{
   // ..implementation
}
```

src/Tests/Tests/LocaleUtil.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+using System.Threading;
+
+namespace Tests
+{
+	internal sealed class LocaleUtil : IDisposable
```
russcam: Can remove this and use `UseCultureAttribute`

src/Nest/CommonOptions/Geo/Distance.cs
rickardp: Makes sense if string representation is only used for serialization, which I assume it is.

src/Tests/Tests/CommonOptions/DateMath/DateMathExpressions.doc.cs
rickardp: Cool stuff! Didn't know about that!

rickardp: Now I see, the UseCultureAttribute is copied from xunit samples and available only in Tests.Reproduce. I will move the unit test there unless you think it's a bad idea @russcam

russcam: ah, gotcha. I think moving it to `Tests.Core` project, perhaps under `Xunit` directory might make most sense? That way, both Tests.Reproduce and Tests can use it

rickardp:
When I moved the unit test I found another unit test that was failing if the culture is not en-US (GithubIssue3715.CanDeserializeExtendedStats). Will investigate

russcam:
Feel free to open a separate issue for GithubIssue3715.CanDeserializeExtendedStats. I can take a look at it (tomorrow) :+1:

rickardp:
Created issue #4044 regarding the issue with CanDeserializeExtendedStats. The problem is probably the `double.Parse` calls in `AggregateFormatter` (from just briefly looking at the stacktrace of the failed test).

Might be a good idea to scan the code for `<numeric type>.Parse` without a CultureInfo.InvariantCulture in the serialization code.

rickardp:
I added the fix for issue #4044 as well since I think it was a no-brainer. In AggregateFormatter, I added InvariantCulture to the double.Parse call. I did a full solution search for double.Parse and the only remaining case where InvariantCulture is not used is in StringToDoubleConverter in Utf8Json, and it looks like a fall-back case.

src/Tests/Tests/CommonOptions/DistanceUnit/DistanceUnits.doc.cs
rickardp: I tried to follow the pattern of the other GH issue repro cases where the issue number is part of the unit test file name.

russcam:
ported to 7.x

PR #4043, opened 2019-08-21
Write examples directly to asciidoc

russcam:
This commit replaces the AsciiDoc include in generated examples with the C# source code.
Doing so also allows any examples to include callouts, similar to doc generation e.g.

```c#
var deleteResponse = client.Delete<Tweet>(1, d => d
    .Index("twitter")
    .Routing("kimchy") // <1> Use a routing parameter
);
```

will emit

```asciidoc
[source,csharp]
----
var deleteResponse = client.Delete<Tweet>(1, d => d
    .Index("twitter")
    .Routing("kimchy") //<1>
);
----
<1>  Use a routing parameter
```

### DIFF
--- a/examples/Docs/DeletePage/47b5ff897f26e9c943cee5c06034181d.asciidoc
+++ b/examples/Docs/DeletePage/47b5ff897f26e9c943cee5c06034181d.asciidoc
@@ -1,4 +1,7 @@
 [source, csharp]
 ----
-include::../../../src/Examples/Examples/Docs/DeletePage.cs[tag=47b5ff897f26e9c943cee5c06034181d,indent=0]
+var deleteResponse = client.Delete<Tweet>(1, d => d
+    .Index("twitter")
+    .Routing("kimchy")
+);
 ----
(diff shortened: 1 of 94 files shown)


Mpdreamz:
It might be good to add a notice/comment to the files now that these are generated and should not be edited directly.

russcam:
> It might be good to add a notice/comment to the files now that these are generated and should not be edited directly.

++

@nik9000, If these examples had a multine asciidoc comment in them, would that be an issue for the docs example build process, similar to what we do in the current .NET docs e.g.

```asciidoc
////
IMPORTANT NOTE
==============
This file has been generated from <path to file>. 
If you wish to submit a PR for any spelling mistakes, typos or grammatical errors for this file,
please modify the original csharp file found at the link and submit the PR with that change. Thanks!
////
```

russcam:
@Mpdreamz I've added multiline comments that point back to the original file with line numbers, on GitHub. Would you mind taking another look, please?

Mpdreamz:
Awesome, lgtm :+1:

Issue #4033, opened 2019-08-19
Support for search-as-you-type datatype


rickardp:
**NEST/Elasticsearch.Net version**:
7.2.0, master
**Elasticsearch version**:
7.2.0
**Description of the problem including expected versus actual behavior**:
The `search_as_you_type` datatype is not available in NEST.

**Steps to reproduce**:
 1. Try to use `search_as_you_type` data type or `bool_prefix` multi match type

The examples in https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-as-you-type.html should work with NEST.

PR #4046, opened 2019-08-22
Support for search_as_you_type field type and bool_prefix query type

rickardp:
The search_as_you_type field type was introduced in ElasticSearch 7.2. We missed this functionality in NEST so we went ahead and implemented it.

Also added the bool_prefix query type that we needed to implement our search completions.

Reported in issue #4033

PR #4048, opened 2019-08-23
Support for search_as_you_type field type and bool_prefix query type 

russcam:
Supersedes: #4046 

Closes: #4033

### DIFF
--- a/src/Nest/Mapping/Types/Core/SearchAsYouType/SearchAsYouTypeProperty.cs
+++ b/src/Nest/Mapping/Types/Core/SearchAsYouType/SearchAsYouTypeProperty.cs
@@ -1,39 +1,137 @@
-using System;
 using System.Diagnostics;
 using System.Runtime.Serialization;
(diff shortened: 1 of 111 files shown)


src/Nest/Mapping/Types/Core/SearchAsYouType/SearchAsYouTypeProperty.cs
```diff
@@ -0,0 +1,137 @@
+using System.Diagnostics;
+using System.Runtime.Serialization;
+using Elasticsearch.Net.Utf8Json;
+
+namespace Nest
+{
+	/// <summary>
+	/// A text-like field that is optimized to provide out-of-the-box support for queries that serve an as-you-type completion use case.
```
codebrain:
```suggestion
	/// A text-like field that is optimized to provide out-of-the-box support for the "search as you type" completion use case.
```

rickardp:
Great stuff! Just confirming, the 7.3.0 label means this is backported to 7.x, right?

Issue #4001, opened 2019-08-07
[meta] 7.3 Release


Mpdreamz:
#### Data Frame
~- [x ] Removing format support in date_histogram group_by http://github.com/elastic/elasticsearch/pull/43659~ Data Frame APIs not implemented

#### Features/Java High Level REST Client
~- [x] Deprecate native code info in xpack info api http://github.com/elastic/elasticsearch/pull/43297~ N/A

#### Mapping
~- [x] Deprecate support for chained multi-fields. http://github.com/elastic/elasticsearch/pull/41926 (issue: http://github.com/elastic/elasticsearch/issues/41267)~ no change needed

#### Search
- [x] Deprecate CommonTermsQuery and cutoff_frequency http://github.com/elastic/elasticsearch/pull/42619 (issue: http://github.com/elastic/elasticsearch/issues/37096)

#### Aggregations
- [x] Adds a minimum interval to `auto_date_histogram`. http://github.com/elastic/elasticsearch/pull/42814 (issue: http://github.com/elastic/elasticsearch/issues/41757)
- [x] Add RareTerms aggregation http://github.com/elastic/elasticsearch/pull/35718 (issue: http://github.com/elastic/elasticsearch/issues/20586)

#### Data Frame
~- [x] Add sync api http://github.com/elastic/elasticsearch/pull/41800~ Data Frame APIs not implemented

#### Machine Learning
~- [x] Machine learning data frame analytics http://github.com/elastic/elasticsearch/pull/43544~ Data Frame APIs not implemented 

#### Mapping
- [x] Add support for 'flattened object' fields. http://github.com/elastic/elasticsearch/pull/42541 (issues: http://github.com/elastic/elasticsearch/issues/25312, http://github.com/elastic/elasticsearch/issues/33003)

#### Ranking
~- [x] Move dense_vector and sparse_vector to module http://github.com/elastic/elasticsearch/pull/43280~ N/A

#### Search
- [x] Wildcard intervals http://github.com/elastic/elasticsearch/pull/43691 (issue: http://github.com/elastic/elasticsearch/issues/43198)
- [x] Add prefix intervals source http://github.com/elastic/elasticsearch/pull/43635 (issue: http://github.com/elastic/elasticsearch/issues/43198)
[...]

PR #4053, opened 2019-08-28
Add minimum_interval to AutoDateHistogram aggregation

russcam:
Relates: #4001

This commit adds the minimum_interval property to auto date histogram aggregation. A new enum needs to be introduced for this as DateInterval contains values that are invalid for the minimum_interval.


russcam:
ported to master

PR #4060, opened 2019-08-30
Add wildcard and prefix interval rules

russcam:
Relates: #4001

This commit adds wildcard and prefix rules support to the Intervals query.
Neither wildcard or prefix rules support filters, so a new IIntervalsNoFilter interface
is introduced for these rules. Added TODO to consolidate rules in 8.0

### DIFF
--- a/src/Nest/QueryDsl/FullText/Intervals/IntervalsPrefix.cs
+++ b/src/Nest/QueryDsl/FullText/Intervals/IntervalsPrefix.cs
@@ -38,14 +38,12 @@ namespace Nest
 	public class IntervalsPrefix : IntervalsNoFilterBase, IIntervalsPrefix
 	{
 		/// <inheritdoc />
-		[DataMember(Name = "analyzer")]
 		public string Analyzer { get; set; }
 
 		/// <inheritdoc />
 		public string Prefix { get; set; }
 
 		/// <inheritdoc />
-		[DataMember(Name = "use_field")]
 		public Field UseField { get; set; }
 
 		internal override void WrapInContainer(IIntervalsContainer container) => container.Prefix = this;
--- a/src/Nest/QueryDsl/FullText/Intervals/IntervalsPrefix.cs
+++ b/src/Nest/QueryDsl/FullText/Intervals/IntervalsPrefix.cs
@@ -1,0 +1,73 @@
+using System;
+using System.Linq.Expressions;
+using System.Runtime.Serialization;
+
+namespace Nest
+{
+	/// <summary>
+	/// matches terms that start with a specified set of characters. This prefix can expand to match at most 128 terms.
+	/// If the prefix matches more than 128 terms, Elasticsearch returns an error.
+	/// You can use the index-prefixes option in the field mapping to avoid this limit.
+	/// <para />
+	/// Available in Elasticsearch 7.3.0+
+	/// </summary>
+	[ReadAs(typeof(IntervalsPrefix))]
+	public interface IIntervalsPrefix : IIntervalsNoFilter
+	{
+		/// <summary>
+		/// Analyzer used to normalize the prefix. Defaults to the top-level field's analyzer.
+		/// </summary>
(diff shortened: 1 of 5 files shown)


src/Nest/QueryDsl/FullText/Intervals/IntervalsPrefix.cs
```diff
@@ -0,0 +1,73 @@
+using System;
+using System.Linq.Expressions;
+using System.Runtime.Serialization;
+
+namespace Nest
+{
+	/// <summary>
+	/// matches terms that start with a specified set of characters. This prefix can expand to match at most 128 terms.
```
codebrain:
```suggestion
	/// Matches terms that start with a specified set of characters. This prefix can expand to match at most 128 terms.
```

src/Nest/QueryDsl/FullText/Intervals/IntervalsPrefix.cs
```diff
@@ -0,0 +1,73 @@
...
+		string Analyzer { get; set; }
+
+		/// <summary>
+		/// Beginning characters of terms you wish to find in the top-level field
+		/// </summary>
+		[DataMember(Name = "prefix")]
+		string Prefix { get; set; }
+
+		/// <summary>
+		/// If specified, then match intervals from this field rather than the top-level <field>.
```
codebrain:
```suggestion
		/// If specified, then match intervals from this field rather than the top-level field.
```

src/Nest/QueryDsl/FullText/Intervals/IntervalsPrefix.cs
```diff
@@ -0,0 +1,73 @@
...
+
+	/// <inheritdoc cref="IIntervalsPrefix" />
+	public class IntervalsPrefix : IntervalsNoFilterBase, IIntervalsPrefix
+	{
+		/// <inheritdoc />
+		[DataMember(Name = "analyzer")]
+		public string Analyzer { get; set; }
+
+		/// <inheritdoc />
+		public string Prefix { get; set; }
```
codebrain: Missing `DataMember`?

src/Nest/QueryDsl/FullText/Intervals/IntervalsWildcard.cs
```diff
@@ -0,0 +1,73 @@
+using System;
+using System.Linq.Expressions;
+using System.Runtime.Serialization;
+
+namespace Nest
+{
+	/// <summary>
+	/// matches terms using a wildcard pattern. This pattern can expand to match at most 128 terms.
```
codebrain:
```suggestion
	/// Matches terms using a wildcard pattern. This pattern can expand to match at most 128 terms.
```

src/Nest/QueryDsl/FullText/Intervals/IntervalsWildcard.cs
```diff
@@ -0,0 +1,73 @@
...
+		/// Wildcard pattern used to find matching terms. Supports two wildcard operators:
+		/// <para />?, which matches any single character
+		/// <para />*, which can match zero or more characters, including an empty one
+		/// <para />Warning: Avoid beginning patterns with * or ?. This can increase the iterations needed to find matching terms and slow search performance.
+		/// </summary>
+		[DataMember(Name = "pattern")]
+		string Pattern { get; set; }
+
+		/// <summary>
+		/// If specified, then match intervals from this field rather than the top-level <field>.
```
codebrain:
```suggestion
		/// If specified, then match intervals from this field rather than the top-level field.
```

src/Nest/QueryDsl/FullText/Intervals/IntervalsPrefix.cs
russcam: Thanks for catching this. The concrete impl doesn't need the attributes. Will update

russcam:
ported to master

PR #4064, opened 2019-09-02
Add voting_only to ClusterNodeStats and XPackInfo

russcam:
Relates: #4001

This commit adds voting_only to ClusterNodeStats and XPackUsageResponse.
VotingOnly is supported only in the default distribution


russcam:
pushed a change to skip rare terms for < 7.3.0, as integration tests were failing. Will rebase this change into 7.x once tests pass

PR #4068, opened 2019-09-03
Allow execution of each Watcher action in array

russcam:
Relates: #4001

This commit introduces Foreach on Watcher actions, to allow execution of the Watcher
action on each element in an array specified by the path assigned to foreach.

Add condition to Watcher actions and ensure condition and transform are serialized.

Add integration test for Action foreach, transform and condition.


russcam:
ported to master

PR #4071, opened 2019-09-03
Add max_docs to ReindexOnServer API

russcam:
- Add `MaximumDocuments` to ReindexOnServer API and deprecate `Size`
- Add `MaximumDocuments` and support for SlicedScroll to UpdateByQuery and DeleteByQuery


russcam:
ported to master

PR #4073, opened 2019-09-04
Add support for SourceOnly snapshot repository

russcam:
This commit adds support for SourceOnly repository.
Introduce a ResetOffset() method on JsonReader to reset the offset back to the initial offset
Tidy up XML docs on repositories


russcam:
ported to master

PR #4075, opened 2019-09-05
Skip max_docs query string params

russcam:
This commit skips max_docs as query string params on DeleteByQuery,UpdateByQuery and ReindexOnServer.

The parameter can also be defined in the request body, so prefer this.


russcam:
ported to master

Issue #4078, opened 2019-09-07
BulkIndexDescriptor IfPrimaryTerm sets IfSequenceNumber


takridge:
**NEST/Elasticsearch.Net version**: 7.3

**Elasticsearch version**: 7.3

**Description of the problem including expected versus actual behavior**: When utilizing the .IndexMany() of .Bulk() in the ElasticClient, calling the .IfPrimaryTerm() does not set the IfPrimaryTerm ("if_primary_term") DataMember, but rather the IfSequenceNumber ("if_seq_no") DataMember.

**Steps to reproduce**:
1. Already have index created
2. Instantiate ElasticClient connected to existing cluster
3. Call `.Bulk(x => x.IndexMany(documents, (d, doc) => d.IndexName("index_name").IfSequenceNumber(doc.SequenceNumber).IfPrimaryTerm(doc.PrimaryTerm)))`
4. Receive error from server: `Validation Failed: 1: ifSeqNo is set, but primary term is [0];...`

**Provide `DebugInformation`**:
```
...
{"index":{"_id":"anExampleId","_index":"index_name","if_seq_no":1335}}
...
```

**Relevant Nest Code**:
master: src\Nest\Document\Multiple\Bulk\BulkOperation\BulkIndex.cs:86

`public BulkIndexDescriptor<T> IfPrimaryTerm(long? primaryTerm) => Assign(primaryTerm, (a, v) => a.IfSequenceNumber = v);`

takridge:
This also affects `BulkUpdateDescriptor` in src\Nest\Document\Multiple\Bulk\BulkOperation\BulkUpdate.cs:265-266

```
        public BulkUpdateDescriptor<TDocument, TPartialDocument> IfPrimaryTerm(long? primaryTerm) =>
            Assign(primaryTerm, (a, v) => a.IfSequenceNumber = v);
```

should be

```
        public BulkUpdateDescriptor<TDocument, TPartialDocument> IfPrimaryTerm(long? primaryTerm) =>
            Assign(primaryTerm, (a, v) => a.IfPrimaryTerm = v);
```

russcam:
Closing this as it's fixed in #4080 which has now been merged. It'll be in the next release of 6.x and 7.x

PR #4080, opened 2019-09-08
Fix fluent serialization of if_primary_term (#4078)

andrx:
Fixes fluent serialization of if_primary_term for BulkIndexDescriptor and BulkUpdateDescriptor


andrx:
looks like timeout exceptions during the integration tests

russcam:
I've ported the change from #4087 to 7.x branch, so going to close this PR as it's the same change. Thanks again for opening @andrx 👍

PR #4082, opened 2019-09-09
Fix Stats aggregate sum change type to nullable

JefStat:
Encountered a bug on production due to this response data. Note the production cluster is running Elasticsearch 6.8

```
{
          "key" : "Morning 0",
          "doc_count" : 17,
          "subStatsAgg" : {
            "meta" : {
              "aggType" : "Sum"
            },
            "count" : 0,
            "min" : null,
            "max" : null,
            "avg" : null,
            "sum" : null
          }
        }
```

### DIFF
--- a/src/Nest/Aggregations/AggregateFormatter.cs
+++ b/src/Nest/Aggregations/AggregateFormatter.cs
@@ -419,7 +419,7 @@ namespace Nest
 			reader.ReadNext(); // ,
 			reader.ReadNext(); // sum
 			reader.ReadNext(); // :
-			var sum = reader.ReadDouble();
+			var sum = reader.ReadNullableDouble();
 
 			var statsMetric = new StatsAggregate
 			{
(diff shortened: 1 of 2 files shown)


russcam:
@JefStat which version of the client are you using against Elasticsearch 6.8.0?

Stats aggregate was changed in 7.0.0-beta1 of Elasticsearch to return `sum` of `0` when the `count` is `0` in https://github.com/elastic/elasticsearch/pull/27193, so the client was updated in https://github.com/elastic/elasticsearch-net/commit/4f64167a246ad2f2d9965720bc8df73838086219 to reflect this in master and 7.x. The 6.x client still treats it as `double?`: https://github.com/elastic/elasticsearch-net/blob/f00e0e886b51cf1dd111a7afb6dc40c5faf9bf55/src/Nest/Aggregations/AggregateJsonConverter.cs#L323

Are you using a 7.x client with 6.8, by chance?

JefStat:
> Are you using a 7.x client with 6.8, by chance?

Yes, we are. This is a temporary state before doing a rolling upgrade on Elasticsearch. I tested running the NEST 6.8 client on 7.3 Cluster and it did not work on several issues.

JefStat:
I see a blog post has been released to partly address our issue https://www.elastic.co/blog/nest-and-elasticsearch-net-upgrading-your-codebase I can extrapolate from there

Issue #4085, opened 2019-09-10
PatternReplaceCharFilter is missing a Flags property


awelburn:
**NEST/Elasticsearch.Net version**: 6.x, (but affects "all"?)

**Elasticsearch version**: 6.x

**Describe the feature**:
To match the allowed parameters detailed here: https://www.elastic.co/guide/en/elasticsearch/reference/master/analysis-pattern-replace-charfilter.html

Add a Flags property to:
IPatternReplaceCharFilter
PatternReplaceCharFilter
PatternReplaceCharFilterDescriptor

PR #4086, opened 2019-09-10
Add Flags property, use in tests

awelburn:
PR for issue #4085


russcam:
ported to 7.x and 6.x

PR #4087, opened 2019-09-11
Fix fluent serialization of if_primary_term (#4078)

andrx:
Fixes fluent serialization of if_primary_term for BulkIndexDescriptor and BulkUpdateDescriptor


russcam:
ported to 7.x and 6.x

Issue #4089, opened 2019-09-17
GeoSuggestContextDescriptor's Precision-method accepts array instead of singular value


billbogaiv:
**NEST/Elasticsearch.Net version**: 6

**Elasticsearch version**: 6

**Description of the problem including expected versus actual behavior**:

When applying `Mapping` within `client.CreateIndex`, the `GeoLocation` `Completion`-descriptor takes a collection of `Precision`-values. However, ES only accepts a singular value. Even passing one-value to `.Precision("25mi")` results in the construction of an array. This eventually leads to an ES syntax-error when creating the index.

**Steps to reproduce**:

```csharp
public class PlaceSearchDocument
{
    [GeoPoint]
    public Location Location { get; set; }
}

client.CreateIndex("places", (CreateIndexDescriptor createDescriptor) =>
    createDescriptor.Mappings(mappingDescriptor =>
        mappingDescriptor.Map<PlaceSearchDocument>(map => map
            .AllField(af => af.Enabled(false))
            .AutoMap()
            .Properties(property => property
                .Completion(completion => completion.Name("LocationSuggest").Contexts(context => context
                    .GeoLocation(geolocation => geolocation
                        .Name("Location")
                        .Path("Location")
                        .Precision("25mi"))
                ))))));
```

**Provide `DebugInformation` (if relevant)**:

```json
PUT places
{
  "mappings": {
    "place": {
      "_all": {
        "enabled": false
      },
      "properties": {
        "Location": {
          "type": "geo_point"
        },
        "LocationSuggest": {
          "type": "completion",
          "contexts": [
            {
              "type": "geo",
              "name": "Location",
              "path": "Location",
              "precision": ["25mi"]
            }
          ]
        }
      }
    }
  }
}
```

**Expected value**

```json
PUT places
{
  "mappings": {
    "place": {
      "_all": {
        "enabled": false
      },
      "properties": {
        "Location": {
          "type": "geo_point"
        },
        "LocationSuggest": {
          "type": "completion",
          "contexts": [
            {
              "type": "geo",
              "name": "Location",
              "path": "Location",
              "precision": "25mi"
            }
          ]
        }
      }
    }
  }
}
```

Ref. https://github.com/elastic/elasticsearch-net/blob/c646b98c5e5278ad122cd686fa27c45cbb18b919/src/Nest/Search/Suggesters/ContextSuggester/GeoSuggestContext.cs#L35

khalidabuhakmeh:
[...]

PR #4091, opened 2019-09-18
Precision on GeoSuggestContext Is A Single Value

khalidabuhakmeh:
This should only be one value, and not an array.
This is based on documentation from v6+ and v7+.

 #4089


khalidabuhakmeh:
I don't think you folks have enough build servers :P

russcam:
Thanks for the PR @khalidabuhakmeh 👍

russcam:
opening a backport for 7.x shortly

russcam:
I've opened https://github.com/elastic/elasticsearch-net/pull/4106 to address 7.x

Issue #4093, opened 2019-09-19
Terms include for numeric values


simonness:
**Describe the feature**:  

I'd like to include numeric terms in a terms aggregation e.g.

```
{
  "size": 0,
  "aggs": {
    "numeric_agg": {
      "terms": {
        "field": "my_field",
        "include": [1,2,3]
      }
    }
  }
}
```
However it seems [TermsInclude](https://github.com/elastic/elasticsearch-net/blob/c646b98c5e5278ad122cd686fa27c45cbb18b919/src/Nest/Aggregations/Bucket/Terms/TermsInclude.cs#L25) only accepts an array of strings.  Not sure if this is possible in any other way?

simonness:
After reading

> The reasoning is that numeric types in Elasticsearch are optimized for range queries and a bit slower at exact queries than keyword fields.

https://discuss.elastic.co/t/matching-integers-in-an-array/96534/2

Closing and will work around this by re-indexing documents to use strings rather than ints.

PR #4096, opened 2019-09-24
Support objects in terms include / exclude.

simonness:
See issue #4093.

It wasn't clear from the contributing guide how to handle breaking changes, so initially just raising this PR against master to start the discussion.


simonness:
After reading

> The reasoning is that numeric types in Elasticsearch are optimized for range queries and a bit slower at exact queries than keyword fields.

https://discuss.elastic.co/t/matching-integers-in-an-array/96534/2

Closing and will work around this by re-indexing documents to use strings rather than ints.

PR #4097, opened 2019-09-24
Move to .NET 3.0 and C# 8.0

Mpdreamz:
Initial move to `.NET 3.0` and `C# 8.0`


Mpdreamz:
Integration failures do not seem related to .NET core changes.

PR #4101, opened 2019-09-26
Move to .NET 3.0 and C# 8.0 (#4097)

Mpdreamz:
* initial move to .NET Core 3.0

* bump to .net sdk 3.0.100 and update our build files to install it for now while the images catch up

* move scripts to 3.0 since we call dotnet run on them, exclude new types from codestandards test that all types live in Elasticsearch.Net/NEST namespace

* build tooling moved from 2.2 to 3.0

* move all netcoreapp over to 3.0

* move to range on RegexCollection no longer accepts Last()

* update our own dependencies

* make 2.2 available for canary so that rewrite using 2.2 tools still works

* fix yaml indentation

* disable ingest-attachemnt for now, installl has stderr on 8.0.0 which abstractions does not yet handle

* move rewriter and differ to local dotnet tools

* no longer need to install 2.2 sdk for differ

(cherry picked from commit 9bbffadb32015344b8dc95c026f8cd582f8f4d5a)


Mpdreamz:
Two integration test failures one is due to the seed opened #4104 to fix that.

The other is a `ProcessClusterEventTimeoutException` which is unreproducable and not related to these changes.

`build.bat seed:71011 integrate 7.1.0 "writable" "getrepository"` `

Mpdreamz:
Since the doc generation PR is now pending as per #4113 moving ahead and merging this in.

PR #4107, opened 2019-09-29
Remove Neighbors and use Union<Distance,int> for Precision

russcam:
Relates: #4091

This commit removes the Neighbors property from Completion field data type mapping and updates
the Precision type to a `Union<Distance,int>` to match the context suggest query inputs.

Does not require backporting to any branches.


Mpdreamz:
LGTM but has one test failure:

```
Tests.CodeStandards.Descriptors.DescriptorMethodsTakingSingleValueTypeShouldBeNullable [FAIL]
  X Tests.CodeStandards.Descriptors.DescriptorMethodsTakingSingleValueTypeShouldBeNullable [161ms]
  Error Message:
   Expected collection to be empty, but found {"precision on method Precision of Nest.GeoSuggestContextDescriptor`1 is not nullable"}.
  Stack Trace:
     at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message) 
   at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message) 
   at FluentAssertions.Execution.DefaultAssertionStrategy.HandleFailure(String message) 
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc) 

````

PR #4108, opened 2019-09-30
Honour PropertyNameAttribute Ignore property in mappings

### LINKED ISSUE #4105: [PropertyName("name", Ignore=true)] ignore not honoured.
Relates: https://stackoverflow.com/questions/58157898/c-sharp-elasticsearch-nest-v7-x-ignore-true-not-working-on-fields

As part of the refactoring of serialization in 7.x, the `Ignore` property introduced on `PropertyNameAttribute` is not honoured. This should be fixed. This property was introduced as part of `IJsonProperty` that unifies property serialization mapping across attributes and types.

russcam:
Fixes #4105


russcam:
failing tests don't _appear_ to be related to this change.

PR #4109, opened 2019-09-30
Fix Extended stats deserialization

russcam:
Supersedes: #4103 

Adds a unit test to assert that the change to `AggregateJsonConverter` for ExtendedStats aggregation
can deserialize the provided response.

### DIFF
--- a/src/Nest/Aggregations/AggregateJsonConverter.cs
+++ b/src/Nest/Aggregations/AggregateJsonConverter.cs
@@ -311,18 +311,26 @@ namespace Nest
 			if (reader.TokenType == JsonToken.EndObject) return new GeoCentroidAggregate { Count = count };
 
 			reader.Read();
-			var min = reader.Value as double?;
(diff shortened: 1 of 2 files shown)


benbenwilde:
@russcam I'm still having issues with a datetime extended stats aggregation dropping subsequent aggregations. It looks like we need another `reader.Read()` here: https://github.com/elastic/elasticsearch-net/pull/4109/files#diff-0081650de00611fa377141b59e107f4fL406

I believe once the aggregate is parsed, the reader is supposed to end on the end token of the aggregation object. However, it looks like it's ending on the end token of the `std_deviation_bounds_as_string` instead.

russcam:
Thanks @benbenwilde, that does sound like the case if subsequent aggs are not parsed and returned in the search response instance . I'll take a look at it today

russcam:
@benbenwilde I've opened https://github.com/elastic/elasticsearch-net/pull/4118 to address

PR #4112, opened 2019-09-30
Fix initial randomization of nodes

### LINKED ISSUE #4055: Randomize feature of StaticConnectionPool does nothing
**NEST/Elasticsearch.Net version**: 6.0.2 and 7.2.1 (presumably others)

**Elasticsearch version**: irrelevant, but 6.6.1

See: https://github.com/elastic/elasticsearch-net/blob/master/src/Elasticsearch.Net/ConnectionPool/StaticConnectionPool.cs
(current latest revision is `c330114`)

In theory passing 'true' for the randomize parameter (true is also the default for this parameter) will cause the order of the supplied nodes to be randomized.
This should happen in the function SortNodes() which (if no node scorer is available) will check the Randomize property and, if true, will use OrderBy(Random.Next()) to return the nodes in a random order.

In practice, because of the way the constructors are nested, SortNodes is called from the protected constructor BEFORE the Randomize property is initialized by the public constructor - so when read in SortNodes the Randomize property has not yet been initialized and will always have its default value of false.

**Steps to reproduce**:
```
var uris = new[]
{
	new Uri("https://10.0.0.1:9200/"),
	new Uri("https://10.0.0.2:9200/"),
	new Uri("https://10.0.0.3:9200/")
};

var pool = new StaticConnectionPool(uris);
var nodeView = pool.CreateView();

Debug.Assert(nodeView.First().Uri.ToString() == "https://10.0.0.1:9200/");
```
**Expected:** The assertion should fail most of the time.
**Actual:** The assertion never fails.

russcam:
This commit fixes the initial randomization of nodes in a StaticConnectionPool. Instead of calling the protected ctor,
the initialization logic has been pulled out into a separate method that is now called from the ctor that accepts a
randomize parameter.

Fixes #4055

### DIFF
--- a/src/Tests/Tests/ClientConcepts/ConnectionPooling/BuildingBlocks/ConnectionPooling.doc.cs
+++ b/src/Tests/Tests/ClientConcepts/ConnectionPooling/BuildingBlocks/ConnectionPooling.doc.cs
@@ -214,6 +214,30 @@ namespace Tests.ClientConcepts.ConnectionPooling.BuildingBlocks
 			}
 		}
 
+		// hide
+		[U] public void RandomizedInitialNodes()
+		{
+			StaticConnectionPool CreatStaticConnectionPool()
+			{
+				var uris = new[]
+				{
+					new Uri("https://10.0.0.1:9200/"),
+					new Uri("https://10.0.0.2:9200/"),
+					new Uri("https://10.0.0.3:9200/")
+				};
+
+				var staticConnectionPool = new StaticConnectionPool(uris);
+				return staticConnectionPool;
+			}
+
+			// assertion works on the probability of seeing a Uri other than https://10.0.0.1:9200/
+			// as the first value over 50 runs, when randomized.
+			Enumerable.Repeat(CreatStaticConnectionPool().CreateView().First().Uri.ToString(), 50)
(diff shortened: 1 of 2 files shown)


src/Tests/Tests/ClientConcepts/ConnectionPooling/BuildingBlocks/ConnectionPooling.doc.cs
```diff
@@ -214,6 +214,30 @@ public class ConnectionPooling
...
+					new Uri("https://10.0.0.3:9200/")
+				};
+
+				var staticConnectionPool = new StaticConnectionPool(uris);
+				return staticConnectionPool;
+			}
+
+			// assertion works on the probability of seeing a Uri other than https://10.0.0.1:9200/
+			// as the first value over 50 runs, when randomized.
+			Enumerable.Repeat(CreatStaticConnectionPool().CreateView().First().Uri.ToString(), 50)
```
Mpdreamz: I am looking to seeing this randomly fail :smile_cat:

russcam: 😃

russcam:
ported to 6.x

PR #4113, opened 2019-09-30
Doc Generation move to .NET core 3.0

Mpdreamz:
This fixes the doc generation no longer working under .NET Core 3.0 since we moved (see: #4097)

This moves us to the latest [Buildalyzer](https://github.com/daveaglick/Buildalyzer) which seems to work for the most part :) Sadly this version lost [our project filter contribution](https://github.com/daveaglick/Buildalyzer/pull/35) in a recent refactor. I might go back this week and submit a new PR but in the interest of moving along we now have a dedicated `sln` file under `src/CodeGeneration` that only references the relevant projects for the doc generation. cc @daveaglick

Touched up the folder locating routing by pinning `global.json`.

Rather then using branch or global json version we now document the intended current doc version in `global.json` separately. This allows branches to gen docs without the noise of all the headers changing.

This also now allows `gendocs` to be specified on non windows OS's, for some reason I had to duplicate `LangVersion` in the relevant csproj files directly. This was not needed on Windows but can't be bothered diving in further to learn why.

We still need to call `dotnet codegeneration.dll` rather than `dotnet run` since in the latter case it starts complaining about certain MEF 2.0 types not being available. Again too intricate to investigate and understand why.


daveaglick:
😬 I opened an issue at https://github.com/daveaglick/Buildalyzer/issues/118 to add the project filter back in. I've got some work to do in Buildalyzer so I'll add this to the list. Does the approach in that issue seem correct?

src/CodeGeneration/DocGenerator/Program.cs
```diff
@@ -4,37 +4,34 @@
...
+
+
+			var r = root.FullName;
+			var globalJson = Path.Combine(r, "global.json");
+			InputDirPath = Path.Combine(r, "src");
+			OutputDirPath = Path.Combine(r, "docs");
+			BuildOutputPath = Path.Combine(r, "build", "output");
 
-			var globalJsonVersion = string.Join(".", Regex.Matches(File.ReadAllText(globalJson), "\"version\": \"(.*)\"")
+			var globalJsonVersion = string.Join(".", Regex.Matches(File.ReadAllText(globalJson), "\"doc_current\": \"(.*)\"")
```
russcam: `Json.Net` to parse global.json?

russcam:
Pulling this in

PR #4118, opened 2019-10-03
Read closing brace of extended stats aggregation

russcam:
Relates: #4109

This commit fixes an issue with extended stats aggregation parsing whereby the closing brace of std_deviation_bounds_as_string
is not read when skipping over the object, leading to the reader exiting at this point, resulting in any subsequent aggregations
from not being read from the JSON response and added to SearchResponse.Aggregations.


src/Tests/Tests.Reproduce/GitHubIssue4103.cs
```diff
@@ -66,6 +69,8 @@ public void CanDeserializeExtendedStats()
 
 			var response = client.Search<object>(s => s.AllIndices());
 
+			response.Aggregations.Count.Should().Be(2);
```
Mpdreamz: Assert `sum` value?

PR #4120, opened 2019-10-06
BulkDescriptor DeleteMany usable with IEnumerable<string> without specifying type.

### LINKED ISSUE #3500: BulkDescriptor.DeleteMany requires <T> while not required for overload IEnumerable<string> ids
6.3.1

**Elasticsearch version**:
6.5.0

There are multiple overloads for `BulkDescriptor.DeleteMany` which can be used in for example the `BulkAsync` method:
- `DeleteMany<T>(IEnumerable<T> @objects, Func<BulkDeleteDescriptor<T>, T, IBulkDeleteOperation<T>> bulkDeleteSelector = null)` ([source](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Document/Multiple/Bulk/BulkRequest.cs#L61))
- `DeleteMany<T>(IEnumerable<string> ids, Func<BulkDeleteDescriptor<T>, string, IBulkDeleteOperation<T>> bulkDeleteSelector = null)` ([source](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Document/Multiple/Bulk/BulkRequest.cs#L71))

Note that both have a `bulkDeleteSelector` parameter which is optional.
Now I want to use the 2nd `IEnumerable<string> ids` overload without the optional `bulkDeleteSelector` parameter, for example:

```csharp
IEnumerable<string> ids = new List<string> { "1", "2" };
var bulkResponse = await client.BulkAsync(descriptor => descriptor
    .Index("my-index")
    .DeleteMany(ids)
);
```
But because I didn't provide a `<T>` C# selects the `IEnumerable<T> @objects` overload instead, causing the operation to fail.
Because I don't need the optional `bulkDeleteSelector` parameter (and it's not returned either) I can workaround this issue by using any `<T>` like this:

```csharp
var bulkResponse = await client.BulkAsync(descriptor => descriptor
    .Index("my-index")
    .DeleteMany<object>(ids)
);
```

**Steps to reproduce**:
 1. Call `BulkAsync` with `IEnumerable<string> ids` as input, but without a `<T>`
 2. Bulk call to ES fails

I think the solution could be quite simple: add an extra overload without the `bulkDeleteSelector` overload and remove the default `= null` for the original overload.

antja0:
Fixes #3500 

My first PR here, hopefully you find this useful 😃

I think good solution here is to just simply check for string.

With other `Nest.Id` convertable types like `long` or `Guid` the correct overload will be called automatically because T cannot be a [value type](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/value-types) (`where T : class`). With string we need to do an exception because it's a [reference type](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/reference-types) like classes etc.

Included a very simple test.

### DIFF
--- a/src/Nest/Document/Multiple/Bulk/BulkRequest.cs
+++ b/src/Nest/Document/Multiple/Bulk/BulkRequest.cs
@@ -57,19 +57,9 @@ namespace Nest
 			Func<BulkDeleteDescriptor<T>, T, IBulkDeleteOperation<T>> bulkDeleteSelector = null
 		)
 			where T : class =>
-			AddOperations(@objects, bulkDeleteSelector, o => new BulkDeleteDescriptor<T>().Document(o));
-
-		/// <summary>
-		/// DeleteMany, convenience method to delete many objects at once.
-		/// </summary>
-		/// <param name="ids">Enumerable of string ids to delete</param>
-		/// <param name="bulkDeleteSelector">A func called on each ids to describe the individual delete operation</param>
-		public BulkDescriptor DeleteMany<T>(
-			IEnumerable<string> ids,
-			Func<BulkDeleteDescriptor<T>, string, IBulkDeleteOperation<T>> bulkDeleteSelector = null
-		)
-			where T : class =>
-			AddOperations(ids, bulkDeleteSelector, id => new BulkDeleteDescriptor<T>().Id(id));
+			AddOperations(@objects, bulkDeleteSelector, o => o is string str ?
+				new BulkDeleteDescriptor<T>().Id(str) :
+				new BulkDeleteDescriptor<T>().Document(o));
 
 		/// <summary>
 		/// DeleteMany, convenience method to delete many objects at once.
--- a/src/Tests/Tests.Reproduce/GithubIssue3500.cs
+++ b/src/Tests/Tests.Reproduce/GithubIssue3500.cs
@@ -1,0 +1,22 @@
+﻿using System.Linq;
+using System.Collections.Generic;
+using Elastic.Xunit.XunitPlumbing;
+using FluentAssertions;
+using Nest;
+
+namespace Tests.Reproduce
+{
+	public class GithubIssue3500
+	{
+		[U] public void BulkDescriptorDeleteManyCallableWithStringWithoutTypeDeclaration()
+		{
+			var ids = new List<string> { "1" };
+
+			var bulkDescriptor = new BulkDescriptor().DeleteMany(ids);
+
+			if (bulkDescriptor is IBulkRequest bulkRequest) {
+				bulkRequest.Operations.First().Id.Should().Be(ids.First());
+			}
+		}
+	}
+}


russcam:
Thanks for opening @antja0.

There's an overload of `DeleteMany()` that accepts an `IEnumerable<string>` on `master`

https://github.com/elastic/elasticsearch-net/blob/c3301149632df4611bb09da7eb0526db6f3d8ae0/src/Nest/Document/Multiple/Bulk/BulkRequest.cs#L67-L72

that gets called with

```c#
var ids = new string[] 
{
	"1",
	"2",
	"3",
};

var client = new ElasticClient();

var bulkResponse = client.Bulk(b => b
	.DeleteMany<object>(ids)
);
```

which yields

```
POST http://localhost:9200/_bulk?pretty=true 
{"delete":{"_id":"1","_index":"default_index"}}
{"delete":{"_id":"2","_index":"default_index"}}
{"delete":{"_id":"3","_index":"default_index"}}
```

Am I missing something?

I'm on the fence about whether the `DeleteMany()` overloads that accept `long` and `string` should remain, or whether they should be unified into an overload that accepts `IEnumerable<Id>`

```c#
public BulkDescriptor DeleteMany<T>(
    IEnumerable<Id> ids,
    Func<BulkDeleteDescriptor<T>, string, IBulkDeleteOperation<T>> bulkDeleteSelector = null
) where T : class =>
    AddOperations(ids, bulkDeleteSelector, id => new BulkDeleteDescriptor<T>().Id(id));
```

This would satisfy all types that convert to `Id`. Calling would be

```c#
var ids = new Id[] 
{
	"1",
	"2",
	"3",
};

var client = new ElasticClient();

var bulkResponse = client.Bulk(b => b
	.DeleteMany<object>(ids)
);
```

antja0:
@russcam 
Hi,

I understood that the issue was that you had to specify the type argument:
```c#
.DeleteMany<object>(ids)
```
and you could not (would call wrong overload)
```c#
.DeleteMany(ids)
```
...when doing DeleteMany with strings.

This PR is aimed to fix this _so that the right overload is always called without type arguments_.

> I'm on the fence about whether the DeleteMany() overloads that accept long and string should ? remain, or whether they should be unified into an overload that accepts IEnumerable Id

^ I agree with this idea, but I don't know enough about the project to say if it's a good idea in practise.

russcam:
ah OK, I understand the confusion. 

Type inference will infer the wrong overload, so the generic type parameter is needed. I think I prefer the explicitness of specifying the generic type in this instance, rather than changing the behaviour of the `IEnumerable<T>` method to perform one operation when `T` is `string` and another when `T` is anything else.

antja0:
hmmm I agree it's a bit sketchy.
I like the idea of doing overload for `.DeleteMany` with `IEnumerable<Id>` , should I look into that in this same PR?
I'm guessing that would require some new explicit conversions for `Nest.Id`.

russcam:
> I like the idea of doing overload for .DeleteMany with IEnumerable<Id> , should I look into that in this same PR?

It's something that we will need to discuss internally first, but if it's a change we were looking to do for the next release, would you be interested in submititng a PR for it?

antja0:
> > I like the idea of doing overload for .DeleteMany with IEnumerable , should I look into that in this same PR?
> 
> It's something that we will need to discuss internally first, but if it's a change we were looking to do for the next release, would you be interested in submititng a PR for it?

Sure thing!

russcam:
Great! I've opened https://github.com/elastic/elasticsearch-net/issues/4125 to track this and discuss. For now, I'm going to close this PR and if https://github.com/elastic/elasticsearch-net/issues/4125 collectively looks like a good idea, a new PR can be opened for it. 

Thanks!

PR #4127, opened 2019-10-10
Deserialize boolean strings to dynamic mapping

### LINKED ISSUE #4124: Unable to get server side mappings due to unexpected boolean type (string)
Elastic 7.4.0 and NEST 7.3.0 are failing when I try to get the index mapping:

```
Elasticsearch.Net.Utf8Json.JsonParsingException: expected:'true | false', actual:'"false"', at offset:77
  at at Elasticsearch.Net.Utf8Json.JsonReader.ReadBoolean()
  at at Elasticsearch.Net.Utf8Json.Formatters.BooleanFormatter.Deserialize(JsonReader& reader, IJsonFormatterResolver formatterResolver)
  at at Nest.UnionFormatter`2.TryRead[T](ArraySegment`1& segment, IJsonFormatterResolver formatterResolver, T& v)
````

```
{
  "daily-logevents-v1-2019.10.08" : {
    "mappings" : {
      "dynamic" : "false",
      "properties" : {
        "companyId" : {
          "type" : "keyword"
        },
        "created" : {
          "type" : "alias",
          "path" : "createdUtc"
        },
        "createdUtc" : {
          "type" : "date"
        },
        "id" : {
          "type" : "keyword"
        },
        "message" : {
          "type" : "text",
          "fields" : {
            "keyword" : {
              "type" : "keyword",
              "ignore_above" : 256
            }
          }
        },
        "value" : {
          "type" : "integer"
        }
      }
    }
  }
}
```

image

As you can see I defined dynamic through the nest fluent api and this is breaking.

### PR DESCRIPTION
**Pull request by russcam**
This commit introduces a custom formatter for Union<bool,DynamicMapping>. A boolean value may be returned as a "true" or "false" string, which should be handled correctly.

Closes #4124

### DIFF
--- a/src/Nest/Mapping/DynamicMapping.cs
+++ b/src/Nest/Mapping/DynamicMapping.cs
@@ -1,5 +1,7 @@
 ﻿using System.Runtime.Serialization;
 using Elasticsearch.Net;
+using Elasticsearch.Net.Utf8Json;
+using Elasticsearch.Net.Utf8Json.Internal;
 
 
 namespace Nest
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Tests/Tests.Core/ManagedElasticsearch/NodeSeeders/DefaultSeeder.cs [hunk 1]
```diff
@@ -227,6 +227,7 @@ private async Task SeedIndexDataAsync()
 					.AutoMap()
 					.Name(p => p.Committer)
 					.Properties(DeveloperProperties)
+					.Dynamic()
```
Mpdreamz: 👍

src/Nest/Mapping/DynamicMapping.cs
```diff
@@ -13,4 +15,60 @@ public enum DynamicMapping
 		[EnumMember(Value = "strict")]
 		Strict
 	}
+
+	internal class DynamicMappingFormatter : IJsonFormatter<Union<bool,DynamicMapping>>
```
Mpdreamz: Yet another formatter :) 

Easier said then done but I wonder if we can formulate other strategies around this.

russcam: An alternative would be to have a more lenient `BoolFormatter` that can deserialize from string values. My initial resistance in going this direction is that it is a change that affects the performance of the general case of deserializing boolean values in order to handle the less frequent case of deserializing sting bool values.

Mpdreamz: Was talking more about making writing formatters easier. 

The generalized bool formatter might be a better solution functionality since even if we never write strings for booleans Elasticsearch will happily ingest `"true"` and return as is from _source. 

https://www.elastic.co/guide/en/elasticsearch/reference/current/boolean.html

So data that came in from other places might be stringified books too.

russcam: that's a good point. Perhaps we need to be more lenient on numeric types too?

PR #4129, opened 2019-10-11
Add support for clone index API

russcam:
This commit adds support for the clone index API, introduced in Elasticsearch 7.4.0

### DIFF
--- a/src/Nest/Indices/IndexManagement/CloneIndex/CloneIndexRequest.cs
+++ b/src/Nest/Indices/IndexManagement/CloneIndex/CloneIndexRequest.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Runtime.Serialization;
+
+namespace Nest
+{
+	/// <summary>
+	/// A request to the clone index API
+	/// </summary>
+	[MapsApi("indices.clone")]
+	[ReadAs(typeof(CloneIndexRequest))]
+	public partial interface ICloneIndexRequest
+	{
+		/// <summary>
+		/// The aliases to apply to the target index
+		/// </summary>
+		[DataMember(Name ="aliases")]
+		IAliases Aliases { get; set; }
+
+		/// <summary>
+		/// The settings to apply to the target index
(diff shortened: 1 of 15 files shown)


src/Nest/Indices/IndexManagement/CloneIndex/CloneIndexRequest.cs
```diff
@@ -0,0 +1,48 @@
...
+		[DataMember(Name ="aliases")]
+		IAliases Aliases { get; set; }
+
+		[DataMember(Name ="settings")]
+		IIndexSettings Settings { get; set; }
+	}
+
+	public partial class CloneIndexRequest
+	{
+		public CloneIndexRequest(IndexName index, IndexName target, IIndexState state) : this(index, target)
```
russcam: Should we support this overload, @Mpdreamz, @codebrain ?

Mpdreamz: That's a though one, I rather not since constructors are for path parameters only.

An alternative solution would be to expose `RemoveReadOnlySettings()` as an extension method on `IIndexSettings` and `IndexState`.

russcam: OK, I'll remove this overload for now

Mpdreamz:
I think something in the port to master went wrong:

https://dev.azure.com/azure3/elasticsearch-net/_build/results?buildId=2062

> Api/RequestParameters/RequestParameters.Indices.cs(106,15): error CS0534: 'CloneIndexRequestParameters' does not implement inherited abstract member 'RequestParameters<CloneIndexRequestParameters>.SupportsBody.get'

russcam:
weird, was a straight cherry-pick to master. Can look at first thing tomorrow 👍

russcam:
fixed in 7ef74650d5a69ee38d1cbd8a9844deb50f3df0e4.

PR #4132, opened 2019-10-13
Write null when ISort or SortKey is null

### LINKED ISSUE #4128: 7.0 Query serializer settings for trailing commas?
I'm using the latest version of (7.3.1) nest and I'm going to be digging into this but I figured I'd log it in case it's something simple but we are using nest to build up a query and we just got the following error: `json_parse_exception Reason: "Unexpected character (']' (code 93)): expected a value`

I'm tracking it down but it seems our sort query has a trailing `,` before the end of the sort array.

```json
  "sort": [
    {
      "test.long-000004": {
        "nested": {
          "filter": {
            "term": {
              "tester.userId": {
                "value": "5d9f919617d7af597fd8a451"
              }
            }
          },
          "path": "tester"
        },
        "order": "asc"
      }
    },
  ],
```

In our visitor we are setting the nodes content for sort to something like this:

```cs
new FieldSort
                        {
                            Field = node.Field,
                            Nested = new NestedSort
                            {
                                Path = path,
                                Filter = Query<TEST>.Term(f => f.UserId, userid)
                            },
                            Order = node.IsNodeOrGroupNegated() ? SortOrder.Descending : SortOrder.Ascending
                        }
```

russcam:
This commit writes null when ISort or ISort.SortKey is null. Simply returning without writing a value can result in incorrectly formatted JSON.

Closes #4128

### DIFF
--- a/src/Nest/CommonOptions/Sorting/SortFormatter.cs
+++ b/src/Nest/CommonOptions/Sorting/SortFormatter.cs
@@ -75,7 +75,10 @@ namespace Nest
 		public void Serialize(ref JsonWriter writer, ISort value, IJsonFormatterResolver formatterResolver)
 		{
 			if (value?.SortKey == null)
+			{
+				writer.WriteNull();
 				return;
+			}
 
 			writer.WriteBeginObject();
 			var settings = formatterResolver.GetConnectionSettings();


Mpdreamz:
LGTM the test that failed is:

```
Tests.ClientConcepts.ConnectionPooling.BuildingBlocks.ConnectionPooling.RandomizedInitialNodes
```

I guess this comment came true faster then I expected 

https://github.com/elastic/elasticsearch-net/pull/4112#discussion_r329547991

Mpdreamz:
Opened https://github.com/elastic/elasticsearch-net/pull/4134 for the failing test

PR #4134, opened 2019-10-14
Fix test for random initialized node

Mpdreamz:
The test was using a local method that yielded a single pool while later
.Take(50) was requested on this sequence. This always yields an
enumeration with 1 element which increases the chances for the assertion
to turn false.

Rather than relying on Thread.Sleep to create different seeded randoms
the StaticConnectionPool now defines a protected constructor that allows
one to pass in a seed value.

Continuation of #4112

### DIFF
--- a/src/Elasticsearch.Net/ConnectionPool/StaticConnectionPool.cs
+++ b/src/Elasticsearch.Net/ConnectionPool/StaticConnectionPool.cs
@@ -15,9 +15,15 @@ namespace Elasticsearch.Net
 			: this(uris.Select(uri => new Node(uri)), randomize, dateTimeProvider) { }
 
 		public StaticConnectionPool(IEnumerable<Node> nodes, bool randomize = true, IDateTimeProvider dateTimeProvider = null)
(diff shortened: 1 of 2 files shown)


src/Elasticsearch.Net/ConnectionPool/StaticConnectionPool.cs
```diff
@@ -15,27 +15,35 @@ public StaticConnectionPool(IEnumerable<Uri> uris, bool randomize = true, IDateT
 			: this(uris.Select(uri => new Node(uri)), randomize, dateTimeProvider) { }
 
 		public StaticConnectionPool(IEnumerable<Node> nodes, bool randomize = true, IDateTimeProvider dateTimeProvider = null)
+			: this(nodes, randomize, randomizeSeed: null, dateTimeProvider) { }
+
+		protected StaticConnectionPool(IEnumerable<Node> nodes, bool randomize, int? randomizeSeed = null, IDateTimeProvider dateTimeProvider = null)
```
russcam: I wonder if for master, we should consider accepting an instance of `Random` here instead of `randomize` and `randomizeSeed`

PR #4139, opened 2019-10-14
Add max_iterations configuration to watcher action with foreach execu…

Mpdreamz:
…tion

addresses elastic/elasticsearch#45715

### DIFF
--- a/src/Nest/XPack/Watcher/Action/ActionBase.cs
+++ b/src/Nest/XPack/Watcher/Action/ActionBase.cs
@@ -42,6 +42,9 @@ namespace Nest
 		[IgnoreDataMember]
 		string Foreach { get; set; }
 
+		[IgnoreDataMember]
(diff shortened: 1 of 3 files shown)


src/Nest/XPack/Watcher/Action/ActionBase.cs
```diff
@@ -42,6 +42,9 @@ public interface IAction
 		[IgnoreDataMember]
 		string Foreach { get; set; }
 
+		[IgnoreDataMember]
+		int? MaxIterations  { get; set; }
```
russcam:
```suggestion
        /// <summary>The maximum number of iterations that each watch executes. If this limit is reached,
        /// the execution is gracefully stopped. Defaults to <c>100</c>.
        /// </summary>
		int? MaxIterations  { get; set; }
```

PR #4141, opened 2019-10-14
Include in-progress snapshot for a policy with get SLM policy API

Mpdreamz:
addresses elastic/elasticsearch#45245

Note this only maps `start_time_millis` directly as `DateTimeOffset` in quite some places we map both `XContent.timeField` representations. Would be nice to clean this up in 8.0. Thoughts?

### DIFF
--- a/src/Nest/XPack/Slm/SnapshotLifecyclePolicyMetadata.cs
+++ b/src/Nest/XPack/Slm/SnapshotLifecyclePolicyMetadata.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Runtime.Serialization;
+using Elasticsearch.Net.Utf8Json;
 
 namespace Nest
 {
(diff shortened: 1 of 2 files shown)


src/Nest/XPack/Slm/SnapshotLifecyclePolicyMetadata.cs
```diff
@@ -45,5 +46,36 @@ public class SnapshotLifecyclePolicyMetadata
...
+		/// </summary>
+		[DataMember(Name = "in_progress")]
+		public InProgressLifecycleSnapshot InProgress { get; internal set; }
+	}
+
+	/// <summary>
+	/// If a snapshot is in progress when calling the Get Snapshot Lifecycle metadata
+	/// this will hold some minimal information about the in flight snapshot
+	/// </summary>
+	public class InProgressLifecycleSnapshot
```
russcam: `SnapshotLifecycleInProgress`, to be consistent with other types?

russcam:
> Note this only maps start_time_millis directly as DateTimeOffset in quite some places we map both XContent.timeField representations. Would be nice to clean this up in 8.0. Thoughts?

We can address these for 7.4.0 if we think it's good to. I was in two minds about mapping the `_millis*` properties to `DateTimeOffset` when implementing the SLM APIs. Thinking about it further, the millisecond/second since epoch properties for ML APIs are mapped to `DateTimeOffset`, so the SLM APIs should follow suit - it does make the `Human` property somewhat redundant in the client API.

src/Nest/XPack/Slm/SnapshotLifecyclePolicyMetadata.cs
Mpdreamz: ++

Mpdreamz:
> Thinking about it further, the millisecond/second since epoch properties for ML APIs are mapped to DateTimeOffset, so the SLM APIs should follow suit

Agreed, no sense in exposing these twice. The `human` parameter has little meaning in the high level client as is.

Mpdreamz:
ported to master

PR #4143, opened 2019-10-14
Search enhancement: pinned queries

Mpdreamz:
Addresses elastic/elasticsearch#44345

### DIFF
--- a/src/Nest/QueryDsl/Specialized/Pinned/PinnedQuery.cs
+++ b/src/Nest/QueryDsl/Specialized/Pinned/PinnedQuery.cs
@@ -1,0 +1,61 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using Elasticsearch.Net.Utf8Json;
+
+namespace Nest
+{
+	[InterfaceDataContract]
+	[ReadAs(typeof(PinnedQuery))]
+	public interface IPinnedQuery : IQuery
+	{
+		[DataMember(Name = "ids")]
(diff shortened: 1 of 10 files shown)


src/Nest/QueryDsl/Specialized/Pinned/PinnedQuery.cs
```diff
@@ -0,0 +1,61 @@
...
+	{
+		protected override bool Conditionless => PinnedQuery.IsConditionless(this);
+		IEnumerable<Id> IPinnedQuery.Ids { get; set; }
+		QueryContainer IPinnedQuery.Organic { get; set; }
+
+		public PinnedQueryDescriptor<T> Ids(params Id[] ids) => Assign(ids, (a, v) => a.Ids = v);
+
+		public PinnedQueryDescriptor<T> Ids(IEnumerable<Id> ids) => Ids(ids?.ToArray());
+
+		public PinnedQueryDescriptor<T> Ids(params string[] ids) => Assign(ids?.Select(v => (Id)v), (a, v) => a.Ids = v);
```
russcam: Should we support all overloads for types with implicit conversion to `Id`?

Mpdreamz: I removed the `params` overloads, without them you can actually do `.Ids(1, "", Guid.NewGuid())` to others remain handy to not force folks to cast to `Id`.

This was a copy pasta from the `IdsQuery` so added a note to clean that up in 8.x as well.

Mpdreamz:
merged to master

PR #4144, opened 2019-10-15
Fix test for random initialized node (#4134)

russcam:
The test was using a local method that yielded a single pool while later .Take(50)
was requested on this sequence. This always yields an enumeration with 1 element
which increases the chances for the assertion to turn false.

Rather than relying on Thread.Sleep to create different seeded randoms the StaticConnectionPool
now defines a protected constructor that allows one to pass in a seed value.

(cherry picked from commit 9734f04a161e6cbbe8b632b2993e61cb7f553a6a)

### DIFF
--- a/src/Elasticsearch.Net/ConnectionPool/StaticConnectionPool.cs
+++ b/src/Elasticsearch.Net/ConnectionPool/StaticConnectionPool.cs
@@ -12,18 +12,36 @@ namespace Elasticsearch.Net
 		private readonly Func<Node, float> _nodeScorer;
 
 		public StaticConnectionPool(IEnumerable<Uri> uris, bool randomize = true, IDateTimeProvider dateTimeProvider = null)
-			: this(uris.Select(uri => new Node(uri)), randomize, dateTimeProvider) { }
+			: this(uris.Select(uri => new Node(uri)), randomize, null, dateTimeProvider) { }
 
 		public StaticConnectionPool(IEnumerable<Node> nodes, bool randomize = true, IDateTimeProvider dateTimeProvider = null)
-			: this(nodes, null, randomize, dateTimeProvider) { }
+			: this(nodes, randomize, null, dateTimeProvider) { }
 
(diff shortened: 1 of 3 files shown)


russcam:
port to master of #4134

src/Elasticsearch.Net/ConnectionPool/StaticConnectionPool.cs
```diff
@@ -62,7 +77,7 @@ protected StaticConnectionPool(IEnumerable<Node> nodes, Func<Node, float> nodeSc
 		public virtual bool SupportsReseeding => false;
 
 		/// <inheritdoc />
-		public bool UsingSsl { get; }
+		public bool UsingSsl { get; set; }
```
Mpdreamz:
```suggestion
		public bool UsingSsl { get; }
```

src/Elasticsearch.Net/ConnectionPool/StaticConnectionPool.cs
```diff
@@ -75,10 +90,10 @@ protected List<Node> AliveNodes
 			}
 		}
 
-		protected IDateTimeProvider DateTimeProvider { get; }
+		protected IDateTimeProvider DateTimeProvider { get; set; }
```
Mpdreamz:
```suggestion
		protected IDateTimeProvider DateTimeProvider { get; }
```

src/Elasticsearch.Net/ConnectionPool/StaticConnectionPool.cs
russcam: Should be a private setter

src/Elasticsearch.Net/ConnectionPool/StaticConnectionPool.cs
russcam: Should be a private setter

russcam:
Would you mind taking a look again @Mpdreamz

Mpdreamz:
@russcam compilation failed on CI

russcam:
D'oh, fat fingered a `B` into source when running a build. Fixed now. Let this be a warning to future self when attempting to code in between conference sessions 🤦‍♂

russcam:
Merging this in; failing tests are not related to this change,

PR #4157, opened 2019-10-18
Fix GeoDistance serialization

rmja:
Fixes issue where unintended value separator is added. Example:
"sort":[{"_geo_distance":{,"some.coordinate":[{"lat":1.0,"lon":1.0}]
The "," is incorrectly added causing deserialization issues on the server.

### DIFF
--- a/src/Nest/CommonOptions/Sorting/SortFormatter.cs
+++ b/src/Nest/CommonOptions/Sorting/SortFormatter.cs
@@ -103,7 +103,7 @@ namespace Nest
 					for (var i = buffer.Offset; i < buffer.Count - 1; i++)
 						writer.WriteRawUnsafe(buffer.Array[i]);
 
-					if (buffer.Count > 0)
+					if (buffer.Count - buffer.Offset > 1)
 						writer.WriteValueSeparator();
 
 					writer.WritePropertyName(settings.Inferrer.Field(geo.Field));


russcam:
Thanks for the PR, @rmja. Would you be able to sign the [Contributor License Agreement](https://www.elastic.co/contributor-agreement) so that we may pull the change in?

rmja:
It should be signed now

src/Nest/CommonOptions/Sorting/SortFormatter.cs
```diff
@@ -103,7 +103,7 @@ public void Serialize(ref JsonWriter writer, ISort value, IJsonFormatterResolver
 					for (var i = buffer.Offset; i < buffer.Count - 1; i++)
 						writer.WriteRawUnsafe(buffer.Array[i]);
 
-					if (buffer.Count > 0)
+					if (buffer.Count - buffer.Offset > 1)
```
russcam:
```suggestion
					// does the IGeoDistanceSort have other properties set i.e. is it more than simply {} ?
					if (buffer.Count > 2)
```

russcam: `if (buffer.Count - buffer.Offset > 1)` will always be true because `buffer.Offset` is always 0 and `buffer.Count` will always be _at least_ 2.

rmja: If `buffer.Offset` is always 0, why is it then that the for-loop uses `var i = buffer.Offset`?

russcam: `i` could have been initialized to `0` instead, but I think using `buffer.Offset` is perhaps more clear about where the loop will start from?

russcam:
@rmja I'm going to make the changes and merge this in for the 7.4.0 release, if that's OK?

rmja:
@russcam absolutely 😀

russcam:
ported to 7.x

PR #4161, opened 2019-10-21
Add Circle processor

codebrain:
Relates to: https://github.com/elastic/elasticsearch/pull/43851

### DIFF
--- a/src/Nest/Ingest/Processors/DissectProcessor.cs
+++ b/src/Nest/Ingest/Processors/DissectProcessor.cs
@@ -73,6 +73,7 @@ namespace Nest
 		public DissectProcessorDescriptor<T> Pattern(string pattern) =>
 			Assign(pattern, (a, v) => a.Pattern = v);
 
+		// TODO! rename the parameter in 8.0
 		/// <inheritdoc cref="IDissectProcessor.IgnoreMissing">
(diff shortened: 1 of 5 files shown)


src/Nest/Ingest/Processors/DissectProcessor.cs
```diff
@@ -73,6 +73,7 @@ public class DissectProcessorDescriptor<T>
 		public DissectProcessorDescriptor<T> Pattern(string pattern) =>
 			Assign(pattern, (a, v) => a.Pattern = v);
 
+		// TODO! rename the parameter in 8.0
```
russcam: I think we can rename this here. I understand that it is technically a binary breaking change

Mpdreamz: Let's rename in 8.0 instead, eventhough impact is small.

Mpdreamz:
1 test failure, unrelated: https://github.com/elastic/elasticsearch-net/issues/4167

src/Nest/Ingest/Processors/DissectProcessor.cs
russcam: ok

russcam:
ported to master

PR #4169, opened 2019-10-22
Fix aggs integration tests where buckets can be empty

russcam:
This commit fixes integration tests for aggregations that use a date_histogram aggregation,
where a bucker may have no documents with a specific seed value. For example,

```
build.bat seed:41075 integrate 7.4.0 "readonly"
```


russcam:
ported to master

PR #4170, opened 2019-10-22
Add Update and Update by query examples

russcam:
This commit adds Update and Update by query API examples.

Add ApplyBodyChanges method to Example type to refactor manipulating the JSON string body into one place.


russcam:
Failing integration tests are not related to these changes, so merging in. A separate PR will be opened to fix these tests for master, likely caused by upstream changes in Elasticsearch that are not yet accommodated in the client.

PR #4173, opened 2019-10-23
Convert SLM API dates returned as long to DateTimeOffset

russcam:
This commit updates the SLM API responses to convert date properties that are returned long values representing milliseconds since epoch
to DateTimeOffset. This is consistent with other API responses where longs are returned. What is slightly different with
this API is that Human=true can be specified on the request to also return the properties as string values. With this change, the client
now effectively ignores these values as they serve little purpose, since both would end up being converted to DateTimeOffset when deserialized.


russcam:
ported to master

PR #4177, opened 2019-10-23
Fix shape serialisation tests

codebrain:
- Remove circle from Shape tests
 - Change integration test setup to use Shape() instead of GeoShape()
 - Change default index for Shape types to be `shapes`


russcam:
ported to master

PR #4192, opened 2019-11-01
update abstractions to deal with 8.0.0-SNAPSHOT changes

Mpdreamz:


Mpdreamz:
1 failure was fixed in 7.x

The other is a cluster timeout intermittent failure.

PR #4202, opened 2019-11-11
Allow MemoryStreamFactory to be pluggable

Mpdreamz:


src/Elasticsearch.Net/Providers/MemoryStreamFactory.cs
```diff
@@ -2,11 +2,14 @@
 
 namespace Elasticsearch.Net
 {
+	// TODO we use this in some places but its no longer clear to me why need to circumvent RecyclableMemoryStream in some cases
 	/// <summary>
 	/// A factory for creating memory streams using instances of <see cref="MemoryStream" />
 	/// </summary>
 	public class MemoryStreamFactory : IMemoryStreamFactory
 	{
+		public static MemoryStreamFactory Default { get; } = new MemoryStreamFactory();
```
russcam: Perhaps we should rename these to `.Instance` in future?

Mpdreamz: I'm happy to role with `.Instance` 👍

Mpdreamz:
ported to `master`

PR #4191, opened 2019-10-31
Fix/7.x/bulk formatter

Mpdreamz:
Continuation of #3797 

this consolidates someof the places we either serialize to bytes or call `ToArray()`. 

This also references `Nest7` from our nuget feed, this is a versioned package where all types live in `Nest7` and `Easticsearch.Net7` . This has been around for a while see: https://www.elastic.co/blog/nest-and-elasticsearch-net-upgrading-your-codebase

Here we use it to profile against the integration branch.

On bulk the write path is still a lot hotter then the read path, this does not fix that but does improve allocations.

|                Method |       Jit |       Runtime |     Mean |    Error |   StdDev |      Gen 0 |     Gen 1 | Gen 2 |   Allocated |
|---------------------- |---------- |-------------- |---------:|---------:|---------:|-----------:|----------:|------:|------------:|
| 'NEST updated Bulk()' | LegacyJit |    .NET 4.7.2 | 511.0 ms |  6.89 ms |  9.89 ms |  8000.0000 | 2000.0000 |     - |           - |
|         'NEST Bulk()' | LegacyJit |    .NET 4.7.2 | 470.4 ms | 10.76 ms | 15.44 ms | 16000.0000 | 4000.0000 |     - |           - |
| 'NEST updated Bulk()' |    RyuJit | .NET Core 3.0 | 304.8 ms |  4.38 ms |  6.56 ms |  8000.0000 | 2000.0000 |     - | 153176144 B |
|         'NEST Bulk()' |    RyuJit | .NET Core 3.0 | 317.4 ms |  4.23 ms |  5.93 ms | 16000.0000 | 4000.0000 |     - | 199605512 B |

PR #4204, opened 2019-11-11
Fix/7.x/upgrade recyclable memory stream factory

Mpdreamz:
This upgrades to latest `RecyclableMemoryStreamFactory`

It also adds our own counters into it, isolated from #4191


Mpdreamz:
Aye pretty much, I just extended it to include dotnet performance counters.

PR #4172, opened 2019-10-22
Memory performance counters

Mpdreamz:
This PR updates us to the latest RecycableMemoryStreamManager from:

https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream

This now implements `GetBuffer()` so we can investigate removing our `ToArray()` calls. 
We have a pending PR for this for a while now here: https://github.com/elastic/elasticsearch-net/pull/3797

I've extended it so that it emits metrics using the counters `System.Diagnostics.Tracing.*Counter` 

To aid with visualizing and testing this I imported @russcam [StackOverflow indexer](https://github.com/russcam/stack-overflow) to our codebase (as `Tests.LongRunning`).

I've added `--profile` switch that emits the process id on run and waits for confirmation to run on stdin. At the end of the run it will ask to run again if `--profile` is specified.

This allows you to run `dotnet counters` and other profiles/visualizers/tracers:

Start a cluster: 
```
$ ./build.sh cluster readonly
```

Indexing the askubuntu stack exchange questions and answers:

```
$  dotnet run -c RELEASE -- posts -e http://localhost:9200 -f /home/mpdreamz/datasets/askubuntu/Posts.xml --profile
```

Will emit:

> [17:34:35.074] Current Process Id: 5290
> [17:34:35.080] Press any key to continue...

From another console starting:

```
dotnet counters monitor --refresh-interval 1 Elasticsearch-Net-RecyclableMemoryStream -p 5290
```
```
[Elasticsearch-Net-RecyclableMemoryStream]
    Active memory streams                              0    
    Large buffers active                               0    
    Large buffers free                                 0    
    Large pool in use size                             0    
    Pooled blocks active                             300    
    Small pool free blocks                    39,321,600    
    Small pool in use size                             0    
    Small pool max size                                0 
```

And after a 2nd run:

```
[Elasticsearch-Net-RecyclableMemoryStream]
    Active memory streams                              0    
    Large buffers active                               0    
    Large buffers free                                 0    
    Large pool in use size                             0    
    Pooled blocks active                             300    
    Small pool free blocks                    39,321,600    
    Small pool in use size                             0    
    Small pool max size                                0 
```

Thats `40mb` after indexing a `1GB` xml file twice.

Pools are reused but there is currently no upper limit to this pool this means that the pool will be as large as the users heaviest load and most likely stays around there. 

- Is this bad? Do we want to tackle this?
- Default buffer size is quite high would we benefit from a lower buffer size?
[...]

PR #4206, opened 2019-11-11
Adds versioned benchmarking

Mpdreamz:
Allows the benchmarking project to reference Nest.v7 and so makes it easier for us to benchmark proposed changes.

### DIFF
--- a/src/Tests/Tests.Benchmarking/BenchmarkProgram.cs
+++ b/src/Tests/Tests.Benchmarking/BenchmarkProgram.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.InteropServices;
 using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Diagnosers;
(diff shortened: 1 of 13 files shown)


Mpdreamz:
@codebrain @russcam please review this first, opened a bunch of PR's today but this one is one that I would really like to be in tomorrow so I can start doing PR's with benchmarks.

Mpdreamz:
I opened https://github.com/elastic/elasticsearch-net/pull/4213 to address the integration test failures in this PR.

src/Tests/Tests.Benchmarking/BenchmarkProgram.cs
```diff
@@ -61,14 +62,15 @@ public static int Main(string[] arguments)
 
 		private static IConfig CreateDefaultConfig()
 		{
-			var jobs = new[]
+			var jobs = new List<Job>
 			{
-				Job.ShortRun.With(Runtime.Core).With(Jit.RyuJit),
-				Job.ShortRun.With(Runtime.Clr).With(Jit.RyuJit),
-				Job.ShortRun.With(Runtime.Clr).With(Jit.LegacyJit),
+				Job.MediumRun.With(CoreRuntime.Core30).With(Jit.RyuJit),
```
russcam: Should we be testing all the TFMs that the client supports, or at least the latest patch of each minor?

Mpdreamz: Fair comment, I'm okay with just running with the latest greatest on core.

For now at least I am more interested in the relative performance of branches vs mainline than the performance between platforms. Possibly something to consider in the long run I reckon.

PR #4207, opened 2019-11-11
RestSpecDownloader uses windows path separator

Mpdreamz:

### DIFF
--- a/src/CodeGeneration/ApiGenerator/RestSpecDownloader.cs
+++ b/src/CodeGeneration/ApiGenerator/RestSpecDownloader.cs
@@ -60,7 +60,7 @@ namespace ApiGenerator
 
 		private void FindJsonFilesOnListing(Specification spec, string html, IProgressBar pbar)
 		{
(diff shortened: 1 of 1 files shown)


src/CodeGeneration/ApiGenerator/RestSpecDownloader.cs
```diff
@@ -92,7 +92,7 @@ private void WriteToEndpointsFolder(string folder, string filename, string conte
 		{
 			var f = Path.Combine(GeneratorLocations.RestSpecificationFolder, folder);
 			if (!Directory.Exists(f)) Directory.CreateDirectory(f);
-			File.WriteAllText(f + "\\" + filename, contents);
```
russcam: Possibly simpler with another `Path.Combine(f, filename)`?

Mpdreamz: yarp, amended 😄

Mpdreamz:
ported to `master`

PR #4208, opened 2019-11-11
Use MemoryStreamFactory everywhere

Mpdreamz:
Found a few places, that still new'ed MemoryStream directly (outside of client mainly)


Mpdreamz:
ported to `master`

PR #4209, opened 2019-11-11
Adds more write overloads to Utf8Json

Mpdreamz:
Either memory stream or to serialize on the writer.

This consolidates and funnels places that need to write bytes/streams/objects directly.

### DIFF
--- a/src/Elasticsearch.Net/Serialization/DiagnosticsSerializerProxy.cs
+++ b/src/Elasticsearch.Net/Serialization/DiagnosticsSerializerProxy.cs
@@ -32,12 +32,16 @@ namespace Elasticsearch.Net
 		public override string ToString() => _stringRepresentation;
 	}
 
+	//TODO this no longer needs to be IInternalSerializerWithFormatter
+	// Previous we checked that the serializer we are wrapping implements the interface and if so set the formatter
+	// However its totally fine for formatter to be null. `IJsonFormatter` solves this better.
 	/// <summary>
 	/// Wraps configured serializer so that we can emit diagnostics per configured serializer.
 	/// </summary>
-	internal class DiagnosticsSerializerProxy : IElasticsearchSerializer, IInternalSerializerWithFormatter
+	internal class DiagnosticsSerializerProxy : IElasticsearchSerializer, IInternalSerializerWithFormatter, IInternalSerializer
 	{
 		private readonly IElasticsearchSerializer _serializer;
+		private readonly bool _wrapsUtf8JsonSerializer;
 		private readonly SerializerRegistrationInformation _state;
 		private readonly IJsonFormatterResolver _formatterResolver;
 		private static DiagnosticSource DiagnosticSource { get; } = new DiagnosticListener(DiagnosticSources.Serializer.SourceName);
@@ -47,10 +51,24 @@ namespace Elasticsearch.Net
 			_serializer = serializer;
 			_state = new SerializerRegistrationInformation(serializer.GetType(), purpose);
 			if (serializer is IInternalSerializerWithFormatter withFormatter)
+			{
 				_formatterResolver = withFormatter.FormatterResolver;
+				_wrapsUtf8JsonSerializer = true;
+			}
+			else
+			{
+				_wrapsUtf8JsonSerializer = false;
+				_formatterResolver = null;
+			}
 		}
 
 		public IJsonFormatterResolver FormatterResolver => _formatterResolver;
+
+		public bool TryGetJsonFormatter(out IJsonFormatterResolver formatterResolver)
+		{
+			formatterResolver = _formatterResolver;
+			return _wrapsUtf8JsonSerializer;
+		}
 
 		public object Deserialize(Type type, Stream stream)
 		{
--- a/src/Elasticsearch.Net/Utf8Json/IJsonFormatterResolver.cs
+++ b/src/Elasticsearch.Net/Utf8Json/IJsonFormatterResolver.cs
@@ -23,6 +23,7 @@
 #endregion
 
 using System;
+using System.Collections.Concurrent;
 using System.Reflection;
 
 namespace Elasticsearch.Net.Utf8Json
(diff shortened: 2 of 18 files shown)


src/Nest/CommonAbstractions/SerializationBehavior/JsonFormatters/SourceFormatter.cs
```diff
@@ -30,23 +30,17 @@ public virtual void Serialize(ref JsonWriter writer, T value, IJsonFormatterReso
 			var settings = formatterResolver.GetConnectionSettings();
 
 			// avoid serialization to bytes when not using custom source serializer
-			if (ReferenceEquals(settings.SourceSerializer, settings.RequestResponseSerializer))
+			if (ReferenceEquals(settings.SourceSerializer, settings.RequestResponseSerializer)
+				|| settings.SourceSerializer is IInternalSerializerWithFormatter s && s.FormatterResolver != null)
```
russcam: Would you be able to explain this change?

src/Nest/Document/Multiple/Bulk/BulkRequestFormatter.cs
```diff
@@ -42,12 +41,7 @@ public void Serialize(ref JsonWriter writer, IBulkRequest value, IJsonFormatterR
 				if (body == null)
 					continue;
 
-				var bodySerializer = op.Operation == "update" || body is ILazyDocument
-					? requestResponseSerializer
-					: sourceSerializer;
-
-				var bodyBytes = bodySerializer.SerializeToBytes(body, memoryStreamFactory, SerializationFormatting.None);
-				writer.WriteRaw(bodyBytes);
+				SourceWriter.Serialize(ref writer, body, formatterResolver);
```
russcam: This looks like a behavioural change?

src/Elasticsearch.Net/Utf8Json/IJsonFormatterResolver.cs
```diff
@@ -60,11 +61,10 @@ public static IJsonFormatter<T> GetFormatterWithVerify<T>(this IJsonFormatterRes
             return formatter;
         }
 
+		private static readonly MethodInfo _getFormatterMethod = typeof(IJsonFormatterResolver).GetRuntimeMethod("GetFormatter", Type.EmptyTypes);
```
russcam: ++ to making a static field

src/Nest/CommonAbstractions/SerializationBehavior/JsonFormatters/SourceFormatter.cs
Mpdreamz: I moved this out into its own method with `xmldocs` and used the check in `Deserialize` too.

Please let me know if this change make sense now and thank you for calling out to explain why and what!

src/Nest/Document/Multiple/Bulk/BulkRequestFormatter.cs
Mpdreamz: Because `SerializationFormatting.None` is not specified explicitly?

src/Nest/CommonAbstractions/SerializationBehavior/JsonFormatters/SourceFormatter.cs
russcam: The bit that I'm unsure about is the additional OR check

```
settings.SourceSerializer is IInternalSerializerWithFormatter s && s.FormatterResolver != null)
```

is there a change elsewhere that now requires this?

src/Nest/Document/Multiple/Bulk/BulkRequestFormatter.cs
russcam: The behavioural change looks like the `SourceWriter` is now **always** used. 

Before, the `requestResponseSerializer` was used when the operation was an update or the body was an `ILazyDocument`

src/Nest/CommonAbstractions/SerializationBehavior/JsonFormatters/SourceFormatter.cs
Mpdreamz: Addressed that in a follow up commits

https://github.com/elastic/elasticsearch-net/blob/f365b3070b75fecae14f176d7faf50010a87dfff/src/Nest/CommonAbstractions/SerializationBehavior/JsonFormatters/SourceFormatter.cs#L28-L36

src/Nest/Document/Multiple/Bulk/BulkRequestFormatter.cs
Mpdreamz: Ahh you are right, will address!

src/Elasticsearch.Net/Serialization/DiagnosticsSerializerProxy.cs
```diff
@@ -32,12 +32,16 @@ public SerializerRegistrationInformation(Type type, string purpose)
 		public override string ToString() => _stringRepresentation;
 	}
 
+	//TODO this no longer needs to be IInternalSerializerWithFormatter
```
russcam: Does `DiagnosticsSerializerProxy` still need to implement `IInternalSerializerWithFormatter`?

src/Nest/CommonAbstractions/SerializationBehavior/JsonFormatters/SourceFormatter.cs
russcam: got it, thanks!

src/Elasticsearch.Net/Serialization/DiagnosticsSerializerProxy.cs
Mpdreamz: Remove it instead since its internal

Mpdreamz:
Closing in favor of: https://github.com/elastic/elasticsearch-net/pull/4313

Addressed remaining feedback there as well.

PR #4210, opened 2019-11-11
reintroduce skipdocs for default build target

Mpdreamz:


russcam:
ported to 7.4 and master

PR #4212, opened 2019-11-11
normalize ifdefs in our source included UTF8Json

Mpdreamz:
Removes most of the compiler directives known to be false for the TFM's we publish.

Removes the need to define `NETSTANDARD` for our `net41` TFM.

### DIFF
--- a/src/Elasticsearch.Net/Utf8Json/Formatters/ValueTupleFormatter.cs
+++ b/src/Elasticsearch.Net/Utf8Json/Formatters/ValueTupleFormatter.cs
@@ -22,9 +22,7 @@
 // SOFTWARE.
 #endregion
 
-
 using Elasticsearch.Net.Utf8Json.Internal;
-#if NETSTANDARD
 using System;
 
 namespace Elasticsearch.Net.Utf8Json.Formatters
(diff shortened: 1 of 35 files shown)


src/Elasticsearch.Net/Utf8Json/Formatters/ValueTupleFormatter.cs
```diff
@@ -174,7 +172,7 @@ static ValueTupleFormatterHelper()
 		}
 	}
 
-	#if DOTNETCORE
+	#if NETSTANDARD
```
russcam: Should this be `NETSTANDARD2_1`?

Mpdreamz: `ValueTuple<>` exists in `netstandard2.0` 
`ValueTask<>` only exists in `netstandard2.1`

Mpdreamz:
>  Have the NETSTANDARD directive blocks been tested on net461?

@russcam we used to define `NETSTANDARD` for all TFM's specifically to satisfy UTF8Json, main reason for this PR is to get rid of that constant definition.

Mpdreamz:
ported to `master`

PR #4219, opened 2019-11-13
Serialize own IProperty implementations

### LINKED ISSUE #3655: Support extending NEST types by deriving and implementing properties
With the move to utf8json in #3493, the serialization of NEST types is strict, serializing only those properties specified on implemented interfaces in many places. For example, `PropertyFormatter` only serializes the `I*Property` interface members:

https://github.com/elastic/elasticsearch-net/blob/f3ec7198a979da2fe11ef8e14fbe69b072c92a43/src/Nest/Mapping/Types/PropertyFormatter.cs#L97-L192

With such strictness, it is not currently possible to derive a type from a NEST type, add additional properties, and have those properties serialized by the internal serializer. This is demonstrated in `InjectACustomIPropertyImplementation()` 

https://github.com/elastic/elasticsearch-net/blob/e2b7f7529976d8a47a8db37d9dbb6b6bbcd8d253/src/Tests/Tests/ClientConcepts/HighLevel/Serialization/ExtendingNestTypes.doc.cs#L53-L97

utf8json is able to support this feature through using the fallback formatter, `DynamicObjectTypeFallbackFormatter` ,exposed by resolving a `IJsonFormatter<object>`, which internally inspects the type and generates a custom serialization method for it, caching it in a `ThreadsafeTypeKeyHashTable<KeyValuePair<object, SerializeMethod>>`. This path looks like it would be a plausible way to support extending NEST types but requires some guarding when serializing Attribute types; when a type to serialize is an `Attribute` type, the walking of properties performed by `MetaType` traverses down to a framework enum type with multiple members with the same undelying value, which trips up the `EnumFormatter`. For an Attribute type, this walking should not be performed.

### LINKED ISSUE #4217: Custom property's attributes are not serialized to index mapping
**NEST/Elasticsearch.Net version**: 7.4.0

**Elasticsearch version**: 7.4.0

Hi, while upgrading our codebase from 6.8 to 7.4, we discovered a change of behavior that I cannot attribute to any of the breaking changes. Is this maybe caused by the change of the JSON serializer?

I create an object property with a custom property, but only the `type` gets serialized and the rest of the properties is ignored. In 6.8 it would also serialize the value of `language`. I've tried to mark them as `DataContract`/`DataMember` with no success.

When checking the descriptor's private properties, I can see that everything seems correct, just the serialization somehow ignores the properties.

Help would be appreciated, thanks.

**Steps to reproduce**:
 1. Create a custom property, e.g.:
```c#
    public class SpecialKeywordProperty : IProperty
    {
        public SpecialKeywordProperty(string language)
        {
            Type = "special_keyword";
            Language = language;
        }

        public PropertyName Name { get; set; }

        public IDictionary<string, object> LocalMetadata { get; set; }

        [PropertyName("language")]
    }
```
 2. Create a property visitor that's going to apply the property (on all object properties for simplicity):
```c#
    public class MappingPropertyVisitor : NoopPropertyVisitor
    {
        public override void Visit(IObjectProperty type, PropertyInfo propertyInfo, ElasticsearchPropertyAttributeBase attribute)
        {
            type.Properties.Clear();
            type.Properties.Add("special_field", new SpecialKeywordProperty("en"));
        }
    }
```
 3. Create a document POCO:
```c#
    class Doc
    {
        [Object]
        public string Whatever { get; set; }
    }
```
 4. Create a `CreateIndexDescriptor` mapping the document:
```c#
    new CreateIndexDescriptor("test-index")
             .Map<Doc>(m => m
                 .AutoMap<Doc>(new MappingPropertyVisitor())
                 .Dynamic(DynamicMapping.Strict)
             );
```
 5. Create the index or just serialize the mapping, the property `Language` will not be serialized
```javascript
{
  "mappings": {
    "dynamic": "strict",
    "properties": {
      "whatever": {
        "properties": {
          "special_field": {
            "type": "special_keyword"
          }
        },
        "type": "object"
      }
    }
  }
}

```

russcam:
This commit fixes the ability to serialize own IProperty types within PropertyFormatter, using utf8json. Initially, this feature was not originally available in utf8json due to how enums with the same underlying values but different field names were serialized, and how this interacts with internal enums of attribute runtime types. Since this is now fixed in #elastic/elasticsearch-net/4032, custom IProperty implementations will now serialize correctly.

Closes #3655 
Closes #4217

### DIFF
--- a/docs/client-concepts/connection/configuration-options.asciidoc
+++ b/docs/client-concepts/connection/configuration-options.asciidoc
@@ -190,6 +190,10 @@ Instead of following a c/go like error checking on response.IsValid do throw an 
 +
 Reasons for such exceptions could be search parser errors, index missing exceptions, etc...
 
+`TransferEncodingChunked`::
+
+Whether the request should be sent with chunked Transfer-Encoding. Default is `false`
+
 `UserAgent`::
 
 The user agent string to send with requests. Useful for debugging purposes to understand client and framework versions that initiate requests to Elasticsearch
--- a/docs/client-concepts/high-level/serialization/extending-nest-types.asciidoc
+++ b/docs/client-concepts/high-level/serialization/extending-nest-types.asciidoc
@@ -15,8 +15,16 @@ please modify the original csharp file found at the link and submit the PR with 
 [[extending-nest-types]]
 === Extending NEST types
 
-Sometimes you might want to provide a custom implementation of a type perhaps to work around an issue or because
+Sometimes you might want to provide a custom implementation of a type, perhaps to work around an issue or because
 you're using a third-party plugin that extends the features of Elasticsearch, and NEST does not provide support out of the box.
+
+NEST allows extending its types in some scenarios, discussed here.
+
+==== Creating your own property mapping
+
+As an example, let's imagine we're using a third party plugin that provides support for additional data type
+for field mapping. We can implement a custom `IProperty` implementation so that we can use the field mapping
+type with NEST.
 
 [source,csharp]
 ----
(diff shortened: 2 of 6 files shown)


russcam:
ported to 7.4 and master

PR #4222, opened 2019-11-13
Add support for WKT with GeoLocation type

russcam:
Relates: https://github.com/elastic/elasticsearch/pull/44107

This commit adds support for Well Known Text (WKT) representations for GeoLocation, the type used
to represent geo_point in the client. Similar to WKT support in IGeoShape, an internal format field
is used to store whether an instance was deserialized from JSON or WKT, so that re-serialization
honours the original form deserialized from.


russcam:
ported to 7.4 and master

Issue #4203, opened 2019-11-11
Canary builds fail on windows CI (7.x)


Mpdreamz:
Canary builds fail on windows CI (7.x)

https://dev.azure.com/azure3/elasticsearch-net/_build/results?buildId=2340

>D:\a\1\s\build\output\_packages\tmp\lib\netstandard2.0\Elasticsearch.Net7.dll was signed with official key token 96c599bbe3e70f5d
Assembly: 7.0.0.0 File: 7.5.0.0 Product: 7.5.0-ci20191111T104334 => D:\a\1\s\build\output\_packages\tmp\lib\net461\Elasticsearch.Net.VirtualizedCluster.dll
[90mscripts:[0m [34mvalidate-artifacts[90m:[0m [31mSystem.Collections.Generic.KeyNotFoundException: An index satisfying the predicate was not found in the collection.
   at Microsoft.FSharp.Collections.SeqModule.loop@911-22[T](FSharpFunc`2 predicate, IEnumerator`1 ie, Int32 i) in E:\A\_work\130\s\src\fsharp\FSharp.Core\seq.fs:line 917
   at Microsoft.FSharp.Collections.SeqModule.FindIndex[T](FSharpFunc`2 predicate, IEnumerable`1 source) in E:\A\_work\130\s\src\fsharp\FSharp.Core\seq.fs:line 918
   at Scripts.Versioning.validate(String dll, String name) in D:\a\1\s\build\scripts\Versioning.fs:line 114
   at Scripts.Versioning.validateDllStrongName(String dll, String name) in D:\a\1\s\build\scripts\Versioning.fs:line 134
   at Scripts.Versioning.ValidateArtifacts@144-1.Invoke(String f) in D:\a\1\s\build\scripts\Versioning.fs:line 152
   at Microsoft.FSharp.Collections.SeqModule.Iterate[T](FSharpFunc`2 action, IEnumerable`1 source) in E:\A\_work\130\s\src\fsharp\FSharp.Core\seq.fs:line 495
   at Scripts.Versioning.ValidateArtifacts@141.Invoke(String f) in D:\a\1\s\build\scripts\Versioning.fs:line 143
   at Microsoft.FSharp.Collections.SeqModule.Iterate[T](FSharpFunc`2 action, IEnumerable`1 source) in E:\A\_work\130\s\src\fsharp\FSharp.Core\seq.fs:line 495
[...]

PR #4223, opened 2019-11-14
Diff release version against previous version from Nuget

russcam:
This commit adds a new build target to diff the release version against the previous version on Nuget, to identify breaking API changes. This new build target is added as a dependent target of the release target.


russcam:
There's an issue open for the failing CI tests #4203 which is unrelated to these changes, as it was opened before this PR. merging this in.

russcam:
ported to 7.4 and master

PR #4224, opened 2019-11-14
Sourcelink and include PDBs in nuget packages

### LINKED ISSUE #4095: Double check our source link support
We support sourcelink for a long time now but I haven't double checked if we package this correctly  for a similar long while either.

Creating this ticket to remember to investigate our support for this feature is still good.

russcam:
This commit reinstates sourcelinking of PDB files and including them in nuget packages.

Closes #4095


russcam:
ported to master, 7.4, 6.x

PR #4227, opened 2019-11-15
Sync 7.x with master

Mpdreamz:
This PR syncs some of the changes that went into `master` that did not make it into `7.x`. These are mainly stylistic/whitespace changes.

Doing this so that if we do a folder diff mainly functional changes will light up.


Mpdreamz:
@codebrain @russcam merging this in to start work on other bits.

PR #4230, opened 2019-11-18
Share random seed between test framework and build scripts

Mpdreamz:
Our build scripts allow you to suffix certain targets with `-one` so that they will test one TFM. 

We use this on our CI to minimize run time e.g we have multiple integration tests for each minor, we only run one random TFM per minor on each run. This is a trade-off between full coverage vs sane CI run time.

Our test framework has a lot of randomness as well and decided on a seed when run to ensure we can reproduce failures using the exact same randomness. When tests fail a reproduce line is printed (and surfaced as Github Checks).

This PR now decides the seed in the build scripts so that the randomness in the build script and test framework are both based on the same seed.

### DIFF
--- a/build/scripts/Commandline.fs
+++ b/build/scripts/Commandline.fs
@@ -93,7 +93,7 @@ Execution hints can be provided anywhere on the command line
         NonInteractive: bool;
         SkipTests: bool;
         GenDocs: bool;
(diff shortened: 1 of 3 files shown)


Mpdreamz:
backported to `7.x`

PR #4233, opened 2019-11-20
Bump 7.x to 7.5

Mpdreamz:
This bumps our `7.x` over to `7.5`

Updated code generation
Handle quirks in the spec introduced in 7.5

New `time` enum is different from common option's time, the spec includes descriptions in the values.
Same problem for `bytes`.

`local` and `master_timeout` were removed in a minor. These were falsely documented as the server did not accept them. We add them back in obsolete fashion.

Some API endpoints remove `force` from `version_type` but not all eventhough it was removed in a minor on the server from all endpoints. We still need to keep it around though because our `7.5` client can also speak to `7.0` Elasticsearch that does still accept the parameter.


src/Nest/Descriptors.Cat.cs
```diff
@@ -86,8 +86,6 @@ public CatAllocationDescriptor(NodeIds nodeId): base(r => r.Optional("node_id",
 		///<summary>A comma-separated list of node IDs or names to limit the returned information</summary>
 		public CatAllocationDescriptor NodeId(NodeIds nodeId) => Assign(nodeId, (a, v) => a.RouteValues.Optional("node_id", v));
 		// Request parameters
-		///<summary>The unit in which to display byte values</summary>
-		public CatAllocationDescriptor Bytes(Bytes? bytes) => Qs("bytes", bytes);
```
Mpdreamz: Should not exclude bytes will do a follow up PR

Issue #4196, opened 2019-11-06
The API doesn't type


haf-afa:
1. The API doesn't type, I have to use `op_Implicit` in a bunch of places when I declare index names.
1. The API doesn't type, all the primary functions to actually index stuff take a `Func<SomethingDescriptor<'TDocument>, ISomethingRequest>` but the builder doesn't have a `Finish()` or `Build()` method that converts the descriptor to an actual request. When searching types in this repository there's a bunch of empty stubs, such as `IIndexRequest` and `IndexRequest` but with no implementations and no docs and there are no current docs on the homepage nor in the readme and all examples use the broken invocation. I presume there's a C# implicit conversion somewhere from the descriptor to the request interface, but how to call it is completely undocumented. :( 
1. When adding first a property, `app` and then `app.kubernetes.io`, I would expect them to map to two mappings for two properties, but ES interprets it as one and the same, erroring out with: `"org.elasticsearch.index.mapper.MapperParsingException: Could not dynamically add mapping for field [app.kubernetes.io/instance]. Existing mapping for [labels.app] must be of type object but found [text]."` (`labels` is the parent object) This could be type-safely done, perhaps? Because if the call types, it should work.

Example for 2 that I haven't figured out yet:

```fsharp
  let createBuilder (x: CreateIndexDescriptor): ICreateIndexRequest =
    x.Map<CRD>(fun x -> x.AutoMap().Properties(fun y -> y) :> IPromise<_>) :> ICreateIndexRequest
// doesn't work
```

Workaround for 2:

```fsharp
let indexDocument (client: ElasticClient) (doc: CRD) =
  client.IndexAsync<CRD>(doc, fun i -> i.Id(doc.getId() |> Id.op_Implicit) :> IIndexRequest<_>)
    |> Async.AwaitTask
```

Actionable:

- Create a few F# examples to ensure it's not so utterly frustrating to start using in a language that is not C#
- Document what is what, and that a descriptor is also a request
- Instead of overloading functions, and especially so if you have generic parameters in them, create two functions with clear and distinct names reflecting their usage, so I can choose whether to use a builder or construct the XXXRequest object outside of the call to the function.
- Document whether the callbacks to the builder execute only once per type or once per document.

russcam:
1. There are implicit conversions from a set of types to a type used by the client. For example, 

[...]

PR #4234, opened 2019-11-20
Simple non-breaking refactor that makes the API type, see #4196

haf-afa:
See referenced issue #4196


src/Nest/ElasticClient.Indices.cs
```diff
@@ -137,13 +137,14 @@ internal IndicesNamespace(ElasticClient client): base(client)
 		/// <para></para>
 		/// <a href = "https://www.elastic.co/guide/en/elasticsearch/reference/master/indices-create-index.html">https://www.elastic.co/guide/en/elasticsearch/reference/master/indices-create-index.html</a>
 		/// </summary>
-		public CreateIndexResponse Create(IndexName index, Func<CreateIndexDescriptor, ICreateIndexRequest> selector = null) => Create(selector.InvokeOrDefault(new CreateIndexDescriptor(index: index)));
```
haf-afa: As you all can see from the code still typing, the co-domain of `Func` is contravariant; or in other words; it's not until the `Create` call resolves its argument that you have to pass the explicitly implemented interface in. Now, it is possible callers have previously done a Liskov on you to provide their own interface, or your mocks may have, but it's *very* unlikely users have since this is a builder fluent API.

Mpdreamz:
Hi @haf-afa 

I'm keen to do this for our `8.0` release but will follow up with another PR that does it globally in the repos. 

The reason the selectors return an interface was to allow subclasses to be injected. The need for this became less when we introduced and shipped the more classic object initializer syntax.

As of 7.0 because we switched internal serializers subclassing and having properties picked is no longer even supported for any syntax see: https://github.com/elastic/elasticsearch-net/issues/3655

In the long run I am keen to no longer support this, however nice to the end user its hard to write simple generic serializers for the built in types.

So we will very likely change over to selector Func's returning the type they were injected.

PR #4238, opened 2019-11-25
Expose q querystring parameter on search

Mpdreamz:
This ended up on our skip list in `SearchOverrides.cs` in the code generator but other API's already expose it.


Mpdreamz:
This always sends a `POST` even if `q` is specified.

In the case both `q` and `query` in the body is specified the body wins

```json
POST /_search?q=a:1
{
  "size": 0,
  "query" : {
    "query_string": {
      "default_field": "a",
      "query": "2"
    }
  }
}
```

Mpdreamz:
ported to `7.x`

PR #4239, opened 2019-11-25
Implement monthly round of more examples 

Mpdreamz:
Depends on https://github.com/elastic/elasticsearch-net/pull/4238

### DIFF
--- a/src/Examples/Examples/Example.cs
+++ b/src/Examples/Examples/Example.cs
@@ -33,7 +33,7 @@ namespace Examples
 			Body = body.ToString();
 		}
 
-		public Example QueryStringOnBody(string key, object value)
+		public Example MoveQueryStringToBody(string key, object value)
 		{
 			Uri.Query = Uri.Query.Replace($"{key}={value}", string.Empty);
 			ApplyBodyChanges(body =>
--- a/src/Examples/Examples/Example.cs
+++ b/src/Examples/Examples/Example.cs
@@ -14,7 +14,7 @@ namespace Examples
 		private Example(HttpMethod method, Uri uri, string body)
 		{
 			Method = method;
-			RequestUri = new UriBuilder(uri);
+			Uri = new UriBuilder(uri);
 			Body = body;
 		}
 
@@ -22,7 +22,7 @@ namespace Examples
 
 		public HttpMethod Method { get; set; }
 
-		public UriBuilder RequestUri { get; set; }
+		public UriBuilder Uri { get; set; }
 
 		public void ApplyBodyChanges(Action<JObject> action)
 		{
@@ -35,7 +35,7 @@ namespace Examples
 
 		public Example QueryStringOnBody(string key, object value)
 		{
-			RequestUri.Query = RequestUri.Query.Replace($"{key}={value}", string.Empty);
+			Uri.Query = Uri.Query.Replace($"{key}={value}", string.Empty);
 			ApplyBodyChanges(body =>
 			{
 				body[key] = new JValue(value);
@@ -51,7 +51,7 @@ namespace Examples
 			var path = Callout.Replace(urlParts[1], string.Empty);
 			var body = exampleParts.Length > 1 ? exampleParts[1] : null;
 
-			if (!Uri.TryCreate(BaseUri, path, out var uri))
+			if (!System.Uri.TryCreate(BaseUri, path, out var uri))
 				throw new Exception($"Cannot parse Uri from {path}");
 
 			if (body != null)
--- a/src/Examples/Examples/Example.cs
+++ b/src/Examples/Examples/Example.cs
@@ -14,7 +14,7 @@ namespace Examples
 		private Example(HttpMethod method, Uri uri, string body)
 		{
 			Method = method;
-			Uri = uri;
+			RequestUri = new UriBuilder(uri);
 			Body = body;
 		}
 
@@ -22,7 +22,7 @@ namespace Examples
 
 		public HttpMethod Method { get; set; }
 
-		public Uri Uri { get; set; }
+		public UriBuilder RequestUri { get; set; }
 
 		public void ApplyBodyChanges(Action<JObject> action)
 		{
@@ -31,6 +31,16 @@ namespace Examples
 				: JObject.Parse(Body);
 			action(body);
 			Body = body.ToString();
+		}
+
+		public Example QueryStringOnBody(string key, object value)
+		{
+			RequestUri.Query = RequestUri.Query.Replace($"{key}={value}", string.Empty);
+			ApplyBodyChanges(body =>
+			{
+				body[key] = new JValue(value);
+			});
+			return this;
 		}
 
 		public static Example Create(string example)
--- a/src/Examples/Examples/Example.cs
+++ b/src/Examples/Examples/Example.cs
@@ -9,7 +9,7 @@ namespace Examples
 	{
 		private static readonly Uri BaseUri = new Uri("http://localhost:9200");
 
-		private static readonly Regex Callout = new Regex(@"\\<\d+>\s*$", RegexOptions.Multiline);
+		private static readonly Regex Callout = new Regex(@"(?:\\)?<\d+>\s*$", RegexOptions.Multiline);
 
 		private Example(HttpMethod method, Uri uri, string body)
 		{
--- a/src/Examples/Examples/JTokenExtensions.cs
+++ b/src/Examples/Examples/JTokenExtensions.cs
@@ -49,20 +49,15 @@ namespace Examples
 		/// Docs document bool query without the clauses as arrays, NEST only every sends them as arrays
 		/// This fixes the examples
 		/// </summary>
-		public static (JArray must, JArray mustNot, JArray filter, JArray should) FixBoolQuery(this JToken o)
+		public static void ToLongFormBoolQuery(this JToken o, Action<JObject> mutateBool = null)
 		{
-			var empty = new JArray();
-			if (!(o is JObject obj)) return (empty, empty, empty, empty);
-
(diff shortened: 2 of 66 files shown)


src/Examples/Examples/QueryDsl/QueryStringQueryPage.cs
```diff
@@ -1,16 +1,25 @@
...
 {
 	public class QueryStringQueryPage : ExampleBase
 	{
-		[U(Skip = "Example not implemented")]
+		[U]
 		public void Line42()
 		{
 			// tag::ad6ea0c1e46712aa1fd6d3bfa0ec979e[]
-			var response0 = new SearchResponse<object>();
+			var response0 = client.Search<Blog>(s => s
```
russcam: Rename to `searchResponse`

src/Examples/Examples/ResponseExtensions.cs
```diff
@@ -17,6 +17,13 @@ public static class ResponseExtensions
 	{
 		private static readonly JsonSerializer Serializer = new JsonSerializer();
 
+		public static void MatchesExample(this IResponse response, string content, Action<Example, JObject> clientChanges) =>
```
russcam: Really not keen on this! Seems overly convoluted to save typing an extra line

src/Examples/Examples/ResponseExtensions.cs
```diff
@@ -37,22 +44,35 @@ public static void MatchesExample(this IResponse response, string content, Func<
...
+			// So if we expect `/_search` but `/_all/_search` is passed we feel they are equivalent
+			if (expectedPath != "/_search" || decodedAbsolutePath != "/_all/_search")
+				decodedAbsolutePath.Should().Be(expectedPath);
 
 			// check expected query string params. Rather that _all_ keys match,
 			// only check that the ones in reference doc example are present, because
 			// the client may append more key/values such as "typed_keys"
-			var expectedQueryParams = HttpUtility.ParseQueryString(example.Uri.Query);
+			// Because the tests remove keys with string replace the following bad query string can appear
+			example.RequestUri.Query = example.RequestUri.Query.Replace("?&", "?").Replace("&&", "&");
```
russcam: > example.RequestUri.Query = example.RequestUri.Query.Replace("?&", "?").Replace("&&", "&");

Can we handle this in the key/value replacement logic instead?

src/Examples/Examples/ResponseExtensions.cs
```diff
@@ -37,22 +44,35 @@ public static void MatchesExample(this IResponse response, string content, Func<
...
+			var expectedQueryParams = HttpUtility.ParseQueryString(example.RequestUri.Query);
```
russcam: `HttpUtility.ParseQueryString` cannot handle keys with no associated value, so should we handle this logic ourselves by splitting, `Uri.UnescapeDataString()` and creating a `Dictionary<string,string>` (which I think is OK for all params of Elasticsearch, as none a multi value that requires `NameValueCollection`)? It will also simplify the `foreach` loop below

src/Examples/Examples/Root/MappingPage.cs
```diff
@@ -1,18 +1,27 @@
...
 {
 	public class MappingPage : ExampleBase
 	{
-		[U(Skip = "Example not implemented")]
+		[U]
 		public void Line144()
 		{
 			// tag::d8b2a88b5eca99d3691ad3cd40266736[]
-			var response0 = new SearchResponse<object>();
+			var mapResponse = client.Indices.Create("my-index", c => c
```
russcam: rename to `createIndexResponse`

src/Examples/Examples/Root/MappingPage.cs
```diff
@@ -1,18 +1,27 @@
...
+				.Map<Employee>(m => m
+					.Properties(props => props
+						.Scalar(p => p.Age)
```
russcam: Do we want to get rid of `Scalar` in 8.0.0?

src/Examples/Examples/ResponseExtensions.cs
```diff
@@ -37,22 +44,35 @@ public static void MatchesExample(this IResponse response, string content, Func<
 			// the client encodes characters such as commas, so decode to compare
 			var decodedAbsolutePath = HttpUtility.UrlDecode(response.ApiCall.Uri.AbsolutePath);
 
-			decodedAbsolutePath.Should().Be(example.Uri.AbsolutePath.Length > 1
-				? example.Uri.AbsolutePath.TrimEnd('/')
-				: example.Uri.AbsolutePath);
+			var expectedUri = example.RequestUri.Uri;
+			var expectedPath = expectedUri.AbsolutePath.Length > 1 ? expectedUri.AbsolutePath.TrimEnd('/') : expectedUri.AbsolutePath;
+
+			// A lot of the examples are not constrained to an index which is not necessary what users typically do
```
russcam:
```suggestion
			// A lot of the examples are not constrained to an index which is not necessarily what users typically do.
```

src/Examples/Examples/Example.cs
```diff
@@ -9,20 +9,20 @@ public class Example
...
+			RequestUri = new UriBuilder(uri);
 			Body = body;
 		}
 
 		public string Body { get; private set; }
 
 		public HttpMethod Method { get; set; }
 
-		public Uri Uri { get; set; }
+		public UriBuilder RequestUri { get; set; }
```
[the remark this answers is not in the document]
Mpdreamz: Mainly because the static method below flags `Uri` as ambigious not sure if thats a resharper/rider quirk. I can change it back and use `System.Uri` in the static method.

src/Examples/Examples/Example.cs
```diff
@@ -33,6 +33,16 @@ public void ApplyBodyChanges(Action<JObject> action)
...
 
+		public Example QueryStringOnBody(string key, object value)
+		{
+			RequestUri.Query = RequestUri.Query.Replace($"{key}={value}", string.Empty);
+			ApplyBodyChanges(body =>
+			{
+				body[key] = new JValue(value);
+			});
+			return this;
+		}
```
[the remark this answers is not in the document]
Mpdreamz: :+1:

src/Examples/Examples/JTokenExtensions.cs
```diff
@@ -0,0 +1,47 @@
+using System;
+using Newtonsoft.Json.Linq;
+
+namespace Examples
+{
+	public static class JTokenExtensions
+	{
+		public static JToken ReplaceWithValue(this JToken toReplace, Func<JToken, JToken> newToken)
```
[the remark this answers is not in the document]
Mpdreamz: I want to signal in the method name you get the value of the token your replacing

src/Examples/Examples/JTokenExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+		{
+			var n = newToken(toReplace);
+			toReplace.Replace(n);
+			return n;
+		}
+		public static JToken FixSimpleTerm(this JToken termSimple) => termSimple.ReplaceWithValue(v=>new JObject { { "value", v } });
```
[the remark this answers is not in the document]
Mpdreamz: I went with `ToLongFormQuery` and `ToLongFormTermQuery` instead and added xmldocs.

src/Examples/Examples/Models/Account.cs
```diff
@@ -13,5 +13,9 @@ public class Account
 		public double? Balance { get; set; }
 		public string Gender { get; set; }
 		public string State { get; set; }
+		public string User { get; set; }
+
+		[PropertyName("tag")]
```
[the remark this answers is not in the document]
Mpdreamz: Cool cool, the AccountNumber already had `PropertyName` i am happy to normalize the models to `DataMember`

src/Examples/Examples/Root/MappingPage.cs
Mpdreamz: Aye I like the field name first approach of `Scalar` but I think we can do better in terms of extensiblity

src/Examples/Examples/ResponseExtensions.cs
Mpdreamz: Most tests simply do a `string.Replace`, happy to move them all over to something more structured

src/Examples/Examples/JTokenExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+			var a = new JArray(o);
+			if (o == null) return a;
+			o.Replace(a);
+			return a;
+		}
+		/// <summary>
+		/// Docs document bool query without the clauses as arrays, NEST only every sends them as arrays
+		/// This fixes the examples
+		/// </summary>
+		public static (JArray must, JArray mustNot, JArray filter, JArray should) FixBoolQuery(this JToken o)
```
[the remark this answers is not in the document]
Mpdreamz: Renamed to `ToLongFormBoolQuery` only allowed to be called on `["bool"]`.

This reworked method gets the bool as parameter to an action to further mutate the bool instance if needed.

src/Examples/Examples/ResponseExtensions.cs
Mpdreamz: It's not just about saving one line

https://github.com/elastic/elasticsearch-net/pull/4239/commits/19a0c1c1ea3ee86d7b39c34185d1c8ebbca41bfd#diff-8c3e51d6895a9329f7e6602bb0c00754L69-L81

Saves on explicit return (we never return a new Example), and additional indentation, nested lambda's.

src/Examples/Examples/JTokenExtensions.cs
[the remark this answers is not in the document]
Mpdreamz: xmldocs

Mpdreamz:
@russcam merging this in, we can follow up with improvements in the next months round.

PR #4253, opened 2019-12-09
[DOCS] Added older release notice.

debadair:
Explicitly set page_header.html for 6.x so that the default notice can be updated for OOM versions.

Prerequisite for: elastic/docs#1528


russcam:
Thanks @debadair. Is there an asciidoc attribute we can use to include the admonition, instead of HTML, or is the HTML snippet the recommended approach?

debadair:
@russcam The HTML snippet is currently the only way to inject the banner at the top of the page.

PR #4256, opened 2019-12-10
Add MatchBoolPrefix query

### LINKED ISSUE #4247: Match boolean prefix query not implemented
**Describe the feature**:  
Title sums this up.  Implementation for [the following feature](https://www.elastic.co/guide/en/elasticsearch/reference/7.5/query-dsl-match-bool-prefix-query.html) is missing.  
As a workaround, one can use a `MultiMatch` query, specifying `bool_prefix` as the type.  
I'm not sure if there are drawbacks to using `multi_match` query with one field vs using `match_bool_prefix` query.

russcam:
This commit adds the MatchBoolPrefix query.

Closes #4247


russcam:
ported to master

PR #4258, opened 2019-12-10
Implement SLM Execute Retention API

codebrain:
References: https://github.com/elastic/elasticsearch/pull/47405


src/Tests/Tests/XPack/Slm/SlmApiTests.cs
```diff
@@ -102,6 +103,18 @@ public SlmApiTests(XPackCluster cluster, EndpointUsage usage) : base(new Coordin
 						(v, c, r) => c.SnapshotLifecycleManagement.ExecuteSnapshotLifecycleAsync(r)
 					)
 			},
+			{
+				ExecuteRetentionLifecycleStep, u =>
+					u.Calls<ExecuteRetentionSnapshotLifecycleDescriptor, ExecuteRetentionSnapshotLifecycleRequest, IExecuteRetentionSnapshotLifecycleRequest,
+						ExecuteRetentionSnapshotLifecycleResponse>(
+						v => new ExecuteRetentionSnapshotLifecycleRequest(),
+						(v, d) => d,
+						(v, c, f) => c.SnapshotLifecycleManagement.ExecuteRetentionSnapshotLifecycle(f),
```
Mpdreamz: Should this just be `.SnapshotLifecycleManagement.ExcuteRetention()` based on the API name

codebrain: Good call, done.

src/Elasticsearch.Net/Api/RequestParameters/RequestParameters.SnapshotLifecycleManagement.cs
```diff
@@ -36,6 +36,12 @@ public class ExecuteSnapshotLifecycleRequestParameters : RequestParameters<Execu
 		public override HttpMethod DefaultHttpMethod => HttpMethod.PUT;
 	}
 
+	///<summary>Request options for ExecuteRetention <para>https://www.elastic.co/guide/en/elasticsearch/reference/current/slm-api-execute-retention.html</para></summary>
+	public class ExecuteRetentionRequestParameters : RequestParameters<ExecuteRetentionRequestParameters>
```
russcam: Should the type name(s) be `ExecuteSnapshotLifecycleRetention*` ? The shortened method names are fine, because they reflect what is in the spec and are namespaced on `SnapshotLifecycleManagement`, but with the types, these live in the root `Nest` namespace, which feels like we ought to be more explicit to avoid naming collisions now and in future.

codebrain:
Ported to `master`

PR #4263, opened 2019-12-10
Static property for formatter instance returned new instance everytime

### LINKED ISSUE #4260: .NET ElasticLowLevelClient 7.4.1 serializer extremely slow
**NEST/Elasticsearch.Net version**: 7.4.1

**Elasticsearch version**: 7.4.2

I recently switched from ES 6.8.5 to 7.4.2. I noticed that bulk loading became extremely slow. I narrowed down the problem to the .NET's ElasticLowLevelClient serialization.

For example, serializing the following code 100K times in ES 6 takes ~1 second where as in ES 7 takes about 70 seconds (!!). As I have millions of records, this becomes a real problem.

```
var o = es.Serializer.SerializeToString(new
{
    index = new
    {
        _index = indexName,
        _id = session.SESSION_ID
    }
}, SerializationFormatting.None);
```
I also tried using MemoryStreamFactory.Default for the serialization but results are the same.

When the serialized class is bigger, the performance is even worse.

Currently I temporarily (?) switched to Newtonsoft JSON and the results are satisfying.
Can you please advise? Is there anyway to get the previous serializer or its performance?

**Steps to reproduce**:
Run the code below with Elasticsearch.Net 7.4.1 and with Elasticsearch.Net 6.8.3.

With 6.8.3 the results are:
> ES: 75ms

With 7.4.1 the results are:

>ES: 6824ms

```
        {
            var settings = new ConnectionConfiguration(new Uri("http://localhost:9200"));

            var es = new ElasticLowLevelClient(settings);
            var sw = Stopwatch.StartNew();
            for (int i = 0; i < 10000; i++)
            {
                var o = es.Serializer.SerializeToString(new
                {
                    index = new
                    {
                        _index = "test",
                        _id = i
                    }
                }, SerializationFormatting.None);
            }
            sw.Stop();

            Console.WriteLine($"ES: {sw.ElapsedMilliseconds}ms");
            Console.ReadKey();
        }
```

Mpdreamz:
fixes #4260

### DIFF
--- a/src/Elasticsearch.Net/Serialization/Resolvers/ElasticsearchNetFormatterResolver.cs
+++ b/src/Elasticsearch.Net/Serialization/Resolvers/ElasticsearchNetFormatterResolver.cs
@@ -18,7 +18,7 @@ namespace Elasticsearch.Net
 			_fallbackFormatter = new DynamicObjectTypeFallbackFormatter(_innerFormatterResolver);
 		}
 
-		public static ElasticsearchNetFormatterResolver Instance => new ElasticsearchNetFormatterResolver();
+		public static ElasticsearchNetFormatterResolver Instance { get; } = new ElasticsearchNetFormatterResolver();
 
 		public IJsonFormatter<T> GetFormatter<T>() =>
 			typeof(T) == typeof(object)


Mpdreamz:
backported to `7.4` as well

PR #4265, opened 2019-12-11
Ensure missing SLM policy is handled by client correctly

codebrain:
Integration test to ensure correct handling of: https://github.com/elastic/elasticsearch/pull/47679


codebrain:
Ported to `master`

PR #4268, opened 2019-12-13
Support MemberExpression and specific MethodCallExpression

### LINKED ISSUE #4240: Nest Field no longer support member expressions
7.4
**Elasticsearch version**:
7.4
Hi there!

I have created a topic at the Elastic forums which explains this issue, it can be found here: [Nest Field no longer support MemberExpressions](https://discuss.elastic.co/t/nest-field-no-longer-support-memberexpressions/208656). I was redirected here.

*Expected:*
Fields to be able to take a member expression in the constructor, as this documentation states:
[Member expression only](https://www.elastic.co/guide/en/elasticsearch/client/net-api/current/field-inference.html#member-expressions-only)

*Actual:*
Throws an exception that tells the user the provided expression is not a lambda expression.

I think this commit have changed the behavior: [7x/field resolver performance #3668](https://github.com/elastic/elasticsearch-net/commit/0fb1149c1f84a5e501149c78b222b5ae7dec677c#diff-60957b99fed84039ef5daa9b1861db9f)

We have a "workaround" for this now where we convert the member expression into a lambda expression before passing it to the Field constructor, but I think either the documentation should be updated to match the actual behavior or the constructor should support member expressions.

**Steps to reproduce**:
 1. Create a MemberExpression.
 2. Create a new Field and pass the MemberExpression into the constructor.

russcam:
This commit adds support for MemberExpression and specific MethodCallExpression
as Expressions that can be passed to Field and PropertyName, in order to resolve
a string value from the expression.

The specific MethodCallExpression support is to allow F# quotations enclosing
Lambda expressions to be supported, when converted using

Microsoft.FSharp.Linq.RuntimeHelpers.LeafExpressionConverter.QuotationToExpression

Fixes #4240

### DIFF
--- a/src/Nest/CommonAbstractions/Extensions/ExpressionExtensions.cs
+++ b/src/Nest/CommonAbstractions/Extensions/ExpressionExtensions.cs
@@ -30,9 +30,41 @@ namespace Nest
 
 			if (expression == null) return null;
 
-			if (!(expression is LambdaExpression lambda)) throw new Exception($"Not a lambda expression: {expression}");
+			switch (expression)
+			{
+				case LambdaExpression lambdaExpression:
+					type = lambdaExpression.Parameters.FirstOrDefault()?.Type;
+					break;
+				case MemberExpression memberExpression:
+					type = memberExpression.Member.DeclaringType;
+					break;
+				case MethodCallExpression methodCallExpression:
+					// special case F# method call expressions on FuncConvert
+					// that are used to convert F# quotations representing lambda expressions, to expressions.
+					// https://github.com/dotnet/fsharp/blob/7adaacf150dd79f072efe42d43168c9cd6edbced/src/fsharp/FSharp.Core/Linq.fs#L796
+					//
+					// For example:
+					//
+					// type Doc = { Message: string; State: string }
+					// let field (f:Expr<'a -> 'b>) =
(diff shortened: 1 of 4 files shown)


src/Nest/CommonAbstractions/Extensions/ExpressionExtensions.cs
```diff
@@ -30,9 +30,41 @@ internal static object ComparisonValueFromExpression(this Expression expression,
...
+			switch (expression)
+			{
+				case LambdaExpression lambdaExpression:
+					type = lambdaExpression.Parameters.FirstOrDefault()?.Type;
+					break;
+				case MemberExpression memberExpression:
+					type = memberExpression.Member.DeclaringType;
+					break;
+				case MethodCallExpression methodCallExpression:
+					// special case F# method call expressions on FuncConvert
```
Mpdreamz: ❤️

russcam:
ported to master

src/Nest/CommonAbstractions/Infer/Field/Field.cs
```diff
@@ -56,6 +56,7 @@ public Field(PropertyInfo property, double? boost = null, string format = null)
 		/// </remarks>
 		public string Format { get; set; }
 
+		// TODO: Rename to CacheableExpression in 8.0.0
```
niemyjski: Why not introduce it now, put an obsolete attribute on this and have the getter and setter populate the new property?

PR #4269, opened 2019-12-13
Set Accept header when format set on Cat APIs

### LINKED ISSUE #4243: Using the format=json uri parameter with cat api calls fails
**NEST/Elasticsearch.Net version**: 7.2.1

**Elasticsearch version**: 7.2.1

**Description of the problem including expected versus actual behavior**: Using the Cat.MasterAsync<StringResponse> API and specifying the Format="JSON" parameter returns a response object that has the Success flag set to false but has a status code of 200 and the correct response body. Calling the same method in Elasticsearch.Net 6.8 worked as expected.

I am calling the API like this (this.client is of type IElasticLowLevelClient):
```
await this.client.Cat.MasterAsync<StringResponse>(new CatMasterRequestParameters { Format = "JSON" });
```

Using a SingleNodeConnectionPool
Specifying DisableAutomaticProxyDetection and DisablePing

```
 - [1] BadResponse: Node: http://localhost:9200/ Took: 00:00:00.0312765
# OriginalException: Elasticsearch.Net.ElasticsearchClientException: Request failed to execute. Call: Status code 200 from: GET /_cat/master?format=json
# Request:
<Request stream not captured or already read to completion by serializer. Set DisableDirectStreaming() on ConnectionSettings to force it to be set on the response.>
# Response:
[{\"id\":\"nodeId\",\"host\":\"hostName\",\"ip\":\"ipaddress\",\"node\":\"nodeName\"}]
```

russcam:
This commit updates the ApiGenerator to set the Accept HTTP header when the format
header is specified. To determine whether an API call is successful, the Content-Type header
of the response is checked against the Accept header of the request.

The format parameter only really makes sense for the low level client and associated
ElasticsearchResponse<T> types, as the high level client always expects to deserialize
from JSON; any other format will result in an exception at deserialization.

Fixes #4243

### DIFF
--- a/src/CodeGeneration/ApiGenerator/Generator/CodeGenerator.cs
+++ b/src/CodeGeneration/ApiGenerator/Generator/CodeGenerator.cs
@@ -9,16 +9,32 @@ namespace ApiGenerator.Generator
 	//TODO this should be in views and models
 	public static class CodeGenerator
 	{
+		public static string CatFormatPropertyGenerator(string type, string name, string key, string setter) =>
+			  $"public {type} {name} {{ "
+			+ $"	get => Q<{type}>(\"{key}\");"
+			+ $"	set {{ "
+			+ $"		Q(\"{key}\", {setter});"
+			+ $"		if (RequestConfiguration == null) "
+			+ $"			RequestConfiguration = new RequestConfiguration();"
+			+ $""
+			+ $"		RequestConfiguration.Accept = AcceptHeaderFromFormat({setter});"
+			+ $"	}}"
+			+ $"}}";
+
 		public static string PropertyGenerator(string type, string name, string key, string setter) =>
 			$"public {type} {name} {{ get => Q<{type}>(\"{key}\"); set => Q(\"{key}\", {setter}); }}";
 
-		public static string Property(string type, string name, string key, string setter, string obsolete, params string[] doc)
+		public static string Property(string @namespace, string type, string name, string key, string setter, string obsolete, params string[] doc)
 		{
 			var components = new List<string>();
 			foreach (var d in RenderDocumentation(doc)) A(d);
 			if (!string.IsNullOrWhiteSpace(obsolete)) A($"[Obsolete(\"Scheduled to be removed in 7.0, {obsolete}\")]");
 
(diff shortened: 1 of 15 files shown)


src/CodeGeneration/ApiGenerator/Generator/CodeGenerator.cs
```diff
@@ -9,16 +9,32 @@ namespace ApiGenerator.Generator
...
 	{
+		public static string CatFormatPropertyGenerator(string type, string name, string key, string setter) =>
+			  $"public {type} {name} {{ "
+			+ $"	get => Q<{type}>(\"{key}\");"
+			+ $"	set {{ "
+			+ $"		Q(\"{key}\", {setter});"
+			+ $"		if (RequestConfiguration == null) "
+			+ $"			RequestConfiguration = new RequestConfiguration();"
+			+ $""
+			+ $"		RequestConfiguration.Accept = AcceptHeaderFromFormat({setter});"
```
Mpdreamz: Can we hide this behind a single lined method?

src/Tests/Tests.Reproduce/GithubIssue4243.cs
```diff
@@ -0,0 +1,36 @@
...
+using Tests.Core.ManagedElasticsearch.Clusters;
+using Tests.Domain;
+
+namespace Tests.Reproduce
+{
+	public class GithubIssue4243 : IClusterFixture<ReadOnlyCluster>
+	{
+		private readonly ReadOnlyCluster _cluster;
+
+		// use intrusive operation cluster because we're changing the underlying http handler
```
Mpdreamz: This comment is not relevant.

russcam: Thanks, will fix

src/CodeGeneration/ApiGenerator/Generator/CodeGenerator.cs
russcam: I've moved this into its own method

russcam:
I've addressed the PR comments, if you wouldn't mind having another look @Mpdreamz

Mpdreamz:
ported to `master`

PR #4270, opened 2019-12-13
Get highest AssemblyFileVersionAttribute version

### LINKED ISSUE #4189: Unable to initialize ConnectionConfiguration in certain conditions
**NEST/Elasticsearch.Net 7.0+**:

**Elasticsearch any**:

**Indirect initialization of ConnectionConfiguration throws AmbiguousMatchException**:
When an instance of ConnectionConfiguration is being initialized the static constructor is invoked indirectly by accessing to read-only static property to hold default user agent. By call stack it crash when accessing to RuntimeInformation.FrameworkDescription

```
[AmbiguousMatchException: Multiple custom attributes of the same type found.]
  System.Attribute.GetCustomAttribute(Assembly element, Type attributeType, Boolean inherit) +119
  Elasticsearch.Net.RuntimeInformation.get_FrameworkDescription() +96
  Elasticsearch.Net.ConnectionConfiguration..cctor() +402
[TypeInitializationException: The type initializer for 'Elasticsearch.Net.ConnectionConfiguration' threw an exception.]
```

We was able to localize the bug on RuntimeInformation
```
		public static string FrameworkDescription
		{
			get
			{
				if (_frameworkDescription == null)
				{
					var assemblyFileVersionAttribute =
						(AssemblyFileVersionAttribute)typeof(object).GetTypeInfo().Assembly.GetCustomAttribute(typeof(AssemblyFileVersionAttribute));
					_frameworkDescription = $".NET Framework {assemblyFileVersionAttribute.Version}";
				}
				return _frameworkDescription;
			}
		}
```

In unknown conditions depends on .NET Framework returns more than one attribute AssemblyFileVersionAttribute and Assembly.GetCustomAttribute throws an exception.

According to our investigation both attributes come from the same module but has different version of the file

System.Reflection.AssemblyFileVersionAttribute	file:///C:/Windows/Microsoft.NET/Framework64/v4.0.30319/mscorlib.dll	4.0.30319
System.Reflection.AssemblyFileVersionAttribute	file:///C:/Windows/Microsoft.NET/Framework64/v4.0.30319/mscorlib.dll	2.0.0.205

The first one always has .NET Framework version but the second one indicates version of something unknown for us 2.0.0.205

Issue was introduced in 7.x according to the sources
https://github.com/elastic/elasticsearch-net/blob/7.x/src/Elasticsearch.Net/Api/RuntimeInformation.cs

**Steps to reproduce**:
 1. Initialize Elasticsearch client on web application
 
It can be reproduced only on some Windows Server and relevant only to web application.
On the same machine console application behaves as expected and .NET Framework objects has single attribute, that strange.

Our research was done on Windows Server 2012 & Windows Server 2012 R2 regardless of which build of .NET 4.7.2 is used. 

We continue investigation to isolate it and find out which settings or conditions lead to have more than one attribute for system objects. It's not new issue with .NET

**Add User-Agent string to API calls to Elasticsearch #3784**:
With PR 
https://github.com/elastic/elasticsearch-net/pull/3784

As a suggested fix it's to read all attributes and filter them instead of attempt to read a single.

russcam:
This commit updates the FrameworkDescription logic used to determine
the .NET framework version from the AssemblyFileVersionAttribute on the
assembly containing the base object type to

1. Get all AssemblyFileVersionAttribute attributes
2. order by versioning descending
3. get the first version

This is to workaround an edge case in using this technique to get the
.NET Framework version that appears to affect running an application
under full framework in IIS.

Fixes #4189.

### DIFF
--- a/src/Elasticsearch.Net/Api/RuntimeInformation.cs
+++ b/src/Elasticsearch.Net/Api/RuntimeInformation.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 #if NET461
 using System.Reflection;
 
@@ -16,7 +17,11 @@ namespace Elasticsearch.Net
 				if (_frameworkDescription == null)
 				{
 					var assemblyFileVersionAttribute =
-						(AssemblyFileVersionAttribute)typeof(object).GetTypeInfo().Assembly.GetCustomAttribute(typeof(AssemblyFileVersionAttribute));
+						((AssemblyFileVersionAttribute[])Attribute.GetCustomAttributes(
+							typeof(object).GetTypeInfo().Assembly,
+							typeof(AssemblyFileVersionAttribute)))
+						.OrderByDescending(a => a.Version)
+						.First();
 					_frameworkDescription = $".NET Framework {assemblyFileVersionAttribute.Version}";
 				}
 				return _frameworkDescription;
--- a/src/Nest/CrossPlatform/RuntimeInformation.cs
+++ b/src/Nest/CrossPlatform/RuntimeInformation.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 #if NET461
 using System.Reflection;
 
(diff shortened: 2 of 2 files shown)


russcam:
ported to master

PR #4272, opened 2019-12-16
Add support to retrieve all API keys if user has privilege.

codebrain:
Test to assert server returns a response when no parameters are specified.

Implements https://github.com/elastic/elasticsearch/pull/47274


src/Tests/Tests/XPack/Security/ApiKey/SecurityApiKeyTests.cs
```diff
@@ -120,6 +121,25 @@ public SecurityApiKeyTests(XPackCluster cluster, EndpointUsage usage) : base(new
 						(v, c, r) => c.Security.CreateApiKeyAsync(r)
 					)
 			},
+			{
```
Mpdreamz: This needs to be denoted with a version somehow (maybe an additional parameter taking a version range string?)

Mpdreamz:
Failing tests relate to #4280  and `ProcessClusterEventTimeoutException`

codebrain:
Nice addition @Mpdreamz - merging now.

codebrain:
Ported to `master`

PR #4278, opened 2019-12-17
DateMath has ToString() value for DateTime.MinValue

### LINKED ISSUE #4228: DateTime.MinValue doesn't get parsed for dateRangeQuery
7.4.0
**Elasticsearch version**:
7.3.0
In the search descriptor `DateRange.GreaterThanOrEquals` does not parse `DateTime.MinValue`, instead it sends a request with an empty string which ES 7.4 throws an error `cannot parse empty date`
**Steps to reproduce**:
 1. Set DateTime.MinValue
 2. Create any search query

russcam:
This commit updates the ToString() implementation of DateMath to emit a value
when DateTime.MinValue is passed.

Update IsValid property to allow any DateTime to be valid.

Fixes #4228

### DIFF
--- a/src/Nest/CommonOptions/DateMath/DateMath.cs
+++ b/src/Nest/CommonOptions/DateMath/DateMath.cs
@@ -83,7 +83,7 @@ namespace Nest
 
 		internal static bool IsValidDateMathString(string dateMath) => dateMath != null && DateMathRegex.IsMatch(dateMath);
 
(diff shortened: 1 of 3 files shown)


Mpdreamz:
Opened https://github.com/elastic/elasticsearch-net/issues/4280 for the unrelated failing test

Mpdreamz:
ported to `master`

PR #4291, opened 2019-12-19
Replace old github organisation (elasticsearch) urls with new organisation (elastic).

codebrain:
Not sure if this needs porting as there are only two code mentions.


codebrain:
Ported to `7.x`

PR #4302, opened 2020-01-03
Documentation generation improvements 

Mpdreamz:
* Generate to temp location `docs-temp` first
* This folder is gitignored
* Only if generation is succesful swap `docs-temp` to `docs` this will
make `git status` a whole less noisy on generation failures.
* Make sure IncludeFiles references the new `tests/Tests` location
* `breaking-changes` folders are still preserved
* `Litup.Go` should return an exit code
* Call `build.sh documentation` to generate only the docs
* Improve [resilience of calling the documentation generator](https://github.com/elastic/elasticsearch-net/pull/4302#discussion_r362972586) from the command line
* Add azure devops job that fails if a PR has uncommitted documentation changes, which `master` has so [it currently fails on this PR](https://github.com/elastic/elasticsearch-net/pull/4302/checks?check_run_id=373092851). Will follow up with a new PR that solely updates the docs.

@codebrain @russcam tagging you as reviewers while you are out. Merging this in but please have a gander.

### DIFF
--- a/build/scripts/Documentation.fs
+++ b/build/scripts/Documentation.fs
@@ -7,6 +7,7 @@ open Fake.Core
 
 module Documentation = 
 
+    exception DocGenError of string 
     let Generate args = 
         let path = Paths.InplaceBuildOutput "DocGenerator" "netcoreapp3.0"
         let generator = sprintf "%s.dll" "DocGenerator"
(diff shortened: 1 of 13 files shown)


build/scripts/Documentation.fs
```diff
@@ -24,7 +25,33 @@ module Documentation =
...
+            printfn "Attempt %i to generate the docs" (System.Math.Abs(times - 5) + 1)
+            let result = Tooling.DotNet.ReadQuietIn path dotnetArgs
+            if result.ExitCode = 0 || times = 1 || File.Exists(Paths.MagicDocumentationFile) then
+                result.Output |> Seq.iter (fun l -> if l.Error then printfn "%s" l.Line else eprintfn "%s" l.Line)
+                if result.ExitCode <> 0 then
+                    eprintfn "documentation generation failed"
+                    raise (DocGenError("documentation generation failed"))
+            else
+                eprintfn "documentation generation failed"
+                raise (DocGenError("documentation generation failed"))
```
Mpdreamz: @codebrain @russcam drawing attention to this.

Noticed that on a clean checkout or after `git clean -fxd` the doc generator spits out vague compiler errors.

First about unsupported tokens (c# 8?) then about friendly assembly access and on third invocation everything works and continues to work. I traced it down to `Elasticsearch.Net.csprojAssemblyReference.cache` appearing in `obj` after a few `dotnet run` invocations but can not find much online about what this file is or does or how I can force it to be generated. Hence the brute force approach here.

PR #4305, opened 2020-01-03
update doc_current and doc_branch when calling release

Mpdreamz:
Calling `build release <version>` would clear `doc_branch` in `global.json` causing all doc files to updated when calling the documentation generator.


codebrain:
LGTM

PR #4309, opened 2020-01-06
Add run-repository powershell script

russcam:
Relates: #4301

This commit adds a PowerShell script equivalents to the CI bash scripts


Mpdreamz:
`StaleDocs` failed here because BuildAlazyer returned `Could not find build environment` which happens if it can't locate SDK bits, this happens on some checkouts. Will keep an eye out on this.

PR #4318, opened 2020-01-17
Reference an older Newtonsoft.Json package

horvatal:
Reference an older Newtonsoft.Json package, because some clients depend on lower versions.


russcam:
@horvatal I can't see us accepting this pull request; Json.NET 12.x came out in November 2018 and we tend to keep the client updated with the latest major for supported client versions. Plus, there will undoubtedly be users on 12.x that would want the dependency to be updated to 12.x if 11.x were referenced 😃 

I'm going to close this.

PR #4322, opened 2020-01-23
Handle StartObject or StartArray when reading QueryContainer collection

### LINKED ISSUE #4311: GetIndex() Unable to Deserialize Response for Certain Alias Filters
6.8.3
**Elasticsearch version**:
6.8.3

I'm having an issue with NEST loading index using `await client.GetIndexAsync(indexName)` method if the index has alias setup with certain filter format. Cluster accepts the filter, but NEST client fails.

Alias with filter:
```json
{ "actions": [
	{
            "add": {
                "index": ".es-test-repro",
                "alias": ".es-test-repro-alias1",
                "filter": {
                    "bool": { "must_not": { "exists": { "field": "field1" } } }
                }
            }
	} ]
}
```

Error:
```
Cannot deserialize the current JSON object (e.g. {"name":"value"}) into type 'System.Collections.Generic.IEnumerable`1[Nest.QueryContainer]' because the type requires a JSON array (e.g. [1,2,3]) to deserialize correctly.
To fix this error either change the JSON to a JSON array (e.g. [1,2,3]) or change the deserialized type so that it is a normal .NET type (e.g. not a primitive type like integer, not a collection type like an array or List<T>) that can be deserialized from a JSON object. JsonObjectAttribute can also be added to the type to force it to deserialize from a JSON object.
Path '['.es-test-repro'].aliases['.es-test-repro-alias1'].filter.bool.must_not.exists', line 1, position 90.
Cannot deserialize the current JSON object (e.g. {"name":"value"}) into type 'System.Collections.Generic.IEnumerable`1[Nest.QueryContainer]' because the type requires a JSON array (e.g. [1,2,3]) to deserialize correctly.
To fix this error either change the JSON to a JSON array (e.g. [1,2,3]) or change the deserialized type so that it is a normal .NET type (e.g. not a primitive type like integer, not a collection type like an array or List<T>) that can be deserialized from a JSON object. JsonObjectAttribute can also be added to the type to force it to deserialize from a JSON object.
Path '['.es-test-repro'].aliases['.es-test-repro-alias1'].filter.bool.must_not.exists', line 1, position 90.
```

Repro project: https://github.com/opnarius/nest-alias-filter-repro

I can get around the error by reformatting the filter as follow:
```
 "filter": {
  "bool": 
    { 
      "must_not": [ { "exists": { "field": "field1" } }  ]
     } 
}
```

Should Nest support this alias filter if cluster is allowing it?

russcam:
This commit implements `ReadJson()` for `QueryContainerCollectionJsonConverter`, to allow
handling a single query within a JSON object, or multiple queries within a JSON array.

Fixes #4311


russcam:
Failures are related to audit trail tests and not this change. Merging in.

PR #4323, opened 2020-01-23
Handle short form queries

### LINKED ISSUE #2855: Support deserialization of query short form
Relates #2836 

Deserialization in NEST should handle the short form of queries such that it is able to correctly deserialize queries that have been constructed outside of NEST.

russcam:
This commit handles the short form of the following queries:

- term query
- match query
- prefix query
- match phrase query
- match phrase prefix query

Closes #2855

### DIFF
--- a/src/Nest/QueryDsl/Abstractions/FieldName/FieldNameQueryJsonConverter.cs
+++ b/src/Nest/QueryDsl/Abstractions/FieldName/FieldNameQueryJsonConverter.cs
@@ -8,14 +8,39 @@ namespace Nest
 	{
 		protected override object DeserializeJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
 		{
-			//{
 			reader.Read(); //property name
-			var fieldName = reader.Value as string;
-			reader.Read(); //{
-			var query = ReadAs(reader, objectType, existingValue, serializer);
+			var fieldName = (string)reader.Value;
+			reader.Read();
+			TReadAs query = null;
+
+			switch (reader.TokenType)
+			{
+				case JsonToken.StartObject:
+					query = ReadAs(reader, objectType, existingValue, serializer);
+					break;
+				case JsonToken.Null:
+					break;
+				default:
+					query = new TReadAs();
+					switch (query)
+					{
+						case ITermQuery termQuery:
+							termQuery.Value = reader.Value;
+							break;
+						case IMatchQuery matchQuery:
+							matchQuery.Query = (string)reader.Value;
+							break;
+						case IMatchPhraseQuery matchPhraseQuery:
+							matchPhraseQuery.Query = (string)reader.Value;
+							break;
+						case IMatchPhrasePrefixQuery matchPhrasePrefixQuery:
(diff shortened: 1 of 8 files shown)


russcam:
This should also be ported to 7.x and master. Introducing an interface for full-text queries with a `Query` property may be a good idea.

src/Tests/Tests/QueryDsl/TermLevel/Prefix/PrefixQueryUsageTests.cs
```diff
@@ -47,5 +52,16 @@ public class PrefixQueryUsageTests : QueryDslUsageTestsBase
 				.Value("proj")
 				.Rewrite(MultiTermQueryRewrite.TopTerms(10))
 			);
+
+		//hide
+		[U] public void DeserializeShortForm()
+		{
+			var json = JsonConvert.SerializeObject(new { description = "project description" });
```
Mpdreamz: Can we move this to `QueryDslUsageTestsBase` and reuse in the other `DeserializeShortForm` ?

src/Nest/QueryDsl/Abstractions/FieldName/FieldNameQueryJsonConverter.cs
```diff
@@ -8,14 +8,39 @@ internal class FieldNameQueryJsonConverter<TReadAs> : ReserializeJsonConverter<T
...
+			{
+				case JsonToken.StartObject:
+					query = ReadAs(reader, objectType, existingValue, serializer);
+					break;
+				case JsonToken.Null:
+					break;
+				default:
+					query = new TReadAs();
+					switch (query)
+					{
```
Mpdreamz: We need to cover all of these: 

https://github.com/elastic/clients-flight-recorder/blob/master/script/lib/normalizer.js#L430-L434

russcam:
```json
  const fieldQueries = [
    'range', 'common', 'intervals', 'match', 'match_bool_prefix', 'match_phrase', 'match_phrase_prefix',
    'geo_bounding_box', 'geo_distance', 'geo_polygon', 'geo_shape', 'shape', 'distance_feature', 'rank_feature',
    'term', 'terms'
  ]
```

I don't think the other cases are needed. It looks like 'range', 'common', 'intervals', 'geo_bounding_box', 'geo_distance', 'geo_polygon', 'geo_shape', 'shape', 'distance_feature', 'rank_feature' and 'terms' query do not support a query short form. The `switch` statement on type of query only applies when the value assigned to the field name key is not a JSON object.

src/Tests/Tests/QueryDsl/TermLevel/Prefix/PrefixQueryUsageTests.cs
russcam: addressed in 306aff9e3bc5418aad7f058ab1e9a1469530b9e0

russcam:
Would you mind taking another look please, @Mpdreamz ?

russcam:
Spoke with @Mpdreamz about this offline, so merging in.

Issue #4280, opened 2019-12-17
[tests] Security Index unavailable 


Mpdreamz:
Sometimes the security index has unavailable shards

Raised this before: https://github.com/elastic/elasticsearch/issues/45250

Would like to understand if we're doing something particularly destructive in our integration tests this happens

```
[2019-12-17T08:49:27,039][INFO ][o.e.x.w.WatcherService   ] [xpack-node-d9e48e9200] reloading watcher, reason [new local watcher shard allocation ids], cancelled [0] queued tasks
  X Tests.XPack.Enrich.ExecutePolicy.ExecutePolicyApiTests.ReturnsExpectedStatusCode [3s 159ms]
  Error Message:
   System.Exception : Failure setting up integration test: Invalid NEST response built from a unsuccessful (503) low level call on PUT: /_enrich/policy/nest-fluentasync-c48b1eeb?pretty=true&error_trace=true
# Audit trail of this API call:
 - [1] BadResponse: Node: https://localhost:9200/ Took: 00:00:00.0159654
# OriginalException: Elasticsearch.Net.ElasticsearchClientException: The remote server returned an error: (503) Server Unavailable.. Call: Status code 503 from: PUT /_enrich/policy/nest-fluentasync-c48b1eeb?pretty=true&error_trace=true. ServerError: Type: unavailable_shards_exception Reason: "at least one primary shard for the index [.security-7] is unavailable" ---> System.Net.WebException: The remote server returned an error: (503) Server Unavailable.
   at System.Net.HttpWebRequest.GetResponse() 
   at Elasticsearch.Net.HttpWebRequestConnection.Request[TResponse](RequestData requestData) in D:\a\1\s\src\Elasticsearch.Net\Connection\HttpWebRequestConnection.cs:line 59
   --- End of inner exception stack trace ---
# Request:
{"match":{"indices":["project"],"match_field":"name","enrich_fields":["description","tags"]}}
# Response:
{
  "error" : {
    "root_cause" : [
      {
        "type" : "unavailable_shards_exception",
        "reason" : "at least one primary shard for the index [.security-7] is unavailable",
[...]

PR #4324, opened 2020-01-24
Handle short form queries

### LINKED ISSUE #2855: Support deserialization of query short form

russcam:
This commit handles the short form of the following queries:

- term query
- match query
- prefix query
- match phrase query
- match phrase prefix query
- match bool prefix query

Closes #2855

### DIFF
--- a/src/Nest/QueryDsl/Abstractions/FieldName/FieldNameQueryFormatter.cs
+++ b/src/Nest/QueryDsl/Abstractions/FieldName/FieldNameQueryFormatter.cs
@@ -1,4 +1,6 @@
-﻿using Elasticsearch.Net.Utf8Json;
+﻿using Elasticsearch.Net.Extensions;
+using Elasticsearch.Net.Utf8Json;
+using Elasticsearch.Net.Utf8Json.Internal;
 
 namespace Nest
 {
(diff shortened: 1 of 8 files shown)


src/Nest/QueryDsl/Abstractions/FieldName/FieldNameQueryFormatter.cs
```diff
@@ -34,18 +36,66 @@ public override void Serialize(ref JsonWriter writer, TInterface value, IJsonFor
...
+										termQuery.Value = NumberConverter.ReadDouble(segment.Array, segment.Offset, out var count);
+									break;
+								case JsonToken.True:
+								case JsonToken.False:
+									termQuery.Value = reader.ReadBoolean();
+									break;
+							}
+							reader.ReadIsEndObjectWithVerify();
+							break;
+						case IMatchQuery matchQuery:
```
Mpdreamz: We need more cases while we are it:

https://github.com/elastic/clients-flight-recorder/blob/master/script/lib/normalizer.js#L430-L434

tests/Tests/QueryDsl/FullText/Match/MatchUsageTests.cs
```diff
@@ -69,5 +74,16 @@ public class MatchUsageTests : QueryDslUsageTestsBase
 				.Name("named_query")
 				.AutoGenerateSynonymsPhraseQuery(false)
 			);
+
+		//hide
+		[U] public void DeserializeShortForm()
+		{
+			var json = JsonConvert.SerializeObject(new { description = "project description" });
```
Mpdreamz: Move to base and reuse for other DeserializeShortForm tests

src/Nest/QueryDsl/Abstractions/FieldName/FieldNameQueryFormatter.cs
russcam:
```json
  const fieldQueries = [
    'range', 'common', 'intervals', 'match', 'match_bool_prefix', 'match_phrase', 'match_phrase_prefix',
    'geo_bounding_box', 'geo_distance', 'geo_polygon', 'geo_shape', 'shape', 'distance_feature', 'rank_feature',
    'term', 'terms'
  ]
```

I don't think the other cases are needed. It looks like 'range', 'common', 'intervals', 'geo_bounding_box', 'geo_distance', 'geo_polygon', 'geo_shape', 'shape', 'distance_feature', 'rank_feature' and 'terms' query do not support a query short form. The `switch` statement on type of query only applies when the value assigned to the field name key is not a JSON object.

tests/Tests/QueryDsl/FullText/Match/MatchUsageTests.cs
russcam: Addressed in 5445bc9dc1ba2bbc093f43e3c982ebaaecf866ac

russcam:
Would you mind taking another look please, @Mpdreamz ?

russcam:
Spoke with @Mpdreamz about this offline, merging in

russcam:
Failing tests don't look related to this change, and #4280 is open to investigate

russcam:
ported to master in 5d7df16fa2bf22214ec331dbcf559e8d53dc8060

PR #4325, opened 2020-01-24
Add index setting method for index.blocks.read_only_allow_delete

russcam:
Relates: https://stackoverflow.com/questions/59882293/how-to-set-settings-in-elastic-search-nest-7-x/59889463#59889463

This PR adds a property/method on DynamicIndexSettings and the
respective descriptor to be able to set the index.blocks.read_only_allow_delete
setting.


russcam:
ported to master

PR #4343, opened 2020-02-03
Fix typo in code that cause the docs stale check to trigger

Mpdreamz:

PR #4344, opened 2020-02-03
Fix typo in code that cause the docs stale check to trigger (#4343)

Mpdreamz:
* Fix typo in code that cause the docs stale check to trigger

* GenerateDocumentationFile evaluated and read before

build.targets evaluates we there for need to manually set it on the
projects with intend to publish.

Make sure the doc generator reads this file rather then always forcing
its own xmldocs emit.

(cherry picked from commit 0e739bc119375604d69f72ef573421543c1f1c5a)


Mpdreamz:
`StaleDocs` is happy, merging with the `Self Reviewed` label

PR #4336, opened 2020-01-30
Feature/master/runner

Mpdreamz:
Can now run the `opensource` and `xpack` test suites fully.

- Support catching of do's
- Support stashes in the form of ${var}
- Support feature `headers`
- Support proper skip version ranges
- Test/Teardown now run between test sections
- Per test suite setup
- Support `warnings`
- Introduce `NotSkipped` Execution result to indicate `skip` should continue
- Better proxy detection (only works for for oss on linux)
- TestExported now skips setup/teardown sections
- Summary reports on errors properly
- Support `ignore` parameters
- Make sure export has both DebugInformation and StackTrace info.

PR #4346, opened 2020-02-03
Allow the code generator to generate only the low level client

Mpdreamz:
Isolated from: #4336

This allows us to start generating the low level client ahead of time and automate this in the future.


Mpdreamz:
Please merge https://github.com/elastic/elasticsearch-net/pull/4347 first and use rebase and commit so that the commit from https://github.com/elastic/elasticsearch-net/pull/4347 stays isolated

src/ApiGenerator/Configuration/GeneratorLocations.cs
```diff
@@ -27,15 +27,15 @@ public static string Root
 
 				var directoryInfo = new DirectoryInfo(Directory.GetCurrentDirectory());
 
-				var runningAsDnx =
+				var dotnetRun =
```
russcam: better name for it now 😃

src/ApiGenerator/Program.cs
```diff
@@ -40,18 +32,29 @@ private static async Task Main(string[] args)
...
+
+		private static bool Ask(string question, bool defaultAnswer = true)
+		{
+			var answer = "invalid";
+			var defaultResponse = defaultAnswer ? "y" : "n";
+
 			while (answer != "y" && answer != "n" && answer != "")
 			{
-				Console.Write("Generate code from the specification files on disk? [Y/n] (default Y): ");
+				Console.Write($"{question}[y/N] (default {defaultResponse}): ");
```
russcam: nitpick: The default answer in `[y/n]` should be the capitalized one

Mpdreamz: Good call out will take this on board when introducing command line arguments 👍

PR #4347, opened 2020-02-03
Update low level client with missing methods

Mpdreamz:
Continuation of: #4346 

Isolated from: #4336

This updates the low level client and introduces all the API's we haven't mapped in the high level client yet.


src/Elasticsearch.Net/ElasticLowLevelClient.NoNamespace.cs
```diff
@@ -162,6 +170,12 @@ public LowLevelTasksNamespace Tasks
 			private set;
 		}
 
+		public LowLevelTransformNamespace Transform
```
russcam: Is there some overlap in data_frames and transform APIs?

src/Elasticsearch.Net/ElasticLowLevelClient.NoNamespace.cs
```diff
@@ -72,6 +74,12 @@ public LowLevelCrossClusterReplicationNamespace CrossClusterReplication
 			private set;
 		}
 
+		public LowLevelDataFrameNamespace DataFrame
```
russcam: Is there some overlap in data_frames and transform APIs?

Mpdreamz: Discussed, decided to merge as is and tackle discuss this in the wider clients setting

PR #4349, opened 2020-02-03
Low level client should sent null for dictionary values.

Mpdreamz:
Isolated from: #4336


russcam:
Doesn't this change also affect `Dictionary<TKey, TValue>` and `SortedDictionary<TKey, TValue>` serialized by the high level client too?

Mpdreamz:
Discussed, we feel this is the better default for `8.x` and will call it out on the release notes.

PR #4351, opened 2020-02-03
DynamicValue can now dispatch into arrays

Mpdreamz:
e.g `.Get<string[]>("dataframes.id")` will work even if `dataframes` is a an array of objects. This makes picking values dynamically a ton easier.

(cherry picked from commit 1218888e47afe2147ec787d87fea4a2600c80da4)

Isolated from #4336

### DIFF
--- a/src/Elasticsearch.Net/Responses/Dynamic/DynamicValue.cs
+++ b/src/Elasticsearch.Net/Responses/Dynamic/DynamicValue.cs
@@ -8,6 +8,7 @@ using System.Globalization;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Runtime.CompilerServices;
+using Elasticsearch.Net.Extensions;
 using Microsoft.CSharp.RuntimeBinder;
 
 // ReSharper disable ArrangeConstructorOrDestructorBody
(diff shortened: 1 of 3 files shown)


tests/Tests/ClientConcepts/LowLevel/LowLevelResponseTypes.doc.cs
```diff
@@ -28,30 +28,30 @@ public class LowLevelResponseTypes
 
 		public static string Response()
 		{
-			return @"{
-			""boolean"" : true,	
+			return $@"{{
```
russcam: Looks like string interpolation is not needed?

src/Elasticsearch.Net/Responses/Dynamic/DynamicValue.cs
```diff
@@ -472,54 +512,117 @@ public T Default<T>(T defaultValue = default(T))
...
+					if (!valueType.IsArray)
+					{
+						return false;
+					}
+					var ar = (object[])value;
+					var t = targetReturnType.GetElementType();
+					var objectArray = ar
+						.Select(a => TryParse(defaultValue, t, a, out var o) ? o : null)
+						.Where(a => a != null)
+						//.Select(a => Convert.ChangeType(a, t))
```
russcam: can be removed?

PR #4353, opened 2020-02-04
Return distinct items from GetMany and SourceMany

### LINKED ISSUE #4342: SourceMany unexpected behavior
**NEST/Elasticsearch.Net version**: 7.3.0

**Elasticsearch version**: 7.2

I used Nest.IElasticClient for search by SourceMany and found some 
strange behavior. Api of method looks like you can send ids and you will get IEnumerable<Document> back. But if you send ["id1","id1","id1"] and Index contains one document with id = "id1", you will get 9 documents back instead of 3 for example. Because if you send multi get request to elastic api with 3 same ids you will get 3 documents in response. And I found that you have some join operation after elastic api response in MultiGetResponse.SourceMany, it creates duplicates.

**Steps to reproduce**:
 1. Save one document to index with Id = "id1"
 2. Call elasticClient.SourceMany<Document>(ids, IndexName);
 Ids = ["id1","id1","id1"]
 3. You get 9 documents back.

russcam:
This commit fixes a bug where a GetMany or SourceMany API call
with a repeated id would return a cartesian product of id and documents.

Take for example the same id repeated 3 times; Elasticsearch will return
3 documents in the response, where each JSON object is a representation
of the same underlying document. Since all 3 documents have the same id,
GetMany and SourceMany would return all 3 documents as a match for the first id,
all 3 as a match for the second id, and so on.

With this fix, For each id, only the first matching document is returned.
One _could_ argue that only a single document should be returned for the same
repeated id. This fix however tries to reflect what is in the
Elasticsearch response.

Fixes #4342

### DIFF
--- a/src/Nest/Document/Multiple/MultiGet/Response/MultiGetResponse.cs
+++ b/src/Nest/Document/Multiple/MultiGet/Response/MultiGetResponse.cs
@@ -63,28 +63,15 @@ namespace Nest
 		public T Source<T>(long id) where T : class => Source<T>(id.ToString(CultureInfo.InvariantCulture));
 
 		/// <summary>
-		/// Retrieves the source for each distinct id.
+		/// Retrieves the source, if available, for each distinct id.
 		/// </summary>
 		/// <param name="ids">The ids to retrieve source for</param>
 		/// <typeparam name="T">The document type for the hits to return</typeparam>
 		/// <returns>An IEnumerable{T} of sources</returns>
-		public IEnumerable<T> SourceMany<T>(IEnumerable<string> ids) where T : class
-		{
-			HashSet<string> seenIndices = null;
-			foreach (var id in ids.Distinct())
-			{
-				if (seenIndices == null)
-					seenIndices = new HashSet<string>();
-				else
-					seenIndices.Clear();
-
-				foreach (var doc in Hits.OfType<IMultiGetHit<T>>())
-				{
-					if (string.Equals(doc.Id, id) && doc.Found && seenIndices.Add(doc.Index))
-						yield return doc.Source;
-				}
-			}
-		}
+		public IEnumerable<T> SourceMany<T>(IEnumerable<string> ids) where T : class =>
+			from hit in GetMany<T>(ids)
+			where hit.Found
+			select hit.Source;
 
 		public IEnumerable<T> SourceMany<T>(IEnumerable<long> ids) where T : class =>
 			SourceMany<T>(ids.Select(i => i.ToString(CultureInfo.InvariantCulture)));
--- a/src/Nest/Document/Multiple/MultiGet/Response/MultiGetResponse.cs
+++ b/src/Nest/Document/Multiple/MultiGet/Response/MultiGetResponse.cs
@@ -27,15 +27,23 @@ namespace Nest
 			return multiHit?.Fields ?? FieldValues.Empty;
 		}
 
+		/// <summary>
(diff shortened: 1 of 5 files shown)


russcam:
This PR needs further work; an mget API call can return documents from one or more indices, such that documents with the same id from different indices can be returned.

russcam:
One approach that may be acceptable would be to return only a document _per_ index matching each `id` in the `IEnumerable<string>` id input. This does require keeping track of the indices seen for a given id, which means allocating a `HashSet<string>`, but would address the initial bug in #4342. It would still however return multiple documents for the same id but from different indices, but in the absence of some `Type` -> index name mapping, I can't see how this could be avoided.

Mpdreamz:
I think the returning multiple per `id` can not be avoided. Returning once per index seems fine to me.

russcam:
I've added a commit to keep track of the seen indices for each `id`. Thinking about it further, we could `.Distinct()` the passed `IEnumerable<T>` to `GetMany<T>()` and `SourceMany<T>()`, so that specifying the same id multiple times for a single targeted index would return a single document.

Thoughts?

russcam:
Ready for review again, @Mpdreamz, when you get a chance

src/Nest/Document/Multiple/MultiGet/Response/MultiGetResponse.cs
```diff
@@ -46,13 +62,28 @@ public class MultiGetResponse : ResponseBase
...
+		/// <returns>An IEnumerable{T} of sources</returns>
 		public IEnumerable<T> SourceMany<T>(IEnumerable<string> ids) where T : class
 		{
-			var docs = Hits.OfType<IMultiGetHit<T>>();
-			return from d in docs
-				join id in ids on d.Id equals id
-				where d.Found
-				select d.Source;
+			HashSet<string> seenIndices = null;
+			foreach (var id in ids.Distinct())
```
Mpdreamz: Piggy back on `GetMany` and select `.Source`

src/Nest/Document/Multiple/MultiGet/Response/MultiGetResponse.cs
```diff
@@ -27,12 +27,28 @@ public class MultiGetResponse : ResponseBase
...
 		public IEnumerable<IMultiGetHit<T>> GetMany<T>(IEnumerable<string> ids) where T : class
 		{
-			var docs = Hits.OfType<IMultiGetHit<T>>();
-			return from d in docs
-				join id in ids on d.Id equals id
-				select d;
+			HashSet<string> seenIndices = null;
+			foreach (var id in ids.Distinct())
+			{
+				if (seenIndices == null)
```
Mpdreamz: I wonder if we could rewrite this using `GroupBy()` instead. I understand how this initialization works but will probably do a double take every time I will read it in the future

russcam: I considered `GroupBy`, though I was thinking it'd probably allocate more than a single `HashSet<string>` (and the underlying `Set<T>` in `Distinct()`). Probably want to group by a composite of `doc.Id` and `doc.Index`.

The `HashSet<string>` _could_ be initialized on line 38, but the allocation might be unneccessary if there are no ids.

codebrain: `GroupBy()` needs allocation or a comparer, I think the `HashSet<string>` although more imperative, is easier to read.

PR #4364, opened 2020-02-04
[test] Unknown query now returns a different error

Mpdreamz:
Update our tests to reflect this


Mpdreamz:
Should be ported to `7.x`

https://dev.azure.com/azure3/elasticsearch-net/_build/results?buildId=2859&view=ms.vss-test-web.build-test-results-tab&runId=34478&resultId=100183&paneView=debug

Issue #4332, opened 2020-01-29
Refactor Metric aggregations


russcam:
As part of #4331, it looks like not all `IMetricAggregation` implementations support scripting i.e. the `Script` property 

https://github.com/elastic/elasticsearch-net/blob/4adf18a05ee51abdfa16e3e336bcdc40b4e48052/src/Nest/Aggregations/Metric/MetricAggregation.cs#L17

Going through the aggregation builder types in Elasticsearch, whether an aggregation supports scripting, formatting, etc. is controlled by declaring the supported fields on the parser. For example, for `MinAggregationBuilder`

https://github.com/elastic/elasticsearch/blob/17358b5af7b34a2b9a15846ba672f5a04001fcd5/server/src/main/java/org/elasticsearch/search/aggregations/metrics/MinAggregationBuilder.java#L48

```java
public class MinAggregationBuilder extends ValuesSourceAggregationBuilder.LeafOnly<ValuesSource.Numeric, MinAggregationBuilder> {
    public static final String NAME = "min";

    private static final ObjectParser<MinAggregationBuilder, Void> PARSER;
    static {
        PARSER = new ObjectParser<>(AvgAggregationBuilder.NAME);
        ValuesSourceParserHelper.declareNumericFields(PARSER, true, true, false); <---
    }

   ...
```

where the `bool` values are `scriptable`, `formattable` and `timezoneAware`, respectively.

Only the `IMetricAggregation` that support scripting should allow a `Script` property to be set. Going through the 7.x Java source, this is the support found

| agg name                      | scriptable         | formattable        |
| ----------------------------- | ------------------ | ------------------ |
| avg                           | :heavy_check_mark: | :heavy_check_mark: |
| cardinality                   | :heavy_check_mark: |  |
| extended_stats                | :heavy_check_mark: | :heavy_check_mark: |
| geo_centroid                  | :heavy_check_mark: |  |
| max                           | :heavy_check_mark: | :heavy_check_mark: |
| median_absolute_derivation    | :heavy_check_mark: | :heavy_check_mark: |
| min                           | :heavy_check_mark: | :heavy_check_mark: |
| percentile_ranks              | :heavy_check_mark: | :heavy_check_mark: |
| percentiles                   | :heavy_check_mark: | :heavy_check_mark: |
| stats                         | :heavy_check_mark: | :heavy_check_mark: |
| sum                           | :heavy_check_mark: | :heavy_check_mark: |
[...]

PR #4370, opened 2020-02-06
Add StringStats aggregation

### LINKED ISSUE #4369: Add String Stats Aggregation
Relates: #4341

Implement stats aggregation for string terms (elastic/elasticsearch#47468)

russcam:
This commit adds support for the string_stats aggregation introduced in
Elasticsearch 7.6.0. It is a metric aggregation but does not implement
IMetricAggregation because the type of Missing field is a string and not a
double value. Missing is implemented as object as the hierarchy of
metric aggregations will be changed as part of #4332, and Missing
will be an object.

The documentation for StringStats indicates that the distribution is returned
in descending probability order, but are modelled as a JSON object. Following
internal discussion, this is modelled as a dictionary on the response as it is
considered this modelling will not diminish functionality.

Closes #4369


src/Nest/Aggregations/AggregateFormatter.cs
```diff
@@ -397,16 +398,67 @@ private IAggregate GetSingleBucketAggregate(ref JsonReader reader, IJsonFormatte
...
 		}
 
-		private IAggregate GetStatsAggregate(ref JsonReader reader, IReadOnlyDictionary<string, object> meta)
+		private IAggregate GetStringStatsAggregate(ref JsonReader reader, IJsonFormatterResolver formatterResolver,
+			IReadOnlyDictionary<string, object> meta, long count
+		)
+		{
+			// string stats aggregation
+			var minLength = reader.ReadInt32();
+			reader.ReadNext(); // ,
```
Mpdreamz: comments 👍

russcam:
ported to master 9917ff687905711a28e06c18264a7ba1b14a740e

PR #4372, opened 2020-02-06
Add missing visit methods to aggregation visitor

russcam:
This commit updates the AggregationVisitorTests to ensure
that there is a Visit method for all aggregations and that
all aggregations (expect the intermediate aggregations) have a
Visit method. Missing visit methods added.


russcam:
@Mpdreamz, @codebrain should this be backported to 7.x? It is technically a binary breaking change as it adds new members to an interface that a consumer may have implemented (`IAggregationVisitor`). We have in the past added new members to interfaces within majors.

russcam:
ported to 7.x fc2d1723313b4c81f3e5dd791e90773f08cf31e3

PR #4376, opened 2020-02-06
Add per-field metadata

russcam:
Relates: #4341

This commit adds the ability to define per-field metadata in mappings,
which will be returned via the Get Mapping API and Field Capabilities API.


russcam:
This technically introduces a binary breaking change as it introduces the `Meta` property on `IProperty`. There's a reasonable chance that a user has implemented this interface for a custom property for a plufgin, for example. 

An alternative is to introduce the `Meta` property on a separate interface in 7.x, and include on `IProperty` in master. Thoughts @Mpdreamz, @codebrain?

codebrain:
`IProperty` is likely used by others for custom properties, so would suggest adding to a new interface for the time being and then unify in `8.0`.

Mpdreamz:
I'm okay with adding the property on `IProperty`, yes its breaking but similarly if new properties appear on `ISearchRequest` we would not introduce more interfaces.

Most of these interfaces are for for DTO, not behavioral components.

russcam:
I'm OK with including it in 7.x, so long as we explicitly call out in the release notes for 7.6.0 👍

russcam:
rebased on top of 7.x for a green test run

russcam:
ported to master 3b7a84729cc7d4944f0629b1a3298bf0a1701f44

PR #4378, opened 2020-02-10
Create exception formatters for different exceptions

### LINKED ISSUE #4294: Cannot serialize Exception as object
7.4.2
**Elasticsearch version**:
7.3.1

If I set a class that inherits Exception to the value of `Dictionary<string, object>` and post it to index, the following InvalidCastException is thrown.

```
Message: Unable to cast object of type 'Elasticsearch.Net.ExceptionFormatter' to type  'Elasticsearch.Net.Utf8Json.IJsonFormatter`1 [System.NullReferenceException]'.
StackTrace: Elasticsearch.Net.ExceptionFormatterResolver.GetFormatter<T>()
	Elasticsearch.Net.ElasticsearchNetFormatterResolver.InnerResolver.GetFormatter.AnonymousMethod__4_0<T>(System.Type)
	System.Collections.Concurrent.ConcurrentDictionary<TKey, TValue>.GetOrAdd(TKey, System.Func<TKey, TValue>)
	Elasticsearch.Net.ElasticsearchNetFormatterResolver.InnerResolver.GetFormatter<T>()
```

I hope it can be sent without problems.

**Steps to reproduce**:
Since it occurs at the time of serializing `PostData`, i attach the reproduction code
```csharp
var settings = new ConnectionConfiguration();

var dic = new Dictionary<string, object>()
{
    ["Exception"] = new NullReferenceException(),
};

var data = PostData.Serializable(dic);
using var ms = new MemoryStream();
data.Write(ms, settings); // Exception is thrown!!
```

russcam:
This commit changes the ExceptionFormatterResolver to return a new instance
of the now generic ExceptionFormatter<TException> for the type of exception.
This fixes a bug where an ExceptionFormatter of type IJsonFormatter<Exception>
is returned for derived exception types, resulting in an InvalidCastException
at runtime.

Formatter instances created are cached in ElasticsearchNetFormatterResolver
for the type T, so will be reused once created.

Fixes #4294

### DIFF
--- a/src/Elasticsearch.Net/Serialization/Formatters/ExceptionFormatter.cs
+++ b/src/Elasticsearch.Net/Serialization/Formatters/ExceptionFormatter.cs
@@ -3,30 +3,32 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Reflection;
 using System.Runtime.Serialization;
+using Elasticsearch.Net.Extensions;
 using Elasticsearch.Net.Utf8Json;
 
 namespace Elasticsearch.Net
 {
 	internal class ExceptionFormatterResolver : IJsonFormatterResolver
 	{
-		public static ExceptionFormatterResolver Instance = new ExceptionFormatterResolver();
+		public static readonly ExceptionFormatterResolver Instance = new ExceptionFormatterResolver();
 
 		private ExceptionFormatterResolver() { }
-
-		private static readonly ExceptionFormatter ExceptionFormatter = new ExceptionFormatter();
 
 		public IJsonFormatter<T> GetFormatter<T>()
 		{
 			if (typeof(Exception).IsAssignableFrom(typeof(T)))
-				return (IJsonFormatter<T>)ExceptionFormatter;
+			{
+				var type = typeof(ExceptionFormatter<>).MakeGenericType(typeof(T));
+				return (IJsonFormatter<T>)type.CreateInstance();
+			}
 
 			return null;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


src/Elasticsearch.Net/Serialization/Formatters/ExceptionFormatter.cs
```diff
@@ -3,30 +3,32 @@
...
 		private ExceptionFormatterResolver() { }
 
-		private static readonly ExceptionFormatter ExceptionFormatter = new ExceptionFormatter();
-
 		public IJsonFormatter<T> GetFormatter<T>()
 		{
 			if (typeof(Exception).IsAssignableFrom(typeof(T)))
-				return (IJsonFormatter<T>)ExceptionFormatter;
+			{
+				var type = typeof(ExceptionFormatter<>).MakeGenericType(typeof(T));
```
Mpdreamz: `.CreateGenericInstance()`

russcam: `CreateGenericInstance` exists in Nest, but `ExceptionFormatter` is in `Elasticsearch.Net`. Can move the extension method to `Elasticsearch.Net`?

Mpdreamz: I think we can, we can also follow up in a new issue.

russcam:
ported to master 2d63ce07d7d05fe21d20cf9b9784738a8450cfee

PR #4379, opened 2020-02-10
Add support for fuzzy intervals in interval query

russcam:
Relates: #4341

This commit adds support for fuzzy intervals source in
Intervals query


russcam:
Do we want the docs checks on 7.x, before the changes make it into a release?

russcam:
I'm gonna merge this in for now. The docs check makes sense to me for the master branch, but less so for other branches. For example, 7.x may contain doc additions that are not relevant until the features to which the docs relate are part of a release e.g. StringStats aggregation docs are relevant only when NEST 7.6.0 is released

russcam:
ported to master 46cd761588ede9f41ee36788c40cc146c751bd93

PR #4381, opened 2020-02-10
Feature/7.x/update ll client

Mpdreamz:
Update the `7.x` branch to latest greatest from `7.x`


russcam:
Is this generated from `7.x` or `7.6`? Should it be generated from `7.6`, if it's not?

Mpdreamz:
Should be 7.x. We should discuss branching our next minor immediately after releasing our current minor.

russcam:
> Is this generated from 7.x or 7.6? Should it be generated from 7.6, if it's not?

Should have clarified in case it was ambiguous, Elasticsearch branch from which the REST specs are pulled. 

> Should be 7.x. We should discuss branching our next minor immediately after releasing our current minor.

Might Elasticsearch 7.x contain changes in it though that are not for the 7.6.0 client release e.g. APIs that are scheduled for Elasticsearch 7.7.0.

codebrain:
My understanding is that we pull from _elastic/elasticsearch_ `7.6` branch into our _elastic/elasticsearch-net_ `7.x` branch.

That way we avoid picking up anything in the _elastic/elasticsearch_ `7.x` branch that might be unfinished/in progress/destined for a later minor version.

Mpdreamz:
Discussed offline `7.x` will track `7.x` from elastic/elasticsearch

PR #4383, opened 2020-02-11
Use correct IJsonFormatter<T> in attributes

### LINKED ISSUE #4382: VerifyRepositoryResponse uses incorrect JsonFormatter
Relates: https://stackoverflow.com/questions/60062480/elasticsearch-fails-verifyrepositoryrequest-in-c-sharp-nest-client

The `Nodes` member on `VerifyRepositoryResponse` uses the incorrect `IJsonFormatter<T>` in JsonFormatterAttribute, causing an `EntryPointNotFoundException`.

russcam:
This commit fixes a bug where VerifyRepositoryResponse was attributed
with the wrong formatter type in JsonFormatterAttribute applied at the type level.
Utf8Json's formatters are strongly typed for speed at the cost of less flexibility.

Fix other cases where the incorrect formatter type is used, and introduce a
convention test to assert that types and their properties attributed with
JsonFormatterAttribute use the correct formatter type.

Fixes #4382

### DIFF
--- a/src/Nest/Ingest/Processors/ScriptProcessor.cs
+++ b/src/Nest/Ingest/Processors/ScriptProcessor.cs
@@ -27,7 +27,7 @@ namespace Nest
 		/// Parameters for the script
 		/// </summary>
 		[DataMember(Name ="params")]
-		[JsonFormatter(typeof(VerbatimDictionaryInterfaceKeysFormatter<string, object>))]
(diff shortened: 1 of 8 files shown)


russcam:
opened https://github.com/elastic/elasticsearch-net/issues/4385 to tackle failing enrich policy test, which are unrelated to this PR.

russcam:
ported to master ec504010b51a6215ac4e604f4eb90f9ce3bd1950

PR #4395, opened 2020-02-19
Update low level client on 7.x

Mpdreamz:


Mpdreamz:
Client CI is failing because these skips https://github.com/elastic/elasticsearch/blob/7.x/x-pack/plugin/src/test/resources/rest-api-spec/test/analytics/top_metrics.yml#L62

Have not made their way to the latest snapshot yet

PR #4400, opened 2020-02-20
Return distinct items from GetMany and SourceMany (#4353)

### LINKED ISSUE #4342: SourceMany unexpected behavior

russcam:
This commit fixes a bug where a GetMany or SourceMany API call
with a repeated id would return a cartesian product of id and documents.

- enumerate only distinct ids when retrieving hits
or source from GetMany and SourceMany, so that the same id input
to either will return only a single document per target index.

- fix a bug in the MultiGetRequestFormatter whereby
the document index is removed when a request index is specified,
without checking whether the document index matches the request index.

Fixes #4342

(cherry-picked from commit 8cbc1fe35d1d3fe618115219e9394ae0f3adc27a)

### DIFF
--- a/docs/aggregations/bucket/terms/terms-aggregation-usage.asciidoc
+++ b/docs/aggregations/bucket/terms/terms-aggregation-usage.asciidoc
@@ -206,7 +206,7 @@ new TermsAggregation("states")
 [source,csharp]
 ----
 response.ShouldBeValid();
-var states = response.Aggregations.Terms("states");
+var states = response.Aggregations.Terms<StateOfBeing>("states");
 states.Should().NotBeNull();
 states.DocCountErrorUpperBound.Should().HaveValue();
 states.SumOtherDocCount.Should().HaveValue();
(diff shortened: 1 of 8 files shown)


russcam:
this is a backport of #4353 to 6.x

PR #4447, opened 2020-02-24
Fix yaml tests not deleting newly defined hidden indices

Mpdreamz:
Before we were not deleting system indices only `.ml-*`

However this started to become an issue because it left closed indices
around. This might be related to .ml indices now being hidden.

Warrants more investigation as the new solution deleting all indices all
the time is definitely a bit slower


Mpdreamz:
Merging this in as `client-ci` is  happy.

Mpdreamz:
Manually backported to `master` and ran the yaml tests locally

PR #4456, opened 2020-02-25
Fix API generator

russcam:
This commit fixes the API generator and removes REST API specs
ported from later version branches:

1. The collection of high level endpoints were being used in some
of the low level generators
2. High level endpoints excluded need to be compared to the list by filename
3. Exclude new ML experimental APIs


codebrain:
Co-authored over Zoom, LGTM

PR #4457, opened 2020-02-25
Fix build release for 7.6.0

russcam:
This commit fixes the release build for 7.6.0.
The global.json and auto-label.json version should be
the last 7.x released Nuget package


codebrain:
Co-authored over Zoom, LGTM

PR #4460, opened 2020-02-26
Update readme.md for LLC compatibility

codebrain:
Add remarks around beta and experimental functionality in the low level client.

Suggest we wait for merge until `7.7.0`

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -62,7 +62,7 @@ Repository for both **NEST** and **Elasticsearch.Net**, the two official [Elasti
 
 Please refer to the [end-of-life policy](https://www.elastic.co/support/eol) for complete information.
 
-### Further Compatibility Clarifications
+## Further Compatibility Clarifications
 
 #### Can I use a `7.0.0` client against a `5.6` server (new against old), or a `6.0` client against a `7.2` server (old against new)?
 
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -88,6 +88,14 @@ Any incompatibilities between minor versions are documented against the release.
...
 
+### Low Level Client Compatibility
+
+The Elasticsearch.Net low level client will expose functionality in REST APIs that are marked as either `experimental` or `beta`. This functionality is marked as such using the `<remarks/>` XML documentation comments, which contains the stability and a description of the compatibility guarantees. Examples below:
+
+> /// <remarks>Note: Experimental within the Elasticsearch server, this functionality is experimental and may be changed or removed completely in a future release. Elastic will take a best effort approach to fix any issues, but experimental features are not subject to the support SLA of official GA features.</remarks>
+
+> /// <remarks>Note: Beta within the Elasticsearch server, this functionality is in beta and is subject to change. The design and code is less mature than official GA features and is being provided as-is with no warranties. Beta features are not subject to the support SLA of official GA features.</remarks>
```
russcam: I think we ought to say something about how this affects binary compatibility too, if a user decides to use an experimental or beta feature that ends up changing in a breaking way within a major. As far as I am aware, experimental and beta APIs are never deleted within a major, but they can change in ways that would manifest as breaking changes in the LLC.

codebrain: Agreed.

This is not 100% clear in the XML comments that we currently emit, so I think a refresh in there is in order too.

Mpdreamz:
I think this can be merged and backported to `7.x` now thats tracking `7.x-SNAPSHOT` 👍

codebrain:
Merged into `master` and backported to `7.x` manually.

PR #4476, opened 2020-03-04
Adds a comment with source:line in the asciidoc code examples.

codebrain:
Using this metadata we can have a reference of the example with the source file and the line code. This can be useful to match outdated examples.

I considered adding some additional nomenclature in the generated C#, but figured it was cleaner and easier to simply require the reference file again in the generation process and perform a lookup.


src/ExamplesGenerator/AsciiDocGenerator.cs
```diff
@@ -28,8 +30,17 @@ static AsciiDocGenerator()
...
+		public static void GenerateExampleAsciiDoc(IEnumerable<ImplementedExample> examples, string branchName, string path)
 		{
+			var referencePages = AsciiDocParser.ParsePages(path);
+
+			string GetReferencePage(ImplementedExample e)
+			{
+				return referencePages.Select(p => new { Hashes = p.Examples.Select(e => e.Hash), File = p.Name })
+					.Where(o => o.Hashes.Contains(e.Hash))
+					.Select(h => h.File).FirstOrDefault();
+			}
```
russcam: I think it would be better to write the reference page file into the generate example method e.g. as a comment on the method, and parse it out when creating instances of `ImplementedExample`

src/ExamplesGenerator/AsciiDocGenerator.cs
```diff
@@ -65,6 +76,9 @@ public static void GenerateExampleAsciiDoc(IEnumerable<ImplementedExample> examp
 					.AppendLine("If you wish to submit a PR to change this example, please change the source method above")
 					.AppendLine("and run dotnet run -- asciidoc in the ExamplesGenerator project directory.")
 					.AppendLine("////")
+					.AppendLine()
+					.AppendLine($"// {GetReferencePage(example)}.asciidoc:{example.StartLineNumber}")
```
russcam: Think this should be on the first line of the file, based on the proposal?

russcam: `{example.StartLineNumber}` refers to the startline of the C# method in the file and not the line number of the console example in the alternative report.

src/ExamplesGenerator/AsciiDocGenerator.cs
codebrain: Potentially, I think having the lookup is also an additional useful bookeeping step.

src/ExamplesGenerator/AsciiDocGenerator.cs
codebrain: OK, I will replace with the alternate line number and push to the top of the file.

src/ExamplesGenerator/CSharpParser.cs
```diff
@@ -39,11 +39,17 @@ public static IEnumerable<ImplementedExample> ParseImplementedExamples()
...
 				{
-					var uAttribute = methodDeclaration.AttributeLists.Single().Attributes.Single();
+					var uAttribute = methodDeclaration.AttributeLists.Single(a => a.Attributes.Any(aa => aa.Name.ToString() == "U")).Attributes.Single();
 
 					// No Skip property present on [U] attribute
 					if (uAttribute.ArgumentList == null)
 					{
+						var descriptionAttributeSyntax = methodDeclaration.AttributeLists.SingleOrDefault(a => a.Attributes.Any(aa => aa.Name.ToString() == "Description"));
+
+						var referenceFileAndLineNumber = "";
```
russcam: ternary conditional operator assignment

Mpdreamz:
This was merged but did not compile:

```
Build FAILED.

Setup\RestartClusterPage.cs(27,15): error CS0111: Type 'RestartClusterPage' already defines a member called 'Line30' with the same parameter types [D:\a\1\s\tests\Examples\Examples.csproj]
    0 Warning(s)
    1 Error(s)
```

PR #4477, opened 2020-03-04
Delete dependant fles generated by the ApiGenerator before generating them

### LINKED ISSUE #4458: ApiGenerator must delete existing generated files
When the API generator is run to generate components of the low level and high level clients, it does not delete previously generated files. It should delete existing generated files

codebrain:
Fixes #4458

### DIFF
--- a/src/ApiGenerator/Generator/Razor/DescriptorsGenerator.cs
+++ b/src/ApiGenerator/Generator/Razor/DescriptorsGenerator.cs
@@ -1,3 +1,4 @@
+using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using ApiGenerator.Configuration;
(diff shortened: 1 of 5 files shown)


src/ApiGenerator/Generator/Razor/DescriptorsGenerator.cs
```diff
@@ -12,6 +13,10 @@ public class DescriptorsGenerator : RazorGeneratorBase
 
 		public override async Task Generate(RestApiSpec spec, ProgressBar progressBar)
 		{
+			// Delete existing files
+			foreach (var file in Directory.GetFiles(GeneratorLocations.NestFolder, "Descriptors.*.cs"))
```
russcam: pull out `"Descriptors"` into a `const` e.g.

```c#
const string Descriptors = nameof(Descriptors);
```

Might be good to for all generators

PR #4478, opened 2020-03-04
Update .ci folder to latest incarnation

Mpdreamz:
The .ci folder cleanup some of its cleanup/wait routines so that they
can be reused. It also significantly reduced the available environment
variable toggles. In lieu of those toggles `run-repository.sh` can now
start multiple nodes using `NUMBER_OF_NODES`


Mpdreamz:
Going to merge this in clients-ci is happy but won't run until this is in master due to

PR #4485, opened 2020-03-09
Deduplicate example method names

russcam:
This commit deduplicates the generated example method names,
when generating C# methods from the examples. The method names
follow the convention of `Line<linenum>`, where `linenum` is the
line number in the alternative report. The alternative report
can report duplicate line numbers however, as reported in

elastic/docs#1760

so deduplicate in the generator.


russcam:
Added a commit to fix a bug where the `DescriptionAttribute` was not updated when the line number changed.

PR #4502, opened 2020-03-20
Remove MemoryStreamFactory and replace with RecyclableMemoryStreamFactory

codebrain:
The only usage of `MemoryStreamFactory` internally is withing tests/documentation.

Removing these uses as its doubtful if use in this situation is worthwhile.

This does remove the `MemoryStreamFactory` implementation as well, so this is technically a breaking change, so could be made non-breaking by reinstating this class.


Mpdreamz:
One of its use cases is also to benchmark/profile the recyclable memorystream factory against one that does not pool. 

I'm okay reinstating the class existence for folks for whom the memory characteristic of recyclable memory stream come as a surprise. We had tickets in the past where folks suspected a memory leak due to recyclable memorystreams buffers. In some cases it might also be preferable to have GC to be more aggressive (e.g cron tooling)

codebrain:
> One of its use cases is also to benchmark/profile the recyclable memorystream factory against one that does not pool.
> 
> I'm okay reinstating the class existence for folks for whom the memory characteristic of recyclable memory stream come as a surprise. We had tickets in the past where folks suspected a memory leak due to recyclable memorystreams buffers. In some cases it might also be preferable to have GC to be more aggressive (e.g cron tooling)

I think reinstating is fine, lets just not use it internally.

codebrain:
Closed in favor of: https://github.com/elastic/elasticsearch-net/pull/4509

PR #4503, opened 2020-03-20
ISO8601TimeSpanFormatter to use local byte[] array, as opposed to renting from pool.

codebrain:
The minimum buffer size that can be rented from the shared memory pool is 65535 bytes. This amount of memory for the purposes of serializing the day component of a string timespan is excessive. This commit changes the procedure to use a local byte array

Adds test for max size of 64bit value.


codebrain:
Closed in favor of: https://github.com/elastic/elasticsearch-net/pull/4509

PR #4504, opened 2020-03-20
Removes pretty print serialisation as it is not used.

codebrain:


codebrain:
Closed in favor of https://github.com/elastic/elasticsearch-net/pull/4509

PR #4544, opened 2020-04-02
Fix Node Info

codebrain:
This commit adds some missing roles to NodeRole and some additional properties to NodeInfo.

Fixes broken integration test against 7.7.0-SNAPSHOT

### DIFF
--- a/src/Nest/Cluster/NodesInfo/NodeInfo.cs
+++ b/src/Nest/Cluster/NodesInfo/NodeInfo.cs
@@ -8,35 +8,32 @@ namespace Nest
 	[DataContract]
 	public class NodeInfo
 	{
-		[DataMember(Name = "build_hash")]
-		public string BuildHash { get; internal set; }
+		[DataMember(Name = "name")]
+		public string Name { get; internal set; }
+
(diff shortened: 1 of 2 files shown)


src/Nest/Cluster/NodesInfo/NodeInfo.cs
```diff
@@ -8,35 +8,32 @@ namespace Nest
...
 
-		[DataMember(Name = "plugins")]
-		public List<PluginStats> Plugins { get; internal set; }
+		[DataMember(Name = "build_hash")]
+		public string BuildHash { get; internal set; }
 
-		[DataMember(Name = "process")]
-		public NodeProcessInfo Process { get; internal set; }
+		[DataMember(Name = "total_indexing_buffer")]
+		public long TotalIndexingBuffer { get; internal set; }
```
russcam, line 36: is this a new property in 7.7.0? I think we should add an XML comment to say which versions it's relevant to, if not relevant for all 7.x versions.

codebrain: Not sure, I will make it nullable.

PR #4549, opened 2020-04-03
Remove snapshot testing from YAML run

codebrain:
Remove snapshot tests from YAML run. Tests fail due to this setup having a non directory location:

```
setup:

  - do:
      snapshot.create_repository:
        repository: test_repo_create_1
        body:
          type: fs
          settings:
            location: "test_repo_create_1_loc"
```

Subsequent tests then fail as `test_repo_create_1` has not been created.


Mpdreamz:
We do specify `path.repos=/tmp` at startup like the java runner does:

https://github.com/elastic/elasticsearch/commit/177180ac94ea08709bc52036483319f722750be1#diff-357e4854869b2e21c38b1b437f11095aR1055

So this should have been sufficient enough.

PR #4562, opened 2020-04-06
Implement cat.ml_trained_models

codebrain:
Implements: https://www.elastic.co/guide/en/elasticsearch/reference/7.7/cat-trained-model.html


codebrain:
Manually backported to master

PR #4579, opened 2020-04-13
Use `params` modifier for methods in RequestDescriptorBase

### LINKED ISSUE #4471: Improve LLC descriptors generation
Consider generating `params` versions of array on querystring types, for example `filter_path`:

```
///<summary>A comma-separated list of filters used to reduce the response.<para>Use of response filtering can result in a response from Elasticsearch that cannot be correctly deserialized to the respective response type for the request. In such situations, use the low level client to issue the request and handle response deserialization</para></summary>
public TDescriptor FilterPath(string[] filterpath) => Qs("filter_path", filterpath);
```

It'd be nice to write:

`var getIndexTemplateResponse = client.Indices.GetTemplate("template_1", t => t.FilterPath("*.version"));`

instead of:

`var getIndexTemplateResponse = client.Indices.GetTemplate("template_1", t => t.FilterPath(new[] { "*.version" }));`

and avoid the extra baggage of newing the array outside of the method

tkirill:
Hello!

This PR tries to fix #4471.

The only place I found to fix is `RequestDescriptorBase.FilterPath` so either this is a really low hanging fruit or I'm completely missing something.  I searched for potential places to fix with naive search for `(string[]` and with looking at usings of `TypeHighLevel` in views.

I noticed that methods in `Descriptors.*.cs` already have `params` modifier and then found that it is implemented with `.DescriptorArgumentType` in `Descriptor.cshtml`.  Looks like it is suitable for `RequestDescriptorBase` too.

I'm not sure if `Descriptors.cs` should be commited.  The whole process of testing codegen is a little bit confusing and I didn't figured out the right way to run tests locally for my changes (and will be glad if someone teach me).  Anyway it was a very interesting exercise to look through the code generation tool.

### DIFF
--- a/src/ApiGenerator/Domain/Specification/QueryParameters.cs
+++ b/src/ApiGenerator/Domain/Specification/QueryParameters.cs
@@ -55,8 +55,14 @@ namespace ApiGenerator.Domain.Specification
 			}
 		}
 
-		public string DescriptorArgumentType =>
-			Type == "list" && TypeHighLevel.EndsWith("[]") ? "params " + TypeHighLevel : TypeHighLevel;
+		public bool IsArray => Type == "list" && TypeHighLevel.EndsWith("[]");
+
+		public string DescriptorArgumentType => IsArray ? "params " + TypeHighLevel : TypeHighLevel;
+
+		public string DescriptorEnumerableArgumentType =>
+			IsArray
+				? $"IEnumerable<{TypeHighLevel.TrimEnd('[', ']')}>"
+				: throw new InvalidOperationException("Only array arguments have IEnumerable overload");
 
 		public Func<string, string, string, string, string> FluentGenerator { get; set; }
 		public bool IsFieldParam => TypeHighLevel == "Field";
--- a/src/ApiGenerator/Views/HighLevel/Descriptors/RequestDescriptorBase.cshtml
+++ b/src/ApiGenerator/Views/HighLevel/Descriptors/RequestDescriptorBase.cshtml
@@ -7,6 +7,8 @@
 @{
 	RestApiSpec m = Model;
 }
+using System.Collections.Generic;
+
 namespace Nest
 {
 	// ReSharper disable UnusedTypeParameter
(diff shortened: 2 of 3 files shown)


Mpdreamz:
Thanks @tkirill for taking on this challenge!

Also your input into the documentation of this process is invaluable we'll work on improving our Contributing.md 👍 

This is not a breaking change as per: https://docs.microsoft.com/en-us/dotnet/core/compatibility/#code-changes so LGTM. 

I do think wherever we accept `params` we should create a non allocating overload with `IEnumerable` too. Are you up for the challenge to update this PR? If not no sweat we'll amend afterwards.

tkirill:
@Mpdreamz Yes, sure I want, this is very interesting challenge.  At least I'll try.

I guess I don't fully understand the scope of the task.  This is how I see it:

* For each API endpoint there are two parts: low-level (RequestParameters.*.cs, ElasticLowLevelClient.*.cs) in Elasticsearch.Net and high-level in Nest.  This task is about high-level code.
* High-level code can be auto-generated (Descriptors.*.cs, Requests.*.cs, ElasticClient.*.cs) and manual-writed.  Manual-writed descriptors always have both `IEnumerable` and `params` overloads.  Auto-generated descriptors have only `params` version and we want to add `IEnumerable` overload.
* So the targets are Descriptor.cshtml and RequestDescriptorBase.cshtml

If this is far from being the whole story then I guess it will take unreasonably much of your time to explain me what to do.  In this case I'll be happy to stop on current tiny contribution and just wait and look at full solution later.  And if everything is actually fine then I can do it and it will be very interesting.

Mpdreamz:
Hey @tkirill your suggested scope is the whole nine yards :)

I was talking about simply extending this PR so that ``src/Nest/Descriptors.cs also includes the overload. Keeps the PR small and manageable 👍

tkirill:
Hey @Mpdreamz I added `IEnumerable` overload.

codebrain:
This is great @tkirill! thank you for your contribution.
I will get this merged in and ready for `7.7.0`.

codebrain:
Backported to `7.x` and `7.7`

PR #4580, opened 2020-04-13
Update custom serialization docs

russcam:
This commit updates the custom serialization docs that discuss using
JsonNetSerializer.

1. Remove the implication that the internal serializer
is based on Json.NET.
2. Include recommendation to hook up JsonNetSerializer when using
Newtonsoft.Json.Linq types like JObject


russcam:
Failures look like they're related to 

> org.elasticsearch.bootstrap.StartupException: java.lang.IllegalArgumentException: unknown setting [cluster.remote.connect] did you mean [cluster.remote.node.attr]  A change has come in on master thats failing our integration tests.

and not this change, which is simply a doc change.

PR #4592, opened 2020-04-15
Implement Boxplot aggregation

russcam:
Relates: elastic/elasticsearch#51948

This commit implements the boxplot aggregation.
Integration tests run against XPackCluster because
it requires a license to use.


russcam:
Will update this to run against ReadOnlyCluster, which uses the default distribution, so has a license applied.

russcam:
manually ported to 7.x and 7.7

PR #4599, opened 2020-04-16
Implement hidden indices and aliases

russcam:
Relates: elastic/elasticsearch#52547
Relates: elastic/elasticsearch#53248

This commit adds the ability to mark indices
and aliases as hidden.


russcam:
manually backported to 7.7 and 7.x

PR #4602, opened 2020-04-16
Regnerate from new specifications

codebrain:


src/Elasticsearch.Net/ElasticLowLevelClient.Nodes.cs
```diff
@@ -108,25 +108,29 @@ public TResponse Info<TResponse>(string nodeId, string metric, NodesInfoRequestP
...
 			where TResponse : class, IElasticsearchResponse, new() => DoRequestAsync<TResponse>(GET, Url($"_nodes/{nodeId:nodeId}/{metric:metric}"), ctx, null, RequestParams(requestParameters));
 		///<summary>POST on /_nodes/reload_secure_settings <para>https://www.elastic.co/guide/en/elasticsearch/reference/master/secure-settings.html#reloadable-secure-settings</para></summary>
+		///<param name = "body">An object containing the password for the elasticsearch keystore</param>
 		///<param name = "requestParameters">Request specific configuration such as querystring parameters &amp; request specific connection settings.</param>
-		public TResponse ReloadSecureSettingsForAll<TResponse>(ReloadSecureSettingsRequestParameters requestParameters = null)
-			where TResponse : class, IElasticsearchResponse, new() => DoRequest<TResponse>(POST, "_nodes/reload_secure_settings", null, RequestParams(requestParameters));
+		public TResponse ReloadSecureSettingsForAll<TResponse>(PostData body, ReloadSecureSettingsRequestParameters requestParameters = null)
```
russcam, line 113: Is this a breaking change to a stable API i.e. has this API been released in a previous client version?

Mpdreamz: Good catch: The previous commit on this method was https://github.com/elastic/elasticsearch-net/pull/3762 from last year so sadly a breaking change.

https://github.com/elastic/elasticsearch/pull/54771

We'll have to come up with a way to generate an additional overload instead marking the others deprecated.

codebrain: ++

I think the most sensible thing to do here is generate the llc namespaces as partials and then patch in with a custom class.

codebrain:
Manually ported to 7.x

Issue #4610, opened 2020-04-16
7.7.1 meta ticket


codebrain:
## New APIS (bumped from 7.7.0)

 - [x] async_search.delete.json
 - [x] async_search.get.json
 - [x] async_search.submit.json
 - [x] transform.delete_transform.json
 - [x] transform.get_transform.json
 - [x] transform.get_transform_stats.json
 - [x] transform.preview_transform.json
 - [x] transform.put_transform.json
 - [x] transform.start_transform.json
 - [x] transform.stop_transform.json
 - [x] transform.update_transform.json

russcam:
Closing as 7.7.1 is released

PR #4637, opened 2020-04-20
Implement async_search APIs

russcam:
Relates: #4610 

This commit implements the async_search APIs, introduced in Elasticsearch 7.7.

The body of `AsyncSearchSubmitRequest` is similar to `SearchRequest` with some differences:
1. async search does not support scrolling, so there is no sliced scroll
2. the `IRequestParameters` implementation is `AsyncSearchSubmitRequestParameters` and `IAsyncSearchSubmitRequest` implements `IRequest<AsyncSearchSubmitRequestParameters>`

The implementation hasn't tried to align `SearchRequest` and `AsyncSearchSubmitRequest` for these reasons.

API generator re-ran against upstream master, in order to pull in any new changes.


russcam:
> I think its better if the integration tests move to a single CoordinatedTest rather then individual API tests wit separate bootstrapping.

Is the branch where state can be shared across coordinate usage test steps checked in? If it is, I can update the tests

Mpdreamz:
@russcam yes it was an additional commit to your branch: https://github.com/elastic/elasticsearch-net/commit/c742a1fd0f5924d3e68311042d9cc4121a17bc0a

PR #4638, opened 2020-04-21
ApiGenerator should remove old specification files from disk before redownloading.

codebrain:
This is to ensure that the specifications are up-to-date and accurate from the download branch and no "orphans" are carried forward.

Fixes #4636

PR #4640, opened 2020-04-21
Clean spec folder before downloading a fresh copy

Mpdreamz:
Alternative approach to #4638, keeping it isolated to RestSpecDownloader


Mpdreamz:
Self merging this one in since, a lot of follow up work depends on it.

PR #4649, opened 2020-04-21
PrimitiveObjectFormatter eagerly reads integer where double could be preferred.

Mpdreamz:
A bug in `PrimitiveObjectFormatter` which is used when deserializing
`object` was eagerly reading doubles as integers. This was flagged in
failing yaml test reproduce:

> ./build.sh rest-spec-tests xpack -f analytics -t t_test.yml -s "heteroscedastic t-test"

Scrollbars now also collapse when running the yaml tests now that
ShellProgressbar properly scrolls children into view this got to be a
tad excessive

Issue #4652, opened 2020-04-21
CSV empty_value might not be available on master


Mpdreamz:
#4597 introduces `empty_value` on `master` which is documented to be available as per: https://www.elastic.co/guide/en/elasticsearch/reference/master/csv-processor.html

However we did see an integration test failure pop up related to this:

https://dev.azure.com/azure3/elasticsearch-net/_build/results?buildId=3689&view=logs&j=70d57b1a-3bb3-5656-1da8-d847c6de6c92&t=1327d875-a30a-5de5-c423-74b851577e38

```
[xUnit.net 00:14:30.6164584]     Tests.Ingest.PutPipeline.PutPipelineApiTests.ReturnsExpectedIsValid [FAIL]
[2020-04-21T20:07:29,326][INFO ][o.e.x.s.a.r.TransportPutRoleAction] [xpack-node-7bea8e9200] added role [role-nest-initializer-96b8dc0e]
[2020-04-21T20:07:29,356][INFO ][o.e.x.s.a.r.TransportPutRoleAction] [xpack-node-7bea8e9200] added role [role-ois-baae57e0-role]
  X Tests.Ingest.PutPipeline.PutPipelineApiTests.ReturnsExpectedResponse [7s 187ms]
  Error Message:
   Tests.Framework.EndpointTests.ResponseAssertionException : Expected response.Acknowledged to be true, but found False.
Response Under Test:
Invalid NEST response built from a unsuccessful (400) low level call on PUT: /_ingest/pipeline/pipeline-1?pretty=true&error_trace=true
# Audit trail of this API call:
 - [1] BadResponse: Node: https://localhost:9200/ Took: 00:00:01.2789750
# OriginalException: Elasticsearch.Net.ElasticsearchClientException: Request failed to execute. Call: Status code 400 from: PUT /_ingest/pipeline/pipeline-1?pretty=true&error_trace=true. ServerError: Type: parse_exception Reason: "processor [csv] doesn't support one or more provided configuration parameters [empty_value]"
# Request:
[...]

Issue #4011, opened 2019-08-09
[tests] putpipeline 7.x ProcessClusterEventTimeoutException


Mpdreamz:
`build.bat seed:9368 integrate 7.0.0 "writable" "putpipeline"`

From: https://github.com/elastic/elasticsearch-net/pull/4004

`ProcessClusterEventTimeoutException` it'd be good if we can somehow improve the error reporting when we get this exception from the server in general.

Capture:
- hotthreads API
- get tasks API

This one is hard to diagnose and replicate. A mitigation might be to move the ingest tests out of writable cluster to make them more isolated.

```
[xUnit.net 00:07:44.3679353]     Tests.Ingest.PutPipeline.PutPipelineApiTests.ReturnsExpectedResponse [FAIL]
Failed   Tests.Ingest.PutPipeline.PutPipelineApiTests.ReturnsExpectedResponse
Error Message:
 Tests.Framework.EndpointTests.ResponseAssertionException : Expected response.Acknowledged to be true, but found False.
Response Under Test:
Invalid NEST response built from a unsuccessful (503) low level call on PUT: /_ingest/pipeline/pipeline-1?pretty=true&error_trace=true
# Audit trail of this API call:
 - [1] BadResponse: Node: http://localhost:9200/ Took: 00:00:30.0527811
# OriginalException: Elasticsearch.Net.ElasticsearchClientException: The remote server returned an error: (503) Server Unavailable.. Call: Status code 503 from: PUT /_ingest/pipeline/pipeline-1?pretty=true&error_trace=true. ServerError: Type: process_cluster_event_timeout_exception Reason: "failed to process cluster event (put-pipeline-pipeline-1) within 30s" ---> System.Net.WebException: The remote server returned an error: (503) Server Unavailable.
   at System.Net.HttpWebRequest.GetResponse() 
   at Elasticsearch.Net.HttpWebRequestConnection.Request[TResponse](RequestData requestData) in D:\a\1\s\src\Elasticsearch.Net\Connection\HttpWebRequestConnection.cs:line 59
   --- End of inner exception stack trace ---
# Request:
[...]

PR #4655, opened 2020-04-22
Fix bug in IsLong implementation

russcam:
Relates: #4649

This commit fixes a bug in the IsLong implementation, and prefers
reading long values over double values in PrimitiveObjectFormatter
when the bytes represent a valid long value, to avoid floating
point rounding errors that may happen when parsing a double from
a long value.

The bug in IsLong was that the loop over the array segment bytes was
exiting early when the count of bytes was the same as long.Min/MaxValue
length, and the current loop byte is less than the byte at index in
long.Min/MaxValue. In this scenario, it should not exit early, but
not continue to check following loop bytes against byte at index in
long.Min/MaxValue. Following loop iterations still need to check
whether remaining bytes are numbers.

### DIFF
--- a/src/Elasticsearch.Net/Extensions/ArraySegmentBytesExtensions.cs
+++ b/src/Elasticsearch.Net/Extensions/ArraySegmentBytesExtensions.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Globalization;
 using System.Runtime.CompilerServices;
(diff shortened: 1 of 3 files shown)


Mpdreamz:
Failing tests are:

- process_cluster_event_timeout_exception #4011 
- CSV empty_value #4652

PR #4656, opened 2020-04-22
Don't sanitize the keys for Filters aggregation named buckets

### LINKED ISSUE #4582: NEST swallows part of an aggregation value
**NEST/Elasticsearch.Net version**: 6.8.1

**Elasticsearch version**: 6.7

I have a filter aggregation where one of the values in the response is "ff3046ce-e547-46b6-8ffb-44832bd8c2ad c# .net".
When I read that response with NEST it says the value is " .net". That is, it swallows the first part of the value for some reason.

This seems to be related to https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Aggregations/AggregateDictionary.cs#L16-L34 which I see is used in the serializer https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Aggregations/AggregateDictionaryFormatter.cs#L31 to handle the `typed_keys` parameter, but regardless of what this is set to in the request, this serializer seems to behave the same?
Are there any workarounds? Can this serializer be overriden somehow?

Thanks

EDIT: I'm using NEST 6.x here but the code I'm linking to is still in master so i'm pretty sure this still happens in the latest versions of NEST.

russcam:
This commit fixes a bug where the keys in the dictionary passed
to filters aggregation are sanitized when copying to the backing
dictionary. Keys in this scenario are the keys for named buckets
which should never be sanitized.

Fixes #4582

### DIFF
--- a/src/Nest/Aggregations/Bucket/Filters/FiltersAggregate.cs
+++ b/src/Nest/Aggregations/Bucket/Filters/FiltersAggregate.cs
@@ -28,6 +28,9 @@ namespace Nest
 
 		public FiltersAggregate(IReadOnlyDictionary<string, IAggregate> aggregations) : base(aggregations) { }
 
+		// Don't sanitize the keys as these are the keys for named buckets
+		protected override string Sanitize(string key) => key;
+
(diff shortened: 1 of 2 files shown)


Mpdreamz:
failing test:

- Csv empty_value #4652

PR #4663, opened 2020-04-22
Fix check for agressive connectionlimit

Mpdreamz:
In https://github.com/dotnet/runtime/issues/22366 we found that if
HttpClient is using the curl handler it will lead to TCP connections
bleeding. Our DefaultConnectionLimit is very restrictive if this is
true. Our check however is too lenient and will default to true always
on .NET core since netcoreapp still ships with CurlHandler as recent as
`3.1.x`

### DIFF
--- a/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
+++ b/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
@@ -5,6 +5,7 @@ using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Net;
 #if DOTNETCORE
 using System.Net.Http;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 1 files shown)


src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs
```diff
@@ -23,12 +23,34 @@ namespace Elasticsearch.Net
...
 	{
-#if DOTNETCORE
-		private static bool IsCurlHandler { get; } = typeof(HttpClientHandler).Assembly.GetType("System.Net.Http.CurlHandler") != null;
+		/// <summary>
+		/// Detects whether we are running on .NET Core without SocketsHttpHandler existing or being enabled
+		/// If this is true we will set a very restrictive <see cref="DefaultConnectionLimit"/>
+		/// As the old curl based handler is known to bleed TCP connections:
+		/// <para>https://github.com/dotnet/runtime/issues/22366</para>
+		/// </summary>
+        private static bool OnDotNetCoreWithNoSocketsHttpHandlerEnabled
```
russcam, line 32: the double negative is quite hard to mentally parse! I think the logic might be better represented as `SocketsHttpHandlerPresentAndDisabled`, or perhaps renaming to `UsingCurlHandler` and detecting `System.Net.Http.CurlHandler` too, as this implementation will also impose the `Environment.ProcessorCount` connection limit on Windows .NET Core when `System.Net.Http.UseSocketsHttpHandler` is disabled and falls back to `System.Net.Http.WinHttpHandler`

russcam:
I've added a commit of a slightly updated implementation based on PR comment

Mpdreamz:
1 failing test
- CSV should be fixed with #4666

Mpdreamz:
Flipping the check was a good call, thanks @russcam 👍

Mpdreamz:
Opened #4669 as manual backport to `6.x`

PR #4665, opened 2020-04-23
Make assertions on job created in test

russcam:
This commit updates the GetJobStatsApiTests to make
assertions on the job configured for the integration test.
This avoids grabbing a random job that may exist in the cluster
at the time the test concurrently runs.

The machine memory available to ML is also increased from
the default 30% to 50%, to avoid exceptions in running out
of memory, which have occurred locally on a machine with
32GB RAM when running all integration tests.


Mpdreamz:
Test failures unrelated:

- CSV #4666 should be fixed not that the PR landed
- Cluster processing timeout #4011

PR #4673, opened 2020-04-24
Allow HttpClient instances to be refreshed periodically

Mpdreamz:
This PR fixes a longstanding wish to be able to recycle `HttpClient` instances periodically. We reuse HttpClient/HttpMessageHandler's to prevent TCP leakage but  [infamously this does not take DNS changes into account](https://github.com/dotnet/runtime/issues/18348).

`IHttpClientFactory` from [Microsoft.Extensions.Http](https://www.nuget.org/packages/Microsoft.Extensions.Http/) solves this but the nuget package depends on too many things we don't want to depend on OOTB like Dependency Injection.  This copies parts of this code (now MIT licensed as of February) so we can reuse the same strategy in the client. 

Starting with `netstandard2.1` `SocketsHttpHander` also supports `PooledConnectionLifetime` this however has a bug in corrolation with [`MaxConnectionsPerServer`](https://github.com/dotnet/corefx/pull/29822). This is fixed in `netcoreapp3.0` and up. 

The goal is to ship this new `RequestDataHttpClientFactory` introduced here in all TFM's. 

In a follow up commit we'll introduce `netcoreapp3.0` and `netcoreapp3.1` TFM's that will only use `RequestDataHttpClientFactory` to create `HttpClient` instances but won't use its recycling capabilities since that's now handled by `PooledConectionLifetime`.

### DIFF
--- a/src/Elasticsearch.Net/Connection/HandlerTracking/ActiveHandlerTrackingEntry.cs
+++ b/src/Elasticsearch.Net/Connection/HandlerTracking/ActiveHandlerTrackingEntry.cs
@@ -9,8 +9,11 @@ using System.Threading;
 
 namespace Elasticsearch.Net
 {
-    // Thread-safety: We treat this class as immutable except for the timer. Creating a new object
-    // for the 'expiry' pool simplifies the threading requirements significantly.
+	/// <summary>
+    /// Thread-safety: We treat this class as immutable except for the timer. Creating a new object
+	/// for the 'expiry' pool simplifies the threading requirements significantly.
+	/// <para>https://github.com/dotnet/runtime/blob/master/src/libraries/Microsoft.Extensions.Http/src/ActiveHandlerTrackingEntry.cs</para>
+	/// </summary>
     internal class ActiveHandlerTrackingEntry
     {
         private static readonly TimerCallback TimerCallback = (s) => ((ActiveHandlerTrackingEntry)s).Timer_Tick();
--- a/src/Elasticsearch.Net/Connection/HandlerTracking/ActiveHandlerTrackingEntry.cs
+++ b/src/Elasticsearch.Net/Connection/HandlerTracking/ActiveHandlerTrackingEntry.cs
@@ -1,0 +1,81 @@
+﻿﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if DOTNETCORE
+using System;
+using System.Diagnostics;
+using System.Threading;
+
+namespace Elasticsearch.Net
+{
+    // Thread-safety: We treat this class as immutable except for the timer. Creating a new object
+    // for the 'expiry' pool simplifies the threading requirements significantly.
+    internal class ActiveHandlerTrackingEntry
+    {
+        private static readonly TimerCallback TimerCallback = (s) => ((ActiveHandlerTrackingEntry)s).Timer_Tick();
+        private readonly object _lock;
+        private bool _timerInitialized;
+        private Timer _timer;
+        private TimerCallback _callback;
+
+        public ActiveHandlerTrackingEntry(
+            int key,
+            LifetimeTrackingHttpMessageHandler handler,
+            TimeSpan lifetime)
+        {
+            Key = key;
+            Handler = handler;
+            Lifetime = lifetime;
+
+            _lock = new object();
+        }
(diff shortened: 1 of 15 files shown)


Mpdreamz:
I finally found where to remove `travis-ci` which was forever pending as we disabled it! It was a required check for master in the branch rules 🎉

src/Elasticsearch.Net/Connection/HandlerTracking/RequestDataHttpClientFactory.cs
```diff
@@ -0,0 +1,254 @@
...
+using System.Linq;
+using System.Net.Http;
+using System.Threading;
+using System.Xml;
+using Elasticsearch.Net;
+
+namespace Elasticsearch.Net
+{
+	/// <summary>
+	/// Heavily modified version of DefaultHttpClientFactory, repurposed for RequestData
```
russcam, line 19:
```suggestion
	/// Heavily modified version of DefaultHttpClientFactory, re-purposed for RequestData
```

minor nit

src/Elasticsearch.Net/Connection/HandlerTracking/RequestDataHttpClientFactory.cs
```diff
@@ -0,0 +1,254 @@
...
+	/// </summary>
+	internal class RequestDataHttpClientFactory : IDisposable
```
russcam, line 21: Can probably just call it `HttpClientFactory`?

src/Elasticsearch.Net/Connection/HandlerTracking/ValueStopWatch.cs
```diff
@@ -0,0 +1,40 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
russcam, line 3: Could we add the original source file URI?

src/Elasticsearch.Net/Connection/HandlerTracking/RequestDataHttpClientFactory.cs
```diff
@@ -0,0 +1,254 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
russcam, line 3: Could we add the original source file URI?

src/Elasticsearch.Net/Connection/HandlerTracking/LifetimeTrackingHttpMessageHandler.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
russcam, line 3: Could we add the original source file URI?

src/Elasticsearch.Net/Connection/HandlerTracking/ExpiredHandlerTrackingEntry.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
russcam, line 3: Could we add the original source file URI?

src/Elasticsearch.Net/Connection/HandlerTracking/ActiveHandlerTrackingEntry.cs
```diff
@@ -0,0 +1,81 @@
+﻿﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
russcam, line 3: Could we add the original source file URI?

src/Elasticsearch.Net/Connection/HandlerTracking/RequestDataHttpClientFactory.cs
Mpdreamz: I renamed it to signal it's been re-purposed specifically for `RequestData`. It's not the same as `DefaultHttpClientFactory`. 

Keen to keep the name less generic.

PR #4674, opened 2020-04-26
Fix #4298 be less strict about `Operations` on `IBulkRequest`

### LINKED ISSUE #4298: BulkOperationsCollection Allocation Improvements?
**Context:**

Currently, `BulkOperationsCollection<T>` is implemented using a `private IList<T>`, along with some locking. The summary tag explains this:

>  This class is used by \<see cref="IBulkRequest.Operations" /\> which needs thread safe adding \<see cref="ICollection{T}.Add" /\> as well as an equivalent of \<see cref="List{T}.AddRange"/\>. Because operations from Elasticsearch are executed in order none of the types in System.Collection.Concurrent can't be used for this. We need to preserve insert order and exposed indexed index because \<see cref="BulkResponse.Items"/\> is ordered and lines up with \<see cref="BulkRequest.Operations"/\> allowing one to zip the two together.

That's fairly concise and specific, but I'm a bit confused by it.

It's pretty trivial in the scheme of things, but I can't help but notice the limitation this propriety type imposes.

**Questions:** 

1. What exactly is it about `IBulkRequest.Operations` that makes this thread-safe adding necessary? It's a nice enough helper class in general, but isn't it true that by the time it gets to the request itself, it's going to be effectively read-only? I wouldn't expect someone to add to the collection after they've called `BulkAsync`.

Here's the formatter, showing that it's just looping over the contents.

https://github.com/elastic/elasticsearch-net/blob/c646b98c5e5278ad122cd686fa27c45cbb18b919/src/Nest/Document/Multiple/Bulk/BulkRequestFormatter.cs#L26

2. Pending that, would it be possible to change `IBulkRequest.Operations` to be an `IList<IBulkOperation>`? Existing implementations should remain in-tact, but I could use (and re-use) an `IBulkOperation[]`.
3. If that's not possible, would it be possible to at least expose the `List<T>(int capacity)` constructor, so that I can avoid some unnecessary list expansion?

Mpdreamz:
`IBulkRequest` dictated that operations should be implemented using `BulkOperationsCollection` which effectively is a `SynchronizedCollection`. We did this back in 2013 because concurrent producers calling `.IndexMany()` were expected to be threadsafe. 

Since then we also introduced the object inititalizer syntax. Here we can push back on this requirement. This PR fixes that so that users using that syntax are not forced to use the `BulkOperationsCollection`.

Some performance numbers attached:

|                  Method |     Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------ |---------:|----------:|----------:|------:|------:|------:|----------:|
|          Descriptors.V8 | 2.581 ms | 0.0529 ms | 0.1383 ms |     - |     - |     - |    1.7 MB |
|             Descriptors | 2.722 ms | 0.0818 ms | 0.2308 ms |     - |     - |     - |    1.7 MB |
| BulkCollection.AddRange | 1.173 ms | 0.0317 ms | 0.0936 ms |     - |     - |     - |   1.09 MB |
|    BulkCollection.const | 1.096 ms | 0.0217 ms | 0.0583 ms |     - |     - |     - |   1.09 MB |
|              'new List' | 1.061 ms | 0.0209 ms | 0.0468 ms |     - |     - |     - |   1.09 MB |

### DIFF
--- a/src/Nest/Document/Multiple/Bulk/BulkRequest.cs
+++ b/src/Nest/Document/Multiple/Bulk/BulkRequest.cs
@@ -25,7 +25,7 @@ namespace Nest
 		IList<IBulkOperation> IBulkRequest.Operations
 		{
 			get => _operations;
-			set { }
+			set => throw new NotImplementedException($"{nameof(BulkDescriptor)} does not allow {nameof(IBulkRequest)}.{nameof(IBulkRequest.Operations)} to be set directly");
 		}
 
 		public BulkDescriptor Create<T>(Func<BulkCreateDescriptor<T>, IBulkCreateOperation<T>> bulkCreateSelector)
--- a/src/Nest/Document/Multiple/Bulk/BulkRequest.cs
+++ b/src/Nest/Document/Multiple/Bulk/BulkRequest.cs
@@ -10,17 +10,23 @@ namespace Nest
 	public partial interface IBulkRequest
 	{
 		[IgnoreDataMember]
-		BulkOperationsCollection<IBulkOperation> Operations { get; set; }
+		IList<IBulkOperation> Operations { get; set; }
 	}
 
 	public partial class BulkRequest
 	{
-		public BulkOperationsCollection<IBulkOperation> Operations { get; set; }
+		public IList<IBulkOperation> Operations { get; set; }
 	}
 
 	public partial class BulkDescriptor
 	{
-		BulkOperationsCollection<IBulkOperation> IBulkRequest.Operations { get; set; } = new BulkOperationsCollection<IBulkOperation>();
+		private readonly BulkOperationsCollection<IBulkOperation> _operations = new BulkOperationsCollection<IBulkOperation>();
+
+		IList<IBulkOperation> IBulkRequest.Operations
+		{
+			get => _operations;
+			set { }
+		}
 
 		public BulkDescriptor Create<T>(Func<BulkCreateDescriptor<T>, IBulkCreateOperation<T>> bulkCreateSelector)
 			where T : class =>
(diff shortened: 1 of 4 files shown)


src/Nest/Document/Multiple/Bulk/BulkRequest.cs
```diff
@@ -10,17 +10,23 @@ namespace Nest
...
 
 	public partial class BulkDescriptor
 	{
-		BulkOperationsCollection<IBulkOperation> IBulkRequest.Operations { get; set; } = new BulkOperationsCollection<IBulkOperation>();
+		private readonly BulkOperationsCollection<IBulkOperation> _operations = new BulkOperationsCollection<IBulkOperation>();
+
+		IList<IBulkOperation> IBulkRequest.Operations
+		{
+			get => _operations;
+			set { }
```
russcam, line 28: I think we should either:

1. `throw new InvalidOperationException()` here, indicating that it cannot be set.

or

2. make _operations an `IList<IBulkOperation>`, initialize a new instance of `new BulkOperationsCollection<IBulkOperation>()` on the getter if the backing _operations is null, and allow someone to set _operations with the setter

Mpdreamz: We should! Will go for option 1.

Mpdreamz:
> Do we want to also expose a way to set the IList<T> type on BulkDescriptor?

Good call out will amend today!

Mpdreamz:
> Do we want to also expose a way to set the IList type on BulkDescriptor? 

Revisiting this I reckon its not worth the hassle, it would require introducing a constructor argument which breaks our selector rules or a complicated `.InitCollection()` method.

PR #4675, opened 2020-04-26
Fix F# project warnings

Mpdreamz:


Mpdreamz:
Failing integration test is covered by: #4674

build/scripts/Building.fs
```diff
@@ -18,7 +17,7 @@ module Build =
 
     let Restore() = DotNet.Exec ["restore"; Solution; ] |> ignore
         
-    let Compile args version = 
+    let Compile _ version =
```
russcam, line 20: I think the _args_ parameter `_` can just be removed here?

PR #4211, opened 2019-11-11
Add System.Text.Json support

Mpdreamz:
Blocked on: dotnet/corefx#42533

This PR adds initial support for a `System.Text.Json` backed `IElasticsearchSerializer`.

PR #4679, opened 2020-04-28
Make low level client use System.text.json

Mpdreamz:
re-targeting #4211 to `master`

This introduces an `IElasticsearchSerializer` implementation for `System.Text.Json`.

TODO:

- [ ] Remove `LowLevelRequestResponseSerializer`
- [ ] Move `Utf8Json` to `Nest` 
- [x] Make sure formatters the lowlevel client relies on are ported over
  - [x] `DynamicDictionaryFormatter`
  

Long term plan is for `Nest` to move over to `System.Text.Json` as well but this will require significant investment in porting formatters over.


Mpdreamz:
Kick this off again to see why `clients-ci` was failing.

russcam:
rebased against master

russcam:
canary packages are currently failing to be built for this because ILRepack cannot resolve the netstandard20 System.Text.Json assembly when running the "merge" for netstandard21. I've opened https://github.com/nullean/assembly-rewriter/pull/7 to allow AssemblyRewriter to sign the rewritten versioned canary packages, which would allow ILRepack to be removed, since it is only being used to sign the assembly in the master and 7.x branches.

tests/Tests/Framework/SerializationTests/ExceptionSerializationTests.cs
```diff
@@ -28,7 +28,8 @@ public ExceptionSerializationTests()
 			_elasticsearchNetSerializer = lowlevelClient.Serializer;
 		}
 
-		[U]
+		//TODO Needs exception formatter
+		[U(Skip="Needs Exception formatter")]
```
russcam, line 32: should this be addressed before merging this in?

Mpdreamz: Yep!

russcam: Done!

Mpdreamz:
@russcam this is starting to look good to me! Some updated on this on my end

* `clients-ci` was silently failing open #5024 to remedy this on all integration branches.
* Rebased this PR, `clients-ci` started failing because `YamlMap` is `Dict<object, object>` which `System.Text.Json` [does not support](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to#dictionary-with-non-string-key). Added an additional converter for it.
* `clients-ci` is now failing on 1 test where the its asserting `{ value: 5.0 }` where `{ value: 5 }` is returned. We have some leniency if numbers are compared directly but we do not (and should not) recursively patch assertion values. This is related to the yaml test failures you identified in the rust client. We don't need to solve this on this PR IMO. 
* Noticed some assertions in `PostData.doc.cs` were commented out (pretty sure I did that), uncommented them again and everthing was still green.



If you have time to:

* ignore that test for now
* run the docs generator to fix the stale docs check

I'm happy if we merge this to `master` 🎉

Mpdreamz:
ResponseBuilder was asking the Utf8Json serializer to deserialize to DynamicDictionary. 
I added a new System.Text.Json converter to handle this now.

This also updates `DynamicValue` to deal with the fact that deserializing to `Dictionary<string, object>` does not materialize `object` these remain `JsonElement`.

Mpdreamz:
The failing integration tests relate to `boost` being available on some properties. 

Merging this in!

Mpdreamz:
Will follow up with a PR to:

- Remove LowLevelRequestResponseSerializer
- Move Utf8Json to Nest

PR #4705, opened 2020-05-25
update CatAllocationRecord

### LINKED ISSUE #4699: Cat.Allocation field mismatch
**NEST/Elasticsearch.Net version**:7.7.0

**Elasticsearch version**:7.7.0

When Cat.Allocation is called, no information about the allocation is actually returned. I think that the mapping in [this file](https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/Cat/CatAllocation/CatAllocationRecord.cs) is wrong/outdated.
From what I receive from Elasticsearch, the name should be separed with a dot instead of camel case: e.g. "disk.avail" instead of "diskAvail".

**Steps to reproduce**:
 1.Just call client.Cat.Allocation() and inspect the response.

**Expected behavior**
I expect the API call to return the allocation information.

russcam:
This commit updates the fields available on CatAllocationRecord
to match what is returned from Elasticsearch.

Fixes #4699

### DIFF
--- a/src/Nest/Cat/CatAllocation/CatAllocationRecord.cs
+++ b/src/Nest/Cat/CatAllocation/CatAllocationRecord.cs
@@ -2,6 +2,7 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
+using System;
 using System.Runtime.Serialization;
 
 namespace Nest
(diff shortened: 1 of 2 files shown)


russcam:
requires manual backport to 7.x, 7.7 and 6.x in order to obsolete `DiskRatio`

Mpdreamz:
Failed build is due to the fact that the appveyor feed for `elasticsearch-net-abstractions` is reset an only has the renamed packages. Updating the branch should pick up changes that went into all integration branches to reflect the renames there.

PR #4706, opened 2020-05-25
Update abstractions

Mpdreamz:
In preperation for the release of the packages in https://github.com/elastic/elasticsearch-net-abstractions/

The packages in that repos have been renamed to be more indicative of their use.

e.g `Elastic.Xunit` was too broad, this is now `Elastic.Elasticsearch.Xunit


Mpdreamz:
`clients-ci` is known broken, will follow up with a fix.

PR #4708, opened 2020-05-25
Fix/7.7/update abstraction

Mpdreamz:
Manual backport of #4706


Mpdreamz:
InvalidSearchApi tests always returns 400 now on `latest-7` will follow up in a new PR since that change requires porting to other branches too.

PR #4712, opened 2020-05-25
regenerate low level client from latest 7.x specification

Mpdreamz:


Mpdreamz:
clients-ci happy I'm happy. Merging this in. 

@russcam something to discuss tomorrow is there are some breaking changes in this one in the lowlevel client on API's we document as potentially breaking in minors.

PR #4713, opened 2020-05-26
Add special serialization handling for nullable ValueTuples

### LINKED ISSUE #4703: SIGSEGV when indexing nullable value tuple with value
7.6.1/7.7

**Elasticsearch version**:
7.6.2

**Runtime/OS info**
dotnet 3.1.201 on Mac OS X 10.14.4 (BuildVersion: 18E226)

Indexing a document that contains a property that is a nullable value tuple such as `(string info, int number)?` with a value that is not null produces a `SIGSEGV` and the whole application crashes.

**Steps to reproduce**:
 
The issue can be reproduced with the following code: 
```cs
using System;
using Nest;

public class ExampleDoc {
	public (string info, int number)? tupleNullable { get; set; }
}

public class Program {
	public static void Main() {
		var node = new Uri("http://localhost:9200");
		var client = new ElasticClient(new ConnectionSettings(node));
		
		var doc = new ExampleDoc {
			tupleNullable = ("somestring", 42),
		};

                var indexName= "index4";
                client.Indices.Create(indexName, index => index
                        .Map<ExampleDoc>(m => m
                        .AutoMap()
                ));

               client.Index(doc, i => i.Index(indexName));
	}
}
```

**Expected behavior**
The expected bahavior is to either throw an exception if this usecase is not supported or to successfully index the document

russcam:
This PR adds special serialization handling for nullable `ValueTuple`s. Specialized handling is required as without, a nullable `ValueTuple` type ends up falling into `CustomDynamicObjectResolver` inside of
`NestFormatterResolver`, which will build a dynamic formatter using `MetaType` et.al, which enumerates interfaces and can cause a CLR error when calling `System.Runtime.CompilerServices.ITuple.get_Length()`.

Fixes #4703

### DIFF
--- a/src/Elasticsearch.Net/Utf8Json/Resolvers/DynamicGenericResolver.cs
+++ b/src/Elasticsearch.Net/Utf8Json/Resolvers/DynamicGenericResolver.cs
@@ -229,6 +229,45 @@ namespace Elasticsearch.Net.Utf8Json.Resolvers
 
 					return CreateInstance(tupleFormatterType, ti.GenericTypeArguments);
 				}
+
+				// Nullable ValueTuple
+				else if (isNullable && nullableElementType.GetTypeInfo().IsConstructedGenericType() &&
+					nullableElementType.GetTypeInfo().FullName.StartsWith("System.ValueTuple"))
+				{
+					Type tupleFormatterType = null;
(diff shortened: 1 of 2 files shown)


russcam:
rebased on top of master

PR #4714, opened 2020-05-26
Fix api url lookups

russcam:
This PR fixes a couple of bugs in the ApiUrlsLookup generator:

- use the endpoint filename when comparing against the list
of high level APIs to ignore.
- a bug in the CSharp method name generation
where Index is removed from the name of the SimulateIndexTemplate
method in the Indices namespace. Removing this causes a collision
with SimulateTemplate


russcam:
ported to master in eed3a385bb3911a475b3e6524d2dda6adaaa7115 
ported to 7.7 in 3afd3c2

PR #4719, opened 2020-05-29
Update REST api specs against 7.8 branch

russcam:


src/ApiGenerator/RestSpecification/Core/indices.get_data_streams.json
```diff
@@ -1,5 +1,5 @@
 {
-  "indices.get_data_stream":{
+  "indices.get_data_streams":{
```
russcam, line 2: @Mpdreamz looks like the singular is coming in in 7.9.0: https://github.com/elastic/elasticsearch/pull/56762. Think this should be patched in the spec here in 7.8

Mpdreamz: ++ don't think we have to but here we can use the foresight to our advantage! Let's patch.

russcam:
failures aren't related to changes - will look to address separately.

Issue #4718, opened 2020-05-29
7.8.0 Meta ticket


russcam:
This meta ticket tracks the changes in Elasticsearch 7.8.0 that require changes in the client:

### New APIs

- [x] Cluster Delete VotingConfigExclusions 
- [x] Cluster Post VotingConfigExclusions 

#### Cluster Coordination
- [x] Voting config exclusions should work with absent nodes [#50836](https://github.com/elastic/elasticsearch/pull/50836) (issue: [#47990](https://github.com/elastic/elasticsearch/issues/47990))

#### Infra/Core
- [x] Deprecate node local storage setting [#54374](https://github.com/elastic/elasticsearch/pull/54374)

#### Aggregations

- [x] Add support for filters to T-Test aggregation [#54980](https://github.com/elastic/elasticsearch/pull/54980) (issue: [#53692](https://github.com/elastic/elasticsearch/issues/53692))
- [x] Add Student's t-test aggregation support [#54469](https://github.com/elastic/elasticsearch/pull/54469) (issue: [#53692](https://github.com/elastic/elasticsearch/issues/53692))

#### Features/Indices APIs
- ~[ ] Add simulate template composition API _index_template/_simulate_index/{name}  [#55686](https://github.com/elastic/elasticsearch/pull/55686) (issue: [#53101](https://github.com/elastic/elasticsearch/issues/53101))~ Simulate index template APIs are experimental

#### Geo

- [x] Add new point field. [#53804](https://github.com/elastic/elasticsearch/pull/53804)

#### Aggregations
- [x] Expose agg usage in Feature Usage API [#55732](https://github.com/elastic/elasticsearch/pull/55732) (issue: [#53746](https://github.com/elastic/elasticsearch/issues/53746))

#### Analysis
- [x] Add preserve_original setting in ngram token filter [#55432](https://github.com/elastic/elasticsearch/pull/55432)

#### Cluster Coordination
- [x] Add voting config exclusion add and clear API spec and integration test cases [#55760](https://github.com/elastic/elasticsearch/pull/55760) (issue: [#48131](https://github.com/elastic/elasticsearch/issues/48131))

#### Features/CAT APIs
- ~[ ] Add support for V2 index templates to /_cat/templates [#55829](https://github.com/elastic/elasticsearch/pull/55829) (issue: [#53101](https://github.com/elastic/elasticsearch/issues/53101))~ v2 index template APIs are experimental

#### Features/Indices APIs

[...]

PR #4727, opened 2020-06-02
Add PreserveOriginal to EdgeNgramTokenFilter

russcam:
Relates: #4718


src/Nest/Analysis/TokenFilters/EdgeNGram/EdgeNGramTokenFilter.cs
```diff
@@ -31,6 +31,13 @@ public interface IEdgeNGramTokenFilter : ITokenFilter
 		/// </summary>
 		[DataMember(Name ="side")]
 		EdgeNGramSide? Side { get; set; }
+
+		/// <summary>
+		/// Emits original token when set to <c>true</c>. Defaults to <c>false</c>.
```
russcam, line 36:
```suggestion
		/// Emits original token when set to <c>true</c>. Defaults to <c>false</c>.
		/// <para />
		/// Available in Elasticsearch 7.8.0+
```

PR #4735, opened 2020-06-04
Add Point property for mapping point datatype

russcam:
Relates: #4718, elastic/elasticsearch#53804

This commit adds a Point property to map the new point data type.
A CartesianPoint type is introduced to model points.

### DIFF
--- a/src/Nest.JsonNetSerializer/Converters/HandleNestTypesOnSourceJsonConverter.cs
+++ b/src/Nest.JsonNetSerializer/Converters/HandleNestTypesOnSourceJsonConverter.cs
@@ -22,7 +22,8 @@ namespace Nest.JsonNetSerializer.Converters
 			typeof(ILazyDocument),
 			typeof(LazyDocument),
 			typeof(GeoCoordinate),
-			typeof(GeoLocation)
+			typeof(GeoLocation),
+			typeof(CartesianPoint),
 		};
 
 		private readonly IElasticsearchSerializer _builtInSerializer;
--- a/src/Nest/Mapping/DynamicTemplate/SingleMapping.cs
+++ b/src/Nest/Mapping/DynamicTemplate/SingleMapping.cs
@@ -55,6 +55,10 @@ namespace Nest
 			selector?.Invoke(new ShapePropertyDescriptor<T>());
 
 		/// <inheritdoc />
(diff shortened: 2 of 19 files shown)


src/Nest/Mapping/AttributeBased/ElasticsearchCorePropertyAttributeBase.cs
```diff
@@ -2,7 +2,7 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
-﻿using System;
+ using System;
```
russcam, line 5: @Mpdreamz, Rider is automatically adding this space in when a file is opened. It looks like it's related to the introduction of the license header.

russcam:
The new point datatype can be used in shape queries. Shape queries however are bound to `IGeoShape`

https://github.com/elastic/elasticsearch-net/blob/f1ffbf260132eb83ea4e7950c1bcf19ef679ee6f/src/Nest/QueryDsl/Specialized/Shape/IShapeQuery.cs#L34-L38

which `CartesianPoint` does not implement. Think we may need to introduce new shape types here

russcam:
After conversation with the geo team, it's clarified some questions around `CartesianPoint`:

`CartesianPoint` can be used for the `null_value` of a Point property field mapping, and allow users to use it on their documents to index into the Point field. To query on a Point field using the shape query, the `PointGeoShape` type can be used.

Another type (`CartesianPoint`) needs to be introduced because of the formats that must be supported for both the `null_value` of Point field mappings, and also for formats that may be used on `_source` documents **outside** of the client that the client must be able to deserialize:

- Object: `{ "x": 41.12, "y": -71.34 }`
- String: `"41.12,-71.34"`
- Array: `[41.12,-71.34]`
- WKT: `POINT (41.12 -71.34)`

russcam:
This PR has been rebased against master and is ready for review, @Mpdreamz.

I think we should seriously consider renaming `IGeoShape` to `IGeometry` in 8.x, and the shape types that implement `IGeoShape` to remove the `Geo` name part. The renames better reflect their intent and usage, though I do appreciate that such a change could be a reasonably large breaking change, as geo queries are popular.

russcam:
Merging this one in; last test run was green and this test run fails with tests unrelated to the changes.

PR #4738, opened 2020-06-04
Add Voting Config Exclusions APIs

russcam:
Relates: #4718, elastic/elasticsearch#55760

This commit adds the voting config exclusions APIs to the high level client.
Custom response builder types are required because the APIs return
\n as the response, which the client will attempt to deserialize to a
response object.


Mpdreamz:
Do we have an issue on the elasticsearch repos tracking the empty new line?

russcam:
Yep, here's the Elasticsearch issue: https://github.com/elastic/elasticsearch/issues/57639

PR #4744, opened 2020-06-09
Add script_cache to Node Stats API

russcam:
Relates: #4718

This commit adds the script_cache to node stats
API response.

### DIFF
--- a/src/Nest/Cluster/NodesStats/NodeStats.cs
+++ b/src/Nest/Cluster/NodesStats/NodeStats.cs
@@ -62,6 +62,13 @@ namespace Nest
 		[DataMember(Name = "script")]
 		public ScriptStats Script { get; internal set; }
 
+		/// <summary>
+		/// Available in Elasticsearch 7.8.0+
+		/// </summary>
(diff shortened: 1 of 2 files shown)


src/Nest/Cluster/NodesStats/NodeStats.cs
```diff
@@ -85,6 +92,9 @@ public class ScriptStats
 
 		[DataMember(Name = "compilations")]
 		public long Compilations { get; internal set; }
+
+		[DataMember(Name = "compilation_limit_triggered")]
```
Mpdreamz, line 96: Just double checking this always comes back as `0` from the server?

russcam: The value returned is always `0`? Might be tricky to have a valid greater than `0` returned in tests.

The server side defines the property here

https://github.com/elastic/elasticsearch/blob/32f46f2214808c0b2c2271bf977f2f08fe6d6fbc/server/src/main/java/org/elasticsearch/script/ScriptCacheStats.java#L108

Issue #4781, opened 2020-06-17
[FLAKY TEST] Use null_value that conforms to format in Date[Nanos]Property tests


russcam:
Relates: https://github.com/elastic/elasticsearch-net/pull/4772/checks?check_run_id=775397697

The value passed for `null_value` on DatePropertyTests and DateNanosPropertyTests must conform to the format used

PR #4789, opened 2020-06-18
Ignore ContainsMetadata field when automapping Attachment

russcam:
This commit adds `[Ignore]` attribute to the `ContainsMetadata`
field to ignore the field when automapping an Attachment.

`[IgnoreDataMember]` ignores the value when serializing an
instance of Attachment, but `NoopPropertyVisitor` does not
take into account `[IgnoreDataMember]` when mapping properties.


russcam:
Opened #4781 for failing tests. Merging this in

PR #4792, opened 2020-06-18
Change default ModelSnapshotRetentionDays

russcam:
This commit fixes assertions on ModelSnapshotRetentionDays
as the default days has changed from 1 to 10.


russcam:
Opened #4781 for Date property tests. Merging this in

PR #4793, opened 2020-06-18
Remove duplicated properties

DanAtkinson:
Changes to remove duplicates on https://github.com/elastic/elasticsearch-net/blob/master/docs/client-concepts/connection/configuration-options.asciidoc.

### DIFF
--- a/docs/client-concepts/connection/configuration-options.asciidoc
+++ b/docs/client-concepts/connection/configuration-options.asciidoc
@@ -30,18 +30,6 @@ the low level and high level client:
 
 Api Key to send with all requests to Elasticsearch
 
-`ApiKeyAuthentication`::
-
-Api Key to send with all requests to Elasticsearch
-
-`ApiKeyAuthentication`::
-
-Api Key to send with all requests to Elasticsearch
-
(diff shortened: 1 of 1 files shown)


russcam:
Thanks for opening @DanAtkinson. This change is probably better handled in the DocsGenerator, to deduplicate the overloaded methods when generating the asciidoc labelled list. Would you like to tackle this? I _think_ it would just need a change in

https://github.com/elastic/elasticsearch-net/blob/1eb3928c82fb9b3364505c865970244a941dc217/src/DocGenerator/AsciiDoc/GeneratedAsciidocVisitor.cs#L313

to distinct by Label too.

DanAtkinson:
@russcam Thanks. I'll take a look. I haven't used the DocsGenerator but it _should_ be fairly straightforward to run, change, test, and verify.

russcam:
Doc generation can be run from the build script in the root of the repository

```
.\build documentation
```

russcam:
Hi @DanAtkinson, I went ahead and made the change to  `GeneratedAsciidocVisitor`, and merged this in. Thanks again 👍

PR #4798, opened 2020-06-22
Remove MapsApi from DeleteExpiredData overload

russcam:
This commit removes the MapsApiAttribute from the DeleteExpiredData overloads, so that the overload with a body are used for yaml tests


russcam:
Failing tests are related to #4781. Merging this in.

PR #4819, opened 2020-06-25
[PR] The proj files have been updated to enable SourceLink

JTOne123:
CSProj files have been updated to enable SourceLink in your nuget
---

*[This pull request was created with an automated workflow]*

I noticed that your repository and Nuget package are important for our .NET community, but you still haven't enabled SourceLink.

**We have to take 2 steps:**
1) Please approve this pull request and make .NET a better place for .NET developers and their debugging.
2) **Then just upload the .snupkg file** to https://www.nuget.org/ (now you can find the snupkg file along with the .nuget file)

You can find more information about SourceLine at the following links  
https://github.com/dotnet/sourcelink
https://www.hanselman.com/blog/ExploringNETCoresSourceLinkSteppingIntoTheSourceCodeOfNuGetPackagesYouDontOwn.aspx

If you are interesting about this automated workflow and how it works  
https://github.com/JTOne123/GitHubMassUpdater

*If you notice any flaws, please comment and I will try to make fixes manually*

### DIFF
--- a/.ci/Jenkins.csproj
+++ b/.ci/Jenkins.csproj
@@ -2,6 +2,15 @@
 
     <PropertyGroup>
         <TargetFramework>netcoreapp2.2</TargetFramework>
-    </PropertyGroup>
+    
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
+  </ItemGroup>
+
 
 </Project>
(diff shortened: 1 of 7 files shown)


russcam:
Hi @JTOne123,  The PDBs are shipped with the nuget packages, and source linking is performed in a [`_PublishArtifacts.Build.props`](https://github.com/elastic/elasticsearch-net/blob/3b239bfee536ff74c7bf74e5ea1e93758a712f5e/src/_PublishArtifacts.Build.props) that is included in the build of projects that create nuget packages. Looking at the nuget packages, as far as I can see, Source linking is enabled

image

This might have come up as a false positive because the automated project is only looking for snupkgs: https://github.com/JTOne123/GitHubMassUpdater/blob/bfa009f9fbd1ec1aa82f02c85239582b018ae63e/GitHubSearch/Program.cs#L144

russcam:
Thought I'd check to see where the embed PDBs vs. snupkgs discussion is, and it looks like snupkgs may still not be support on all Nuget servers (https://github.com/dotnet/sourcelink/issues/255). Since we also publish packages to AppVeyor (and looking at Feedz), embedding PDBs still looks like the way to go for now.

JTOne123:
yes, you are right, only on nuget.org and this is bad

but in case of snuget, it will split into 2 files and size will be decreased of original nuget pack


so, in your case, I'll close the PR

PR #4838, opened 2020-07-01
Introduce .ci/make.sh

Mpdreamz:
Initial stab at common entry script for all the Elasticsearch clients.

This puts our build through a docker container so that folks & CI do not need
to install dependencies on their machines to build artifacts.

> ./.ci/make.sh release VERSION

will output our packages, release notes, API reports in `.ci/output`.

This also updates our build to use: https://github.com/nullean/nupkg-validator

This is the tool version of our build script targets that

- validate the dlls are build in release mode
- validate the dll version numbers
- validate the dll assemly identity
- validate the dll assembly names.

With the added benefit that the dependency on `sn` is gone which is not
available on the dotnet base docker images that we use.

`.ci/make.sh` is not done yet, will iterate on this as the flow in the
unified release manager becomes more apparant. For example right now we
bump the version after release but we might need to do so prior.

### DIFF
--- a/.ci/DockerFile
+++ b/.ci/DockerFile
@@ -3,7 +3,11 @@ FROM mcr.microsoft.com/dotnet/core/sdk:${DOTNET_VERSION} AS elasticsearch-net-bu
 
 WORKDIR /sln
 
+
 COPY ./*.sln ./nuget.config ./*.Build.props ./*.Build.targets ./
+
+COPY ./dotnet-tools.json ./
+RUN dotnet tool restore
 
 # todo standardize on Build.props as Directory.Build.props needs that form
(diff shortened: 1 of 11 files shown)


Mpdreamz:
> We could look to change this though, to lookup the previously released version, similar to what previous-nuget command does in assembly-differ.

++ I am looking to move to `minver` so we can eschew the change to global json atleast. This requires it s own PR though since it'll touch several things in our automation as you highlighted

> If this is used later for official releases, I think the release build need to commit changes to global.json and auto-label.json back to the repository

Aye I am looking to make the act of tagging/committing and publishing to package managers its own target though `release > publish`.

PR #4846, opened 2020-07-08
client.Timeout was not being set after refactoring introduced in #4673

Mpdreamz:
Specifically: https://github.com/elastic/elasticsearch-net/pull/4673/files#diff-6c7514b4a5a813d5985444b99a4f7a4fL167


Mpdreamz:
three failing integration tests are unrelated `"process_cluster_event_timeout_exception"` which is a known flakeyness in the integration suite we need to address

Mpdreamz:
@russcam merging this in while you are on PTO.

PR #4853, opened 2020-07-13
Fix script compilation settings

russcam:
Relates: https://github.com/elastic/elasticsearch/pull/58283

This commit updates integration test cluster
settings to use disable_max_compilations_rate for
7.9.0-SNAPSHOT and above, and remove using
max_compilations_rate which is deprecated (and removed in 8.0.0-SNAPSHOT)


russcam:
Failing tests are not related to this change but to unknown properties on field mappings and node stats API. Merging this in so that tests can run against master snapshots.

PR #4854, opened 2020-07-13
Fix ScriptCacheStats

russcam:
This commit fixes the response type for node stats script_cache
to a type with "sum" and "contexts" fields, to match the shape of returned JSON


russcam:
@Mpdreamz, Unfortunately this was mismapped in 7.8.0 in https://github.com/elastic/elasticsearch-net/commit/d593130fa88e96e0b2314db2aadfb621248615bb. Thoughts on what to do for 7.x

1. backport this breaking change to 7.x (since the current impl will be broken)?

or

2. implementing a custom formatter to deserialize into an `IReadOnlyDictionary<string, ScriptStats>`?

PR #4856, opened 2020-07-14
Fix DateHistogram aggregation intervals

### LINKED ISSUE #4839: DateHistogramAggregation - changing Interval to FixedInterval produces an invalid request
**NEST/Elasticsearch.Net version**: 7.8.1
**Elasticsearch version**: 7.8.0

Converting from 6.x to 7, using [NEST.7xUpgradeAssistant](https://www.nuget.org/packages/NEST.7xUpgradeAssistant/) (really helpful btw!)
When changing code from
`.Interval(DateInterval.Day) `
the comment on the Obsolete attribute implies that the equivalent call should be
`.FixedInterval(DateInterval.Day)`
but this seems to produce an invalid request? (showing only the aggs fragment):
`"aggs": {"doccountbyday": {"date_histogram": {"field": "published","fixed_interval": "day"}}}`
The Reason is
_failed to parse setting [date_histogram.fixedInterval] with value [day] as a time value: unit is missing or unrecognized_

I've figured out that the new code should be
`.FixedInterval(new Time(1, TimeUnit.Day))`
which does produce a valid request:
`{"aggs":{"doccountbyday":{"date_histogram":{"field":"published","fixed_interval":"1d"}}}`

**Expected behavior**
Either remove FixedInterval(DateInterval interval) - it's not a breaking change if it's already broken?!
or perform the conversion from DateInterval to Time within the method?

russcam:
This commit fixes FixedInterval and CalendarInterval on
DateHistogram aggregation to

1. For FixedInterval to accept only Time
2. For CalendarInterval to accept DateInterval or DateMathTime

This aligns with DateHistogramCompositeAggregationSource.

Fixes #4839

### DIFF
--- a/docs/aggregations/bucket/date-histogram/date-histogram-aggregation-usage.asciidoc
+++ b/docs/aggregations/bucket/date-histogram/date-histogram-aggregation-usage.asciidoc
@@ -32,7 +32,7 @@ Be sure to read the Elasticsearch documentation on {ref_current}/search-aggregat
 a => a
 .DateHistogram("projects_started_per_month", date => date
     .Field(p => p.StartedOn)
-    .Interval(DateInterval.Month)
+    .CalendarInterval(DateInterval.Month)
     .MinimumDocumentCount(2)
     .Format("yyyy-MM-dd'T'HH:mm:ss")
     .ExtendedBounds(FixedDate.AddYears(-1), FixedDate.AddYears(1))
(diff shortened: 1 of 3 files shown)


russcam:
This should be ported to 7.x (and possibly 7.8), but should be done in a way that doesn't break binary compatbility.

russcam:
Had an offline conversation with @Mpdreamz, and we have decided to backport this as a breaking change to 7.x and 7.8, since the current implementation is somewhat broken.

PR #4864, opened 2020-07-21
Fix integration tests

russcam:


Mpdreamz:
Soo close: 

```
  X Tests.Indices.IndexManagement.SplitIndex.SplitIndexApiTests.ReturnsExpectedResponse [703ms]
  Error Message:
   Tests.Framework.EndpointTests.ResponseAssertionException : Expected boolean to be true, but found False.
Response Under Test:
Valid NEST response built from a successful (200) low level call on PUT: /nest-initializerasync-9379535f/_split/nest-initializerasync-9379535f-target?pretty=true&error_trace=true
 - [1] HealthyResponse: Node: http://localhost:9200/ Took: 00:00:31.0320917
# Request:
{"settings":{"index.number_of_shards":8}}
# Response:
{
  "acknowledged" : true,
  "shards_acknowledged" : false,
  "index" : "nest-initializerasync-9379535f-target"
}
```

`shards_acknowldged` is extremely unlikely to be false though

PR #4894, opened 2020-07-30
[Docs] order `anchor-list` output explicitly.

Mpdreamz:
`anchor-list` output was not explicitly ordered causing issues with different itteration orders of files/directories between OS's.


russcam:
integration test failures not related to this change (will open separate issue).

PR #4913, opened 2020-08-03
Add data streams APIs

russcam:
This PR adds support for the data streams APIs to the high level client.

Rebase and merge when merging in, to allow backport branches to regenerate APIs from their respective branches, then pull in the data streams API commit.


russcam:
Looks like we have a difference in behaviour between 8.0.0-SNAPSHOT and 7.9.0-SNAPSHOT - the tests pass on 7.9.0-SNAPSHOT.

The rest spec indicates that a body is accepted: https://github.com/elastic/elasticsearch-net/blob/de60a13aed34ca7886310ec89a3bc2426219459e/src/ApiGenerator/RestSpecification/XPack/indices.create_data_stream.json#L26-L28

I'll open an issue to discuss.

russcam:
Merging this in as all integration tests have passed.

PR #4943, opened 2020-08-05
Skip DateHistogram aggregations for <7.2.0

russcam:
This commit skips running DateHistogram aggregations
for versions <7.2.0 as they use calendar_interval and
fixed_interval, which were introduced in 7.2.0.


Mpdreamz:
#4894 made its way back, we have to update the docs on all the branches though.

PR #4952, opened 2020-08-10
NameValueCollection.ToQueryString performance optimisation

### LINKED ISSUE #4951: NameValueCollectionExtensions Performance Optimisation
Would you be open to taking PR(s) for performance optimisations, specifically for cases where the NEST library is consumed within NetStandard 2.1 and NetCoreApp 2.1 apps?

I was exploring [`NameValueCollectionExtensions`](https://github.com/elastic/elasticsearch-net/blob/master/src/Elasticsearch.Net/Extensions/NameValueCollectionExtensions.cs) and one simple example is the `ToQueryString` method. By applying `Span<T>` optimisation, it's possible to reduce the allocations fairly significantly. Given this is called every time a `RequestData` instance is constructed it may have worthwhile affect for consumers.

I've created a benchmark and an initial optimised implementation with the following results:

```
|   Method |     Mean |     Error |    StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------- |---------:|----------:|----------:|------:|--------:|-------:|------:|------:|----------:|
| Original | 4.593 us | 0.1413 us | 0.4099 us |  1.00 |    0.00 | 0.2060 |     - |     - |     896 B |
|      New | 2.685 us | 0.1360 us | 0.3859 us |  0.93 |    0.12 | 0.1221 |     - |     - |     544 B |
```

This reduced allocated bytes by 39.3% and operates 42% quicker with my test data of 5 random keys and values. 

_A note regarding the execution time. My prototype takes a shortcut and uses a rented char[] from the `ArrayPool` with a minimum length of 512 chars. If there is a realistic maximum length expected for the qs, then hard-coding it rather than calculating the length each time is potentially worthwhile. With the length calculation the execution time is 4.227 us, so only slightly faster than current performance._

With my test data (where one key has a char that requires escaping), by applying custom `Uri` escaping, it may be possible to reduce a further 48 bytes.  If it is common for keys/values to require escaping then the allocation reduction could be more significant.

If you feel such allocation reductions would have value to consumers and you'd be open to such a PR, I'd happily submit this?

Example of prototype code...

```c#
internal static string ToQueryStringNew(this NameValueCollection nv)
{
    if (nv == null || nv.AllKeys.Length == 0) return string.Empty;

    var length = 1 + nv.AllKeys.Length - 1; // acct. for '?', and '&' chars
    foreach (var key in nv.AllKeys)
    {
        length += (key.Length * 3) + (nv[key].Length * 3) + 1; // worst case all escaped chars + '=' 
    }

    // possible to reduce execution time ~50% by hard-coding to 512 chars
    var buffer = ArrayPool<char>.Shared.Rent(length); 
    var bufferSpan = buffer.AsSpan();

    try
    {
        var position = 0;
        bufferSpan[position++] = '?';

        foreach (var key in nv.AllKeys)
        {
            if (position != 1)
                bufferSpan[position++] = '&';

            var escapedKey = Uri.EscapeDataString(key);
            escapedKey.AsSpan().CopyTo(bufferSpan.Slice(position));
            position += escapedKey.Length;

            var value = nv[key];
            if (!value.IsNullOrEmpty())
            {
                bufferSpan[position++] = '=';
                var escaped = Uri.EscapeDataString(value);                
                escaped.AsSpan().CopyTo(bufferSpan.Slice(position));
                position += escaped.Length;
            }
        }

        return new string(bufferSpan.Slice(0, position));
    }
    finally
    {
        ArrayPool<char>.Shared.Return(buffer);
    }
}
```

stevejgordon:
This commit optimises the internal `ToQueryString` extension method on the `NameValueCollection` type. It avoids some intermediate string allocations by using a zero-alloc buffer to format the final query string.

- Adds a test for the extension method to validate existing behaviour is unaffected by the change of implementation.
- Adds the `System.Memory` package to support `Span<T>` usage across all target frameworks. This is registered conditionally for targets before NetStandard 2.1. I had some downgrade errors when using the latest package version, so I've aligned with the existing `System.Buffers` version for the time being.

I've opened this against master although it may be valuable to back port to existing supported versions?

The exact performance improvement varies based on the underlying collection, but in all cases resulted in less allocations and executes equal to or faster than the original code.

## .NET Core 2.1 Benchmark Results

```c#
private NameValueCollection _nvc = new NameValueCollection
{
    { "q", "title:\"The Right Way\" AND mod_date:[20020101 TO 20030101]" },
    { "from", "10000" },
    { "request_cache", bool.TrueString },
    { "size", "100" }
};
```

```
|   Method |     Mean |     Error |    StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------- |---------:|----------:|----------:|------:|--------:|-------:|------:|------:|----------:|
| Original | 4.105 us | 0.0901 us | 0.2614 us |  1.00 |    0.00 | 0.3052 |     - |     - |    1288 B |
|      New | 3.652 us | 0.0730 us | 0.1898 us |  0.89 |    0.07 | 0.1755 |     - |     - |     736 B |
```

- Allocations reduced by 42.9%
- 11% faster

For smaller collections and key/value lengths, the allocation improvement can be greater. I've seen up to 77% on some test cases where no URI encoding need occur.

Closes #4951

### DIFF
--- a/src/Elasticsearch.Net/Extensions/NameValueCollectionExtensions.cs
+++ b/src/Elasticsearch.Net/Extensions/NameValueCollectionExtensions.cs
@@ -22,7 +22,9 @@ namespace Elasticsearch.Net.Extensions
 			var maxLength = 1 + nv.AllKeys.Length - 1; // account for '?', and any required '&' chars
 			foreach (var key in nv.AllKeys)
 			{
-				maxLength += 1 + (key.Length + nv[key]?.Length ?? 0) * 3; // '=' char + worst case assume all key/value chars are escaped
+				var bytes = Encoding.UTF8.GetByteCount(key) + Encoding.UTF8.GetByteCount(nv[key] ?? string.Empty);
+				var maxEncodedSize = bytes * 3; // worst case, assume all bytes are URL escaped to 3 chars
+				maxLength += 1 + maxEncodedSize; // '=' + encoded chars
 			}
 
 			// prefer stack allocated array for short lengths
@@ -30,7 +32,7 @@ namespace Elasticsearch.Net.Extensions
 			char[] rentedFromPool = null;
 			var buffer = maxLength > MaxCharsOnStack
 				? rentedFromPool = ArrayPool<char>.Shared.Rent(maxLength)
-				: stackalloc char[MaxCharsOnStack];
+				: stackalloc char[maxLength];
 
 			try
 			{
--- a/src/Elasticsearch.Net/Extensions/NameValueCollectionExtensions.cs
+++ b/src/Elasticsearch.Net/Extensions/NameValueCollectionExtensions.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information
 
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


tests/Tests/Extensions/NameValueCollectionExtensionsTests.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Collections.Specialized;
+using Elasticsearch.Net.Extensions;
+using FluentAssertions;
+using Xunit;
+
+namespace Tests.Extensions
+{
+	public class NameValueCollectionExtensionsTests
+	{
+		[Theory]
```
stevejgordon, line 11: For regular `Fact` tests I have just noticed that there is an Elastic.Elasticsearch.Xunit library and `I` class. I'm not sure if this format is supported?

Mpdreamz: Our `Tests` project can be run in both integration and unit test mode. This to avoid duplicating ceremony around testing API's. To annotate which assertions need to run when we use `[I]` and `[U]`, these do not support `MemberData` though. 

So technically this test will run in both contexts now. For now this is fine, long term we are likely in need to split of `Tests` and reserve the `[I]` and `[U]` tests for API level tests.

tests/Tests/Extensions/NameValueCollectionExtensionsTests.cs
```diff
@@ -0,0 +1,37 @@
...
+
+				new object[] { new NameValueCollection
+				{
+					{ "q", "name:john~1 AND (age:[30 TO 40} OR surname:K*) AND -city" },
+				}, "?q=name%3Ajohn~1%20AND%20%28age%3A%5B30%20TO%2040%7D%20OR%20surname%3AK%2A%29%20AND%20-city" },
+
+				new object[] { new NameValueCollection
+				{
+					{ "q", null },
+				}, "?q" }
```
Mpdreamz, line 34:
```suggestion
				}, "?q" },

				new object[] { new NameValueCollection
				{
					{ "emoji", "😅"}
				}, "?emoji=%F0%9F%98%85" }

```

stevejgordon: Good extra test. In fact I'd like to add this and another for a very unlikely case where all are 4 byte chars. I suspect the buffer may be too small which would be bad. Either it needs to start larger in the first place or grow in slow path to account. I'll test that out and amend if necessary.

stevejgordon:
@Mpdreamz Modified the code slightly to assume worst case that every byte requires 3 char encoding. This avoids overflowing the buffer, even for strings entirely made up of emojis! Added suggested test + one other for 2 byte character.

src/Elasticsearch.Net/Extensions/NameValueCollectionExtensions.cs
```diff
@@ -22,15 +22,17 @@ internal static string ToQueryString(this NameValueCollection nv)
 			var maxLength = 1 + nv.AllKeys.Length - 1; // account for '?', and any required '&' chars
 			foreach (var key in nv.AllKeys)
 			{
-				maxLength += 1 + (key.Length + nv[key]?.Length ?? 0) * 3; // '=' char + worst case assume all key/value chars are escaped
+				var bytes = Encoding.UTF8.GetByteCount(key) + Encoding.UTF8.GetByteCount(nv[key] ?? string.Empty);
+				var maxEncodedSize = bytes * 3; // worst case, assume all bytes are URL escaped to 3 chars
+				maxLength += 1 + maxEncodedSize; // '=' + encoded chars
```
Mpdreamz, line 27: I think the `1` character padding could potentially `0` when `nv[key]` is empty. Since we don't write `=`. Not worth the conditional and added complexity though!

Mpdreamz:
The integration test failures are due to changes on the SNAPSHOT version of the server that we have not addressed yet. Opened https://github.com/elastic/elasticsearch-net/issues/4957 to fix those. 

The auto label failure is due to me creating a new label after this PR was opened and applying it to this PR.


Thanks for the updates @stevejgordon!

PR #4953, opened 2020-08-12
Increase YAML test frequency to twice a day

Mpdreamz:


Mpdreamz:
Manually backported to `7.x`

PR #4972, opened 2020-08-20
Make rules in VirtualCluster explicit

Mpdreamz:
Defining rules that only apply a fixed number of times now fall over to
the next rule.

This also removes the the default once a fixed number of calls
fail/succeed.

In the past if you said calls should succeed twice they would fail
afterwards. If you said they should fail twice they would succeed
afterwards.

Now you need to explicitly set that default


src/Elasticsearch.Net.VirtualizedCluster/Audit/Auditor.cs
```diff
@@ -195,13 +198,13 @@ private Auditor AssertAuditTrails(ClientCall callTrace, int nthCall)
 		public void VisualizeCalls(int numberOfCalls)
 		{
 			var cluster = _cluster ?? Cluster();
-			var messages = new List<string>(numberOfCalls * 2);
-			for (var i = 0; i < numberOfCalls; i++)
+			var messages = new List<string>((numberOfCalls -1) * 2);
+			for (var i = 0; i <= numberOfCalls - 1; i++)
```
russcam, line 202: I'm missing the context of how this is used - can you explain what the logical change is in using `numberOfCalls - 1` (and also doing `<=` in addition)?

src/Elasticsearch.Net.VirtualizedCluster/VirtualClusterConnection.cs
```diff
@@ -218,14 +230,19 @@ int times
...
 
-			return Success<TResponse, TRule>(requestData, beforeReturn, successResponse, rule);
+			// if (rule.Succeeds && times >= state.Successes)
+			// 	return Success<TResponse, TRule>(requestData, beforeReturn, successResponse, rule);
+			// else if (rule.Succeeds) return Fail<TResponse, TRule>(requestData, rule);
+			//
+			// if (!rule.Succeeds && times >= state.Failures)
+			// 	return Fail<TResponse, TRule>(requestData, rule);
+			//
+			// return Success<TResponse, TRule>(requestData, beforeReturn, successResponse, rule);
```
russcam, line 245: can be deleted?

src/Elasticsearch.Net/Connection/InMemoryConnection.cs
```diff
@@ -1,8 +1,8 @@
-// Licensed to Elasticsearch B.V under one or more agreements.
-// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
-// See the LICENSE file in the project root for more information
-
-﻿using System;
+// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+ using System;
```
russcam, line 5: :sadpanda:

src/Elasticsearch.Net.VirtualizedCluster/Audit/Auditor.cs
Mpdreamz: None, this should be reverted 👍.

src/Elasticsearch.Net.VirtualizedCluster/VirtualClusterConnection.cs
Mpdreamz: It can and should!

src/Elasticsearch.Net/Connection/InMemoryConnection.cs
Mpdreamz: Rider is transforming `ZWNBSP` (as checked in) to spaces on linux. 

Will try and clean these up today.

PR #4978, opened 2020-08-24
Fix field value factor serialization

russcam:
Relates: https://stackoverflow.com/questions/63502401/elasticsearch-nest-incorrect-function-score-json-generation

This commit fixes the serialization of IFieldValueFactorFunction whereby
filter and weight were being serialized twice, once inside of
field_value_factor and once outside.


Mpdreamz (approved):
LGTM 👍

PR #4979, opened 2020-08-24
Use SourceSerializer when deserializing Source API response

### LINKED ISSUE #4970: .Source() did not seems to use the given SourceSerializer
7.9.0

**Elasticsearch version**:
7.6.2

We are using the JsonNetSerializer. One of our types should get a list of sub types with an abstract base class ( Event -> EventA ). For this case I have created a custom JsonConverter witch create the subtypes based on an type property. This works well with plain Json.Net and also with client.Get(). But on some places we use client.Source() and there is it where it stops working and we get:

`Elasticsearch.Net.UnexpectedElasticsearchClientException: 'generated serializer for Event does not support deserialize.'`

So it tries to deserialize the abstract base class Event instead of using the JsonConvert. I can see that the JsonConverter is not called at any time with .Source. So my guess is that .Source did not use the client.SourceSerializer. Maybe that is by design, but I did not find anything useful on the documentation. 

**Steps to reproduce**:
Not yet

**Expected behavior**
.Source() also should use the SourceSerializer and so my case should work.

russcam:
This commit uses the SourceSerializer to deserialize the source API response.

Fixes #4970

### DIFF
--- a/src/Elasticsearch.Net/Serialization/CustomResponseBuilderBase.cs
+++ b/src/Elasticsearch.Net/Serialization/CustomResponseBuilderBase.cs
@@ -8,7 +8,7 @@ using System.Threading.Tasks;
 
 namespace Elasticsearch.Net
 {
-	// TODO: Pass both request/response serializer an source serializer in 8.0
+	// TODO: Pass both request/response serializer and source serializer in 8.0
 	public abstract class CustomResponseBuilderBase
 	{
 		public abstract object DeserializeResponse(IElasticsearchSerializer builtInSerializer, IApiCallDetails response, Stream stream);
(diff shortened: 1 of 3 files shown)


Mpdreamz (approved):
LGTM 👍

src/Elasticsearch.Net/Serialization/CustomResponseBuilderBase.cs
```diff
@@ -8,6 +8,7 @@
 
 namespace Elasticsearch.Net
 {
+	// TODO: Pass both request/response serializer an source serializer in 8.0
```
russcam, line 11:
```suggestion
	// TODO: Pass both request/response serializer and source serializer in 8.0
```

Issue #3145, opened 2018-03-21
Refactor datatype mappings


russcam:
Relates #3138 

Within mappings, There are some options exposed in base types that are not valid for the inherited types e,g. `ObjectProperty` does not support `Store`, but inherits it from `CorePropertyBase`.

Revisit the inheritance structure to address supported options for each type.

russcam:
Additional example

- `BinaryProperty` and `BooleanProperty` should not have `Similarity`

stevejgordon:
Closing due to age/inactivity. Will be tackled within code-generator.

PR #4987, opened 2020-08-25
Remove use of similarity and boost from number mappings

russcam:
This commit removes use of fields that now cause a 400 response from Elasticsearch. These fields will be refactored as part of #3145


Mpdreamz (approved):
LGTM 👍

PR #4988, opened 2020-08-25
Add documentation for reindexing documents

russcam:
This commit adds documentation to explain the two different options available in the client for reindexing documents. It also
adds additional information to the Indexing documentation for working with `BulkAllObservable` effectively.


Mpdreamz (approved):
So much ❤️ for this! These are great additional resources for our users. LGTM 💯

PR #4996, opened 2020-09-02
Serialize Epoch milliseconds as a string

russcam:
This commit updates the DateTimeOffsetEpochMillisecondsFormatter
and NullableDateTimeOffsetEpochMillisecondsFormatter to serialize
epoch milliseconds as a string instead of a number. Add additional
parameters to integration tests to assert the change is valid
where used.


Mpdreamz (approved):
Lgtm, I yearn for the days when we can have one converter or even rely on the defaults though.

PR #4997, opened 2020-09-02
Fix test assertions

russcam:
This commit fixes some of the test assertions for 8.x snapshots.


Mpdreamz (approved):
LGTM, security index failure strikes again on windows integration latest

russcam:
Ok, I'll look to see if this test can be hardened further; I didn't see any failures locally.

PR #4999, opened 2020-09-02
Fix formatting conventions.

### LINKED ISSUE #4998: editorconfig braces rule not applying in Visual Studio
Not a product bug, but an observation during development. When editing C# files in Visual Studio the editor is defaulting to putting braces on the same line as class, method declarations.

This appears to be caused by the line:

```
csharp_new_line_before_open_brace = all:error
```

From what I can tell, this is an invalid value causing it not to apply at all. Changing it (as per the line below) fixes it for me.

```
csharp_new_line_before_open_brace = all
```

I can submit this in a small PR, if you agree this is appropriate?

cc @russcam / @Mpdreamz

stevejgordon:
Fixes #4998 by removing invalid severity values.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -82,30 +82,30 @@ csharp_prefer_braces = false:warning
 csharp_preferred_modifier_order = public,private,protected,internal,static,extern,new,virtual,abstract,sealed,override,readonly,unsafe,volatile,async:error
 
 # ---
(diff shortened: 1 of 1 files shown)


Mpdreamz:
LGTM :+1

Mpdreamz (approved):

PR #5008, opened 2020-09-11
Fix date_nanos null_value integration test

russcam:
This commit fixes a new failure with the date_nanos
property tests, updating the null_value to use
the unix epoch as opposed to DateTime.MinValue,
which now throws an error.


russcam:
Merging this in as all checks have passed

PR #5012, opened 2020-09-16
Create `Elastic.Transport` as isolated package.

Mpdreamz:
**WIP**

This moves the connectivity bits out of `Elasticsearch.Net` into `Elastic.Transport`

It also introduces `Elastic.SharedExtensions` which is built in isolation and link referenced in all artifacts as a way to stop the proliferation of too many copy pasted internal extension methods.

This is heavily reliant on #4211 and our goal to move `UTF8` to `Nest`.


Mpdreamz:
Rather then rebasing this PR. I am going to replay this on a new branch. A lot of commits on this branch are no longer relevant!

PR #5017, opened 2020-09-18
Mark all Visit methods virtual on NoopPropertyVisitor

russcam:
This commit marks all  Visit methods virtual on `NoopPropertyVisitor` to allow overriding only specific methods.

Add unit test to verify that methods are virtual, now and in the future.


Mpdreamz (approved):
Awesome!

PR #5020, opened 2020-09-22
Remove ILRepack

russcam:
This commit removes ILRepack from the build. ILRepack is used in
the 6.x branch to internalize Json.NET into the Nest assembly, and
was being used in 7.x and master branch **only** to sign the versioned
canary packages. Since assembly-rewriter can now sign a rewritten
assembly (works cross-platform) in nullean/assembly-rewriter#7,
there is no need for ILRepack at all.

### DIFF
--- a/build/scripts/Building.fs
+++ b/build/scripts/Building.fs
@@ -118,20 +118,11 @@ module Build =
                [mainDll]
                |> Seq.append depAssemblies
                |> Seq.map (fun dll ->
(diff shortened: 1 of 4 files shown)


russcam:
This PR is a dependency for tests passing in https://github.com/elastic/elasticsearch-net/pull/4679. Once this is merged, 4679 should pull this in

Mpdreamz (approved):
LTGM!

PR #5024, opened 2020-09-22
Fail the yaml runner if no folders are found

Mpdreamz:


Mpdreamz:
This will still fail `clients-ci` because we need to update the code on this branch will send additional PR's for that.

PR #5033, opened 2020-09-28
Serialize user defined tokenizer

fate1009:
i'am used user defined and  impl ITokenizer，but  serialize json is not expected.


Mpdreamz:
duplicate of #5039

PR #5034, opened 2020-09-28

fate1009:
custom ITokenizer implementations serialize is not expexted.


Mpdreamz:
duplicate of #5039

PR #5035, opened 2020-09-28

fate1009:
custom tokenizer serialize is not expected.


Mpdreamz:
duplicate of #5039

PR #5038, opened 2020-09-28

fate1009:
user define  ITokenizer implementations serlize is not expected.


Mpdreamz:
duplicate of #5039

PR #5041, opened 2020-09-29
Move Utf8Json to NEST

Mpdreamz:
This PR moves `Utf8Json` from `Elasticsearch.Net` to `Nest` now that the low level client depends
on `System.Text.Json` as it defaults serializer.

- Introduced dedicated converters for Error/ErrorCause/ServerError
- Copied some extensions. This is temporary the PR to move to a dedicated Transport solves this nicer.
- Some fixes to DynamicDictionary `dynamic` binding that were failing now Value is `JsonElement` not the value directly


Mpdreamz:
Integration tests are failing on `boost` on some property mappings. This is known but unrelated. Merging this one in.

PR #5045, opened 2020-10-02
Remove index level boosting

Mpdreamz:
Relates: elastic/elasticsearch#51703

Index level boosting existed in the mapping was not being used at query
time. This removes all boosting in anticipation of the full removal.


Mpdreamz:
Green CI!

PR #5048, opened 2020-10-05
[DOCS] Adds shared/attributes.asciidoc to index

szabosteve:
## Overview

This PR adds the reference of the shared/attribute.asciidoc file to the index.asciidoc, so abbreviations in the attributes file can be used across the .NET documentation.


Mpdreamz:
Doc checks are happy 👍

PR #5049, opened 2020-10-05
Separate connectivity bits to separate package Elastic.Transport

Mpdreamz:

PR #5021, opened 2020-09-22
Add longer timeout for StartTransform API integration tests

russcam:
Closes #5004

PR #5052, opened 2020-10-07
Elastic.Transport phase 2

Mpdreamz:
This is a follow up to #5049 

It renames `Elasticsearch.Net.VirtualizedCluster` to `Elastic.Transport.VirtualizedCluster`. 

This also introduces `Elastic.Transport.Tests` a lot of tests from `Tests could potentially be ported over but they serve a dual purpose as documentation. Moving them would break our docs on elastic.co. 

It further improves on making `Elastic.Transport` product agnostic:

`Node` now has no knowledge of Elasticsearchisms such as `master eligable` and the likes. Instead a tranport is loaded with an `IProductRegistration` that informs the transport whether the product supports pinging and sniffing, and if they do the implementations for the those. A product's sniff routine can add `features` to a node which is simply a `HashSet<string>` (exposed behind a `IReadOnlyCollection<string>`).  

This `IProductRegistration` also steers the default user agent `Transport` sends. 

In the previous PR, the settings was simply copied from `Elasticsearch.Net`. In order to be more future proof the transport now has its own `TransportConfiguration`. 

```
Transport => TransportConfiguration
Elasticsearch.Net => ConnectionConfiguration
Nest => ConnectionSettings
```

All of these extend each other.

```
ConnectionSettings => ConnectionConfiguration => TransportConfiguration
```

Ideally the clients configuration would  be named `NestConfiguration` and `ElasticsearchConfiguration` but that ship has sailed and would be a massive breaking change, I am not yet comfortable doing.


Mpdreamz:
Windows failure unrelated and most likely fixed by #5021

PR #5056, opened 2020-10-19
Update codegeneration for master and make sure we replace stashes in strings before creating assertions

Mpdreamz:


Mpdreamz:
clients-ci is happy. Merging this in.

PR #5077, opened 2020-11-06
Fixing BulkAll indefinitely hanging when failing

rmadisonhaynie:
Ensure BulkAll() is aborted properly if the whole bulk request keeps returning a bad statuscode and ContinueOnDroppedDocuments is true. This was introduced in #4014 when trying to address a similar issue with backOffTries being ignored.


stevejgordon (approved):
Thanks @rmadisonhaynie.

PR #5078, opened 2020-11-09
Make API generation non interactive

Mpdreamz:
Our `ApiGenerator` could only be run interactively. 

This PR adds proper flags and help and makes the tool non-interactive by default. 

In addition to this we switched to async everywhere allowing the tool to be interopted using e.g `CTRL+C`.


Mpdreamz:
Manually backported to `7.10` and `7.x`. 

`6.x` has drifted too far to make backporting this worthwhile.

PR #5082, opened 2020-11-09
Bump .NET sdk to 3.1.202

Mpdreamz:


azure-pipelines.yml
```diff
@@ -18,7 +18,7 @@ jobs:
   steps:
   - task: UseDotNet@2
     inputs:
-       version: '3.0.100'
+       version: '3.1.100'
```
stevejgordon, line 21: @Mpdreamz Any reason docs can be set to 3.1.202 as well here?

Mpdreamz: Updated thanks for catching that :)

stevejgordon (approved):
LGTM

PR #5085, opened 2020-11-10
Add support for the new `data_*` roles.

Mpdreamz:
Our integration tests were failing calling the `nodes.info` API because
it could not map `data_cold` to this enum.


stevejgordon (approved):
LGTM :shipit:

PR #5088, opened 2020-11-12
Remove performance counters from RecyclableMemoryStream

Mpdreamz:
We received a report that the finalizer of `EventPipeEventProvider` for a customer was failing on disposing the `Elasticsearch-Net-RecyclableMemoryStream`. 

By power of elimination we localized it to our counters. Initially we suspected the `instance` reference to be the problem but only removing those counters did not fix the problem.


Mpdreamz:
`backport to master` will likely fail because this moved to `Elastic.Transport` project there.

stevejgordon (approved):
LGTM!

PR #5089, opened 2020-11-12
Mapping time boost was removed in 7.10

Mpdreamz:
This was always a NOOP on the server and was removed from the server.

Here we obsolete them with a compiler warning


stevejgordon (approved):
LGTM!

Mpdreamz:
Opened https://github.com/elastic/elasticsearch/issues/64982 to log the breaking change on the server.

Mpdreamz:
`Boost` already removed on `master`

PR #5090, opened 2020-11-12
Reintroduce set for JsonParsingException.ActualChar

stevejgordon:
Removes a potentially breaking change for7.10 release.


Mpdreamz (approved):
LGTM 👍

PR #5097, opened 2020-11-16
Introduce assemble of SNAPSHOT builds

Mpdreamz:
Without affecting the version number stored in global.json

`.ci/make.sh` is the common entry for all clients builds and one of its
jobs is to build SNAPSHOT builds.

Long term we want to move to minver which means we no longer need to
store the version in global.json to begin with.


stevejgordon (approved):
My F# is pretty much non-existent but I think this all looks good and makes sense.

PR #5098, opened 2020-11-17
Add hard_bounds for histograms

stevejgordon:
Please consider this a WIP for review to check I'm on the best track.

Adds hard_bounds parameter to histogram aggregations per https://github.com/elastic/elasticsearch/pull/59175.

Questions:

For now, I've made `HardBounds` a copy of `ExtendedBounds`. I suspect we can't safely rename this public type to something more general. An option would be to define the properties in a base class from which these two could derive. Perhaps overkill here though.

I've updated two usage tests to confirm that the final JSON is built as expected. Are there any further tests we should be considering on additional such as this one?

### DIFF
--- a/src/Nest/Aggregations/Bucket/DateHistogram/DateHistogramAggregation.cs
+++ b/src/Nest/Aggregations/Bucket/DateHistogram/DateHistogramAggregation.cs
@@ -24,6 +24,13 @@ namespace Nest
 		ExtendedBounds<DateMath> ExtendedBounds { get; set; }
 
 		/// <summary>
+		/// The hard_bounds is a counterpart of extended_bounds and can limit the range of buckets in the histogram.
+		/// It is particularly useful in the case of open data ranges that can result in a very large number of buckets.
+		/// </summary>
+		[DataMember(Name = "hard_bounds")]
+		HardBounds<DateMath> HardBounds { get; set; }
+
+		/// <summary>
 		/// The field to target
 		/// </summary>
 		[DataMember(Name ="field")]
@@ -103,6 +110,8 @@ namespace Nest
 		/// <inheritdoc />
 		public ExtendedBounds<DateMath> ExtendedBounds { get; set; }
 		/// <inheritdoc />
+		public HardBounds<DateMath> HardBounds { get; set; }
+		/// <inheritdoc />
 		public Field Field { get; set; }
 
 		/// <inheritdoc />
@@ -115,7 +124,6 @@ namespace Nest
 					: _format;
 			set => _format = value;
 		}
-
 
 		[Obsolete("Deprecated in version 7.2.0, use CalendarInterval or FixedInterval instead")]
 		public Union<DateInterval, Time> Interval { get; set; }
@@ -147,6 +155,7 @@ namespace Nest
 		private string _format;
 
 		ExtendedBounds<DateMath> IDateHistogramAggregation.ExtendedBounds { get; set; }
+		HardBounds<DateMath> IDateHistogramAggregation.HardBounds { get; set; }
 		Field IDateHistogramAggregation.Field { get; set; }
 
 		//see: https://github.com/elastic/elasticsearch/issues/9725
@@ -227,6 +236,10 @@ namespace Nest
 		/// <inheritdoc cref="IDateHistogramAggregation.ExtendedBounds" />
 		public DateHistogramAggregationDescriptor<T> ExtendedBounds(DateMath min, DateMath max) =>
 			Assign(new ExtendedBounds<DateMath> { Minimum = min, Maximum = max }, (a, v) => a.ExtendedBounds = v);
+
+		/// <inheritdoc cref="IDateHistogramAggregation.HardBounds" />
+		public DateHistogramAggregationDescriptor<T> HardBounds(DateMath min, DateMath max) =>
+			Assign(new HardBounds<DateMath> { Minimum = min, Maximum = max }, (a, v) => a.HardBounds = v);
 
 		/// <inheritdoc cref="IDateHistogramAggregation.Missing" />
 		public DateHistogramAggregationDescriptor<T> Missing(DateTime? missing) => Assign(missing, (a, v) => a.Missing = v);
--- a/src/Nest/Aggregations/Bucket/DateHistogram/DateHistogramAggregation.cs
+++ b/src/Nest/Aggregations/Bucket/DateHistogram/DateHistogramAggregation.cs
@@ -119,7 +119,7 @@ namespace Nest
 		{
 			get => !string.IsNullOrEmpty(_format) &&
 				!_format.Contains("date_optional_time") &&
-				(ExtendedBounds != null || Missing.HasValue)
+				(ExtendedBounds != null || HardBounds != null || Missing.HasValue)
 					? _format + "||date_optional_time"
 					: _format;
 			set => _format = value;
@@ -163,7 +163,7 @@ namespace Nest
 		{
 			get => !string.IsNullOrEmpty(_format) &&
 				!_format.Contains("date_optional_time") &&
-				(Self.ExtendedBounds != null || Self.Missing.HasValue)
+				(Self.ExtendedBounds != null || Self.HardBounds != null || Self.Missing.HasValue)
 					? _format + "||date_optional_time"
 					: _format;
 			set => _format = value;
--- a/tests/Tests/Aggregations/Bucket/DateHistogram/DateHistogramAggregationUsageTests.cs
+++ b/tests/Tests/Aggregations/Bucket/DateHistogram/DateHistogramAggregationUsageTests.cs
@@ -46,6 +46,11 @@ namespace Tests.Aggregations.Bucket.DateHistogram
 						min = FixedDate.AddYears(-1),
 						max = FixedDate.AddYears(1)
 					},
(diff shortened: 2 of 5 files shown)


Mpdreamz:
Those tests are the correct place to add them but they are breaking the integration assertions:

https://github.com/elastic/elasticsearch-net/pull/5098/files#diff-abdcda181162c237e14568b13e1eda0e3aae3cc12efcc1f9b0c33e86bf203ccfR127

```
[xUnit.net 00:01:51.9685150]     Tests.Aggregations.Bucket.DateHistogram.DateHistogramAggregationUsageTests.ReturnsExpectedResponse [FAIL]
  X Tests.Aggregations.Bucket.DateHistogram.DateHistogramAggregationUsageTests.ReturnsExpectedResponse [939ms]
  Error Message:
   Tests.Framework.EndpointTests.ResponseAssertionException : Expected value to be greater than 10, but found 0.
Response Under Test:
Valid NEST response built from a successful (200) low level call on POST: /projects-only/_search?pretty=true&error_trace=true&typed_keys=false
 - [1] HealthyResponse: Node: http://localhost:9200/ Took: 00:00:00.0964947
# Request:
{"aggs":{"projects_started_per_month":{"aggs":{"project_tags":{"aggs":{"tags":{"terms":{"field":"tags.name"}}},"nested":{"path":"tags"}}},"date_histogram":{"extended_bounds":{"max":"2016-06-06T12:01:02.123","min":"2014-06-06T12:01:02.123"},"hard_bounds":{"max":"2016-06-06T12:01:02.123","min":"2014-06-06T12:01:02.123"},"field":"startedOn","format":"yyyy-MM-dd'T'HH:mm:ss||date_optional_time","calendar_interval":"month","min_doc_count":2,"missing":"2015-06-06T12:01:02.1230000","order":{"_count":"asc"}}}},"query":{"term":{"type":{"value":"project"}}},"size":0}
# Response:
{
  "took" : 7,
  "timed_out" : false,
  "_shards" : {
    "total" : 2,
    "successful" : 2,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 100,
      "relation" : "eq"
    },
    "max_score" : null,
    "hits" : [ ]
  },
  "aggregations" : {
    "projects_started_per_month" : {
      "buckets" : [ ]
    }
  }
}
```

This might warrant their own `*UsageTests`. 

You can run just a particular test with:

`.\built.bat integrate latest readonly DateHistogramAggregation`

stevejgordon:
@Mpdreamz I've reworked the tests, adding specific ones as you suggested. These seem to pass locally now too using `.\build.bat integrate 7.10.0 readonly` and `.\build.bat test`.

tests/Tests/Aggregations/Bucket/DateHistogram/DateHistogramAggregationUsageTests.cs
```diff
@@ -209,4 +210,82 @@ protected override void ExpectResponse(ISearchResponse<Project> response)
...
+		
+		public DateHistogramAggregationWithHardBoundsUsageTests(ReadOnlyCluster i, EndpointUsage usage) : base(i, usage)
+		{
+			// Note: If these tests are run against an existing node, and seeding is not forced, it's possible the
+			// dates used will not appear in the index and result in no buckets being returned. The test will still
+			// pass if this is the case. For best results locally, force a reseed. This is not an issue in CI.
+
+			var projects = Project.Projects.OrderBy(p => p.StartedOn).Skip(2).Take(5).ToArray();
+
+			_hardBoundsMinimum = DateTime.SpecifyKind(projects.First().StartedOn.Date, DateTimeKind.Unspecified);
```
Mpdreamz, line 229:
```suggestion
			_hardBoundsMinimum = DateTime.SpecifyKind(projects.Min(p=>p.StartedOn.Date), DateTimeKind.Unspecified);
```

Why is `DateTimeKind.Unspecified` needed?

tests/Tests/Aggregations/Bucket/DateHistogram/DateHistogramAggregationUsageTests.cs
```diff
@@ -209,4 +210,82 @@ protected override void ExpectResponse(ISearchResponse<Project> response)
...
+			_hardBoundsMaximum = DateTime.SpecifyKind(projects.Last().StartedOn.Date, DateTimeKind.Unspecified);
```
Mpdreamz, line 230:
```suggestion
			_hardBoundsMaximum = DateTime.SpecifyKind(projects.Max(p=>p.StartedOn.Date), DateTimeKind.Unspecified);
```

Mpdreamz (changes requested):

tests/Tests/Aggregations/Bucket/DateHistogram/DateHistogramAggregationUsageTests.cs
stevejgordon: Yep, looks nicer. Done!

tests/Tests/Aggregations/Bucket/DateHistogram/DateHistogramAggregationUsageTests.cs
stevejgordon: Done!

`DateTimeKind.Unspecified` is an interesting one. Without that we get back a server error:

```
"type" : "parse_exception",
        "reason" : "failed to parse date field [2019-11-27T00:00:00+00:00] with format [yyyy-MM-dd'T'HH:mm:ss]: [failed to parse date field [2019-11-27T00:00:00+00:00] with format [yyyy-MM-dd'T'HH:mm:ss]]",
```

Dates from `Project.Projects` are created a local kind. Later,`DateMath.ToString` handling of date formatting adds a suffix which ES is not happy about - https://github.com/elastic/elasticsearch-net/blob/master/src/Nest/CommonOptions/DateMath/DateMath.cs#L132

Would love to avoid this if there's a nicer way! 😄

russcam: I _think_ Elasticsearch might be using the format string passed in the test in the `Format` property, which doesn't include an offset, to parse the date into a long bounds:

https://github.com/elastic/elasticsearch/blob/0af410ad0b686af21f9a5cff25ceff2b56bf4863/server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/DateHistogramAggregationBuilder.java#L442-L446

and

https://github.com/elastic/elasticsearch/blob/4aaffc6a3deef28e8dedd3d9211951e3db0e2d8e/server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/LongBounds.java#L155-L171

I think that format is also used to format dates for the bucket key strings, in `ExpectResponse()`

Mpdreamz: ++ for this test it might sufficient to remove `format`, we might want to either add some `xmldocs` on `Format()` that its an expert option and might break the parsing. 

Sadly there is no way for us to carry over the chosen formatting in the response parsing.

stevejgordon: I think I ran into other issues with no formatting set at all. Let me double check that.

stevejgordon: Seems I can't replicate an issue right now. Running tests locally and if that all looks good I'll push a change up.

stevejgordon: Ok! Hopefully the latest change resolves this and works in CI. Everything ran correctly in local dev.

stevejgordon:
@Mpdreamz / @russcam - On further review I noticed that this was already handled for extended_bounds by adding date_optional_time to the format string. I've updated the code so we apply the same here and now formatting works as expected.

One thing I did notice is that the callout from https://github.com/elastic/elasticsearch-net/blob/master/tests/Tests/Aggregations/Bucket/DateHistogram/DateHistogramAggregationUsageTests.cs#L42 is not rendered into the final documentation - https://github.com/elastic/elasticsearch-net/blob/master/docs/aggregations/bucket/date-histogram/date-histogram-aggregation-usage.asciidoc

Is there something extra needed to make that work? It's affecting the current 7.x documentation. https://www.elastic.co/guide/en/elasticsearch/client/net-api/current/date-histogram-aggregation-usage.html

Mpdreamz (approved):

src/Nest/Aggregations/Bucket/DateHistogram/DateHistogramAggregation.cs
```diff
@@ -103,20 +110,21 @@ public class DateHistogramAggregation : BucketAggregationBase, IDateHistogramAgg
...
+		/// <inheritdoc />
 		public Field Field { get; set; }
 
 		/// <inheritdoc />
 		public string Format
 		{
 			get => !string.IsNullOrEmpty(_format) &&
 				!_format.Contains("date_optional_time") &&
-				(ExtendedBounds != null || Missing.HasValue)
+				(ExtendedBounds != null || HardBounds != null || Missing.HasValue)
```
Mpdreamz, line 122: ++ this hack comes back to me now 😄

PR #5099, opened 2020-11-18
Update to .NET 5 SDK

stevejgordon:
One interesting side-effect is a change to `System.Text.Json` which meant that `ErrorWithNullRootCausesTests` in the Tests project broke. Deserialisation now attempts to use ctors with parameters. The parameterless ctor is internal, and despite being reachable `System.Text.Json` is still preferring the ctor with params. However, it expects these to match the property names in the JSON so throws an exception.

A downside of this is we open up the parameterless constructor. Open to other suggestions here. A wider concern may be how this deserialisation behaves today for anyone using the library in a 5.0 runtime.


Mpdreamz (approved):
LGTM, getting there! 👍

Issue #5096, opened 2020-11-16
7.10.1 Meta Ticket


stevejgordon:
### **Breaking Java changes**

**Machine Learning**

- [ ] Deprecate allow_no_jobs and allow_no_datafeeds in favor of allow_no_match [#60601](https://github.com/elastic/elasticsearch/pull/60601) (issue: [#60642](https://github.com/elastic/elasticsearch/issues/60642))

### **Deprecations**

*Cluster Coordination**

~~- [ ] Deprecate and ignore join timeout [#60872](https://github.com/elastic/elasticsearch/pull/60872) (issue: [#60873](https://github.com/elastic/elasticsearch/issues/60873))~~

### **New features**

**Aggregations**

- [ ] Add rate aggregation [#61369](https://github.com/elastic/elasticsearch/pull/61369) (issue: [#60674](https://github.com/elastic/elasticsearch/issues/60674))

**Features/Features**

- [ ] Add data tiers (hot, warm, cold, frozen) as custom node roles [#60994](https://github.com/elastic/elasticsearch/pull/60994) (issue: [#60848](https://github.com/elastic/elasticsearch/issues/60848))

**Features/ILM+SLM**

- [ ] ILM migrate data between tiers [#61377](https://github.com/elastic/elasticsearch/pull/61377) (issue: [#60848](https://github.com/elastic/elasticsearch/issues/60848))
- [ ] ILM: add force-merge step to searchable snapshots action [#60819](https://github.com/elastic/elasticsearch/pull/60819) (issues: [#53488](https://github.com/elastic/elasticsearch/issues/53488), [#56215](https://github.com/elastic/elasticsearch/issues/56215))

**Machine Learning**

- [ ] Implement AucRoc metric for classification [#60502](https://github.com/elastic/elasticsearch/pull/60502) (issue: [#62160](https://github.com/elastic/elasticsearch/issues/62160))

**Mapping**

- [ ] Introduce 64-bit unsigned long field type [#60050](https://github.com/elastic/elasticsearch/pull/60050) (issue: [#32434](https://github.com/elastic/elasticsearch/issues/32434))

**Search**

- [ ] Add search _fields_ option to support high-level field retrieval [#60100](https://github.com/elastic/elasticsearch/pull/60100) (issues: [#49028](https://github.com/elastic/elasticsearch/issues/49028), [#55363](https://github.com/elastic/elasticsearch/issues/55363))

### **Enhancements**

**Aggregations**

[...]

PR #5104, opened 2020-11-19
Add QueryCache columns to CatNodesRecord

stevejgordon:
Adds missing QueryCache properties to CatNodesRecord to align with the possible columns which will be returned by _cat/nodes.

There are potentially other missing columns which we'll tackle in a future issue.

For now this at least contributes to #5096


Mpdreamz (approved):
LGTM

PR #5108, opened 2020-11-20
[DOCS] Update date histogram doc

stevejgordon:
Update date histogram documentation to include `hard_bounds` in the formatting note.

@Mpdreamz - Can we get away with backporting this to 7.10?


Mpdreamz (approved):
I think we can rely on backport for this one :)

PR #5109, opened 2020-11-20
[DOCS] Update date histogram doc

stevejgordon:
Update date histogram documentation to include hard_bounds in the formatting note.

@Mpdreamz - Generated this manually on master since the ref header is expected to be different. Is this the correct approach?


Mpdreamz (approved):

Mpdreamz:
> @Mpdreamz - Generated this manually on master since the ref header is expected to be different. Is this the correct approach?

Yes, for now it is. Looking to put the low level code, docs and examples generation behind github actions soon though.

PR #5120, opened 2020-11-25
Fix missing deprecated attributes on low-level

stevejgordon:
Identified a bug after creating a patch and regenerating code where the low level generate code was not rendering deprecated attributes for some methods.

The deduping logic handles aliases but does not handle non-aliased paths which is corrected by doing a final check to prefer the deprecated versions.

As far as I can tell by running the generator locally, this change doesn't seem to negatively affect any other parts of the code generation.

Additional:
- Updated the version of Spectre.Console
- Fixed breaking change from Spectre.Console
- Fixed typo

### DIFF
--- a/src/ApiGenerator/Domain/Specification/UrlInformation.cs
+++ b/src/ApiGenerator/Domain/Specification/UrlInformation.cs
@@ -2,6 +2,7 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Newtonsoft.Json;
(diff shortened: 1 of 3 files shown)


src/ApiGenerator/Domain/Specification/UrlInformation.cs
```diff
@@ -64,10 +65,23 @@ public IReadOnlyCollection<UrlPath> PathsWithDeprecations
...
 					.Select(grouped => grouped.First().deprecatedPath);
 
-
 				_pathsWithDeprecation = paths
 					.Concat(withoutDeprecatedAliases.Select(p => new UrlPath(p, OriginalParts, Paths)))
 					.ToList();
+
+				// now, check for and prefer deprecated URLs
+
+				foreach (var path in _pathsWithDeprecation.Where(p => p.Deprecation is null).ToArray())
```
Mpdreamz, line 74: LGTM but it would be better if we can do these mutations in a temporary variable and assign to `_pathsWithDeprecation` after the `foreach`

Mpdreamz (changes requested):

src/ApiGenerator/Domain/Specification/UrlInformation.cs
stevejgordon: Yep, good shout, updated.

Mpdreamz (approved):

PR #5121, opened 2020-11-25
Add ignore_keywords flag to word delimiter graph

stevejgordon:
This introduces a new property for the word delimiter graph token filter to configure ignoring of keywords.

It relates to this change https://github.com/elastic/elasticsearch/pull/59563

Contributes to #5096 

@Mpdreamz - I included more verbose XML comments for this to make usage via the Fluent API easier. I noticed that currently all descriptor methods use `/// <inheritdoc />` which seems to have no effect. Is there a reason for the existing comments? If this format is reasonable, I'd propose to spend some time updating other comments for the `WordDelimiterGraphTokenFilterDescriptor` and as I come to them, other descriptors.

### DIFF
--- a/src/Nest/Analysis/TokenFilters/WordDelimiterGraph/WordDelimiterGraphTokenFilter.cs
+++ b/src/Nest/Analysis/TokenFilters/WordDelimiterGraph/WordDelimiterGraphTokenFilter.cs
@@ -57,6 +57,13 @@ namespace Nest
 		bool? GenerateWordParts { get; set; }
 
 		/// <summary>
+		/// If true, the filter skips tokens with a keyword attribute of true. Defaults to false.
+		/// </summary>
+		[DataMember(Name = "ignore_keywords")]
+		[JsonFormatter(typeof(NullableStringBooleanFormatter))]
+		bool? IgnoreKeywords { get; set; }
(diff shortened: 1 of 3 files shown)


Mpdreamz:
We used to have a custom F# build task that patched our xml documentation files so that `<inheritdoc/>` would be filled in based on the xml docs on the interface. 


When pushing the transport bits out to https://github.com/elastic/elastic-transport-net, I re-enabled compilation failing on missing xmldocs. 

It also relies on `<inheritdoc>` but this time through 


https://github.com/saucecontrol/InheritDoc


ALthough more explicitly though `<inheritdoc cref="IInterface.Property" />". 

(small lie in that `saucecontrol/InheritDoc` is not hooked up yet, but I intend too! :))


More xmldocs is always good! 

Long term property descriptions will be picked up from the specification:

https://github.com/elastic/elastic-client-generator/blob/master/specification/specs/analysis/token_filters/word_delimiter_graph/WordDelimiterGraphTokenFilter.ts#L13

In which case I think generating them once on the interface and using `<inheritdoc>` for implementations is still a great way to keep to keep the generated LOC down.

Mpdreamz (approved):

PR #5123, opened 2020-11-26
Generate initial NEST code for 7.10.1

stevejgordon:
Adds patch for flush_synced specification.

Includes update to get-role after change to API spec identifying this is actually a comma separated list.


Mpdreamz (approved):
LGTM 💯

PR #5133, opened 2020-11-26
Update BoxplotAggregate response

stevejgordon:
In 7.11 the boxplot aggregation response includes two additional properties, lower and upper. I've added these to our type and updated the deserialization to handle these when present.


Mpdreamz (approved):
LGTM, 

in 7.10 we need a different fix that makes sure the formatter reads over these without assigning to new properties.

stevejgordon:
@Mpdreamz Sounds good. I'll drop the backport to 7.10 and put in a fresh PR to adjust accordingly. Would such changes to avoid the exception be needed in patches of 7.9 and earlier? Should 7.9.x client work with a 7.11.0 cluster?

PR #5139, opened 2020-11-30
Handle additional JSON properties

stevejgordon:
The current deserialization logic does not handle additional JSON properties, being added for Elasticsearch 7.11.

This PR reads over them, if they are present to avoid an exception. Since these are not available in 7.10, we do not use the values.


Mpdreamz (approved):
LGTM, maybe add a comment why this is needed?

stevejgordon:
Thanks @Mpdreamz. Good shout on the comment. Added!

PR #5140, opened 2020-11-30
Update System.Diagnostics.DiagnosticSource

stevejgordon:
Update to use the latest 5.0.0 version which supports all targets.

### DIFF
--- a/src/Elasticsearch.Net/Diagnostics/Diagnostic.cs
+++ b/src/Elasticsearch.Net/Diagnostics/Diagnostic.cs
@@ -45,8 +45,7 @@ namespace Elasticsearch.Net.Diagnostics
 			}
 		}
 
-
 		//_source can be null if Default instance
-		public void Dispose() => _source?.StopActivity(SetEndTime(DateTime.UtcNow), EndState);
+		protected override void Dispose(bool disposing) => _source?.StopActivity(SetEndTime(DateTime.UtcNow), EndState);
 	}
 }
--- a/src/Elasticsearch.Net/Diagnostics/Diagnostic.cs
+++ b/src/Elasticsearch.Net/Diagnostics/Diagnostic.cs
@@ -18,13 +18,14 @@ namespace Elasticsearch.Net.Diagnostics
 			EndState = state;
 	}
 	
-	internal class Diagnostic<TState, TStateEnd> : Activity, IDisposable
+	internal class Diagnostic<TState, TStateEnd> : Activity
 	{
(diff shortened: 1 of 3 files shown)


src/Elasticsearch.Net/Diagnostics/Diagnostic.cs
```diff
@@ -45,8 +45,7 @@ internal set
 			}
 		}
 
-
 		//_source can be null if Default instance
-		public void Dispose() => _source?.StopActivity(SetEndTime(DateTime.UtcNow), EndState);
+		protected override void Dispose(bool disposing) => _source?.StopActivity(SetEndTime(DateTime.UtcNow), EndState);
```
Mpdreamz, line 49: We did this here as well: https://github.com/elastic/elastic-transport-net/blob/main/src/Elastic.Transport/Diagnostics/Diagnostic.cs#L50

Wonder if we need to guard against `disposing` actually being `true`? 

I remember reading the base implementation and seeing this would only be called once, if this is true maybe we should add a comment about that here (and in `elastic-transport-net` for the next time we reread this line :D)

Mpdreamz (approved):
LGTM, added a note on the `Dispose` on `Diagnostic`

src/Elasticsearch.Net/Diagnostics/Diagnostic.cs
stevejgordon: Yes I copied this from transport! 😄 

I'll take a look into the base and see what needs to happen here. Should probably following the IDisposable pattern properly by checking the `bool`!

stevejgordon: I've pushed an update to properly implement the `IDisposable` pattern for this derived type.

stevejgordon:
@Mpdreamz Requesting a re-review after updating to implement `IDisposable`.

Mpdreamz (approved):
LGTM! mind opening a ticket/pr on elastic/elastic-transport-net straight after merging?

ejsmith:
@stevejgordon any plans to support `ActivitySource` ?

stevejgordon:
@ejsmith It's something I've recently added to my list to investigate and review. We're certainly keen to leverage newer features where they provide value. Beyond the activities built into `HttpClient` etc. do you have specific types of activities which you'd find useful to track on top, within the client? Would love to drive the planning from developer focused requirements.

ejsmith:
@stevejgordon would just be interested in the same things that the `DiagnosticSource`s currently output. Just in the more modern and easier to subscribe to `ActivitySource` format.

PR #5142, opened 2020-12-01
Implement grant API key

stevejgordon:
Contributes to #5096

Perhaps requires a simple test derived from `ApiTestBase` to assert on the expected JSON, URL etc? These are pretty much exercised by the main coordinated test.


Mpdreamz (approved):
LGTM!

PR #5147, opened 2020-12-02
Support clear api key cache

stevejgordon:
I updated `UrlPart` to support the use of the `Ids` type which seemed appropriate.

One query for review... The [docs for this API](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/security-api-clear-api-key-cache.html) suggest that the ids in the path can be empty to clear cache for all keys. The generated code requires the presence of an ID part and I'm not sure if this is okay or how best to address it?

Contributes to #5096

### DIFF
--- a/src/ApiGenerator/RestSpecification/_Patches/security.clear_api_key_cache.patch.json
+++ b/src/ApiGenerator/RestSpecification/_Patches/security.clear_api_key_cache.patch.json
@@ -1,0 +1,7 @@
+{
+  "security.clear_api_key_cache": {
+    "url": {
+      "paths": ["/_security/api_key/{ids}/_clear_cache", "/_security/api_key/*/_clear_cache"]
+    }
+  }
+}
--- a/tests/Tests/XPack/Security/ApiKey/SecurityApiKeyTests.cs
+++ b/tests/Tests/XPack/Security/ApiKey/SecurityApiKeyTests.cs
@@ -214,7 +214,8 @@ namespace Tests.XPack.Security.ApiKey
 					)
 			},
 			{
-				ClearApiKeyCacheStep, u =>
+				// API introduced in 7.10.0
+				ClearApiKeyCacheStep, ">=7.10.0", u =>
 					u.Calls<ClearApiKeyCacheDescriptor, ClearApiKeyCacheRequest, IClearApiKeyCacheRequest, ClearApiKeyCacheResponse>(
 						v => new ClearApiKeyCacheRequest(u.Usage.CallUniqueValues.ExtendedValue<string>("apiKey") ?? string.Empty),
 						(v, d) => d,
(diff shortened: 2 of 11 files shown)


Mpdreamz (changes requested):
We need to send an PR to elastic/elasticsearch to update the source spec

https://github.com/elastic/elasticsearch/blob/7.x/x-pack/plugin/src/test/resources/rest-api-spec/api/security.clear_api_key_cache.json#L9

To include the path with no `ids`. 

We can use a patch file in our codebase to unblock us on this PR.

Mpdreamz:
> One query for review... The docs for this API suggest that the ids in the path can be empty to clear cache for all keys. The generated code requires the presence of an ID part and I'm not sure if this is okay or how best to address it?

Excellent catch @stevejgordon  🔍 👍

stevejgordon:
@Mpdreamz I think I've patched this correctly and now per the updated tests, we can support optional IDs.

Mpdreamz (approved):
LGTM, don't forget to send an upstream PR for this :)

src/ApiGenerator/RestSpecification/_Patches/security.clear_api_key_cache.patch.json
```diff
@@ -0,0 +1,7 @@
+{
+  "security.clear_api_key_cache": {
+    "url": {
+      "paths": ["/_security/api_key/{ids}/_clear_cache", "/_security/api_key/*/_clear_cache"]
```
Mpdreamz, line 4:
```suggestion
      "paths": ["/_security/api_key/{ids}/_clear_cache", "/_security/api_key/_clear_cache"]
```

Mpdreamz: If this route is not available, for now patching it with `*` is fine on our end. Upstream this should not be the long term fix.

tests/Tests/XPack/Security/ApiKey/SecurityApiKeyTests.cs
```diff
@@ -227,5 +277,21 @@ public SecurityApiKeyTests(XPackCluster cluster, EndpointUsage usage) : base(new
 			r.PreviouslyInvalidatedApiKeys.Should().BeEmpty();
 			r.InvalidatedApiKeys.Should().HaveCount(2);
 		});
+
+		[I] public async Task SecurityClearApiKeyCacheResponse() => await Assert<ClearApiKeyCacheResponse>(ClearApiKeyCacheStep, r =>
```
Mpdreamz, line 281: `SkipVersion` attribute?

Mpdreamz (changes requested):

src/ApiGenerator/RestSpecification/_Patches/security.clear_api_key_cache.patch.json
stevejgordon: Yeah, that route is not available (yet) so will leave `*` for the patch for the time being.

tests/Tests/XPack/Security/ApiKey/SecurityApiKeyTests.cs
stevejgordon: `Assert` skips these and returns, if the step has already been identified for skipping based on it's version range. As discussed offline, this is probably preferable.

```csharp
protected async Task Assert<TResponse>(string name, Action<TResponse> assert)
	where TResponse : class, IResponse
{
	if (_coordinatedUsage.Skips(name)) return;
	...
}
```

PR #5148, opened 2020-12-02
Support indexing pressure on NodeStats

stevejgordon:
Indexing pressure stats were added in 7.9.0
https://github.com/elastic/elasticsearch/pull/59467

They were updated to include memory limit stats in 7.10.0
https://github.com/elastic/elasticsearch/pull/60342

Contributes to #5096

### DIFF
--- a/src/Nest/Cluster/NodesStats/NodeStats.cs
+++ b/src/Nest/Cluster/NodesStats/NodeStats.cs
@@ -80,6 +80,9 @@ namespace Nest
 
 		[DataMember(Name = "transport_address")]
 		public string TransportAddress { get; internal set; }
+
+		[DataMember(Name = "indexing_pressure")]
+		public IndexingPressureStats IndexingPressure { get; internal set; }
 	}
 
 	[DataContract]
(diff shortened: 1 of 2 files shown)


src/Nest/Cluster/NodesStats/NodeStats.cs
```diff
@@ -571,4 +574,74 @@ public class HttpStats
...
 	}
+
+	[DataContract]
+	public class IndexingPressureStats
+	{
+		[DataMember(Name = "memory")]
+		public MemoryStats Memory { get; internal set; }
+
+		[DataContract]
+		public class MemoryStats
```
Mpdreamz, line 585: While I think inner classes are the best choice stylistically once we shift from hand rolling these to generating these it will be hard to know when inner classes are preferred. 

I hate saying it but [smurf types](https://blog.codinghorror.com/new-programming-jargon/) are better here

smurf typing

Mpdreamz (changes requested):

src/Nest/Cluster/NodesStats/NodeStats.cs
stevejgordon: That's a new term for me 😄

stevejgordon: I've moved out to non-nested classes @Mpdreamz. If this change is what you expected, hopefully we're good to merge.

src/Nest/Cluster/NodesStats/NodeStats.cs
```diff
@@ -571,4 +574,74 @@ public class HttpStats
...
+		public string Coordinating { get; internal set; }
+
+		[DataMember(Name = "primary_in_bytes")]
+		public long PrimaryInBytes { get; internal set; }
+
+		[DataMember(Name = "primary")]
+		public string Primary { get; internal set; }
+
+		[DataMember(Name = "replica_in_bytes")]
+		public long ReplicaInBytes { get; internal set; }
```
Mpdreamz, line 623: Does this always have a value? if not should be `long?`.

Mpdreamz (approved):
LGTM, I left one comment about value type properties might needing to be nullable depending on wheter they are always present or not.

src/Nest/Cluster/NodesStats/NodeStats.cs
stevejgordon: In my tests against a local 7.10 node, it always returns as 0 so I assume safe to make it non nullable.

stevejgordon: image

PR #5163, opened 2020-12-07
Update with additional sdk version properties

stevejgordon:
After adding rollForward and allowPreRelease to the global.json, this change ensures they remain present after building a release.


Mpdreamz (approved):
LGTM, good catch

PR #5180, opened 2020-12-16
Tweaks .ci/make.sh to clean up after itself

Mpdreamz:
- Tweaks to .ci/make.sh

Now also removes the specified output path, not just the hardcoded version.

Publishing nuget packages now no longer uses a special subfolder `_packages` to publish.
This to fall in line with how other clients  will use `.ci/make.sh`.

- remove references to _packages from other targets

### DIFF
--- a/.ci/make.sh
+++ b/.ci/make.sh
@@ -1,11 +1,9 @@
 #!/usr/bin/env bash
 # parameters are available to this script
 
-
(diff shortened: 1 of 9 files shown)


stevejgordon (approved):
LGTM

Mpdreamz:
Manually backported to `7.x` and `7.10`

PR #5187, opened 2020-12-17
Remove mention of type from Getting Started docs

russcam:
This commit removes the mention of document types from the Getting Started docs.


Mpdreamz (approved):
awesome, thanks @russcam 👍

PR #5190, opened 2020-12-17
Add metadata header on requests

stevejgordon:
Adds a `x-elastic-client-meta` header to HTTP requests sent by Elasticsearch.NET. This header contains information about the runtime environment that is meant to allow analysing usage by collecting this information on the receiving side of requests, like a proxy server in front of ES.

The information in this header has been specified by the clients team to be consistent across language clients. It contains the versions of the client, the language runtime (.NET in this case) and its HTTP library, and the use of the high level client. 

This header is sent by default and can be disabled by setting `DisableMetaHeader` on the `ConnectionSettings` passed to the client.

There is a little more work to do on this PR after an initial review of the approaches taken.

Remaining tasks:

- [x] Update configuration documentation
- [x] Add additional tests
- [x] Final QA against different runtimes
- [x] Review/update XML comments
- [ ] Manual porting to Elastic.Transport and master as appropriate


src/Elasticsearch.Net/Connection/HttpConnection.cs
```diff
@@ -248,6 +248,8 @@ protected virtual HttpMessageHandler CreateHttpClientHandler(RequestData request
 
 		protected virtual HttpRequestMessage CreateHttpRequestMessage(RequestData requestData)
 		{
+			requestData.HttpClientIdentifier = ConnectionInfo.UsingCurlHandler ? "c" : "s";
```
Mpdreamz, line 251: I think we should do this work when instantiating the connection configuration. And have the connection configuration expose two default headers (async, non async).

`HttpClientIdentifier` could become a property all `IConnection`'s have to expose.

src/Elasticsearch.Net/Connection/MetaHeaderProvider.cs
```diff
@@ -0,0 +1,100 @@
...
+				var sb = StringBuilderCache.Acquire(64);
+
+				sb.Append("es=").Append(clientVersion).Append(_separator);
+				sb.Append("a=").Append(requestData.IsAsync ? "1" : "0").Append(_separator);
+				sb.Append("net=").Append(runtimeVersion).Append(_separator);
+
+				if (!string.IsNullOrEmpty(requestData.HttpClientIdentifier))
+					sb.Append(requestData.HttpClientIdentifier).Append("=").Append(runtimeVersion).Append(_separator);
+
+				foreach (var requestMetaData in requestData.RequestMetaData ?? EmptyReadOnly<string, string>.Dictionary)
```
Mpdreamz, line 82: Move `EmptyReadOnly` to getter for `requestData.RequestMetaData` so it can never be null (when we enable nullability checks).

src/Elasticsearch.Net/Connection/RuntimeVersion.cs
```diff
@@ -0,0 +1,316 @@
...
+	/// </summary>
+	internal sealed class RuntimeVersion
+	{
+		internal static readonly RuntimeVersion Default = new RuntimeVersion(new Version(0, 0, 0), false);
+
+		private RuntimeVersion(Version version, bool isPrerelease)
+		{
+			var finalVersion = version;
+
+			if (version.Minor == -1 || version.Build == -1)
```
Mpdreamz, line 51: This should get flagged as a style violation in our .editorconfig. Be keen to find out why if it does not in VS. (e.g no {} on non complex statements.

Mpdreamz (changes requested):

src/Elasticsearch.Net/Connection/MetaHeaderProvider.cs
stevejgordon: Looks like this already happens when it's set from the ctor so I can drop the check here.

src/Elasticsearch.Net/Connection/RuntimeVersion.cs
stevejgordon: Will fix and investigate editor.config

stevejgordon:
@Mpdreamz I've refactored this a fair bit taking your suggestions onboard. I'm happier with this approach now. Would welcome a check to see if it matches what you were thinking?

Mpdreamz:
Aye much cleaner! 

I am still unsure if we need:

```
private readonly List<IHeaderProvider> _customerHeaderProviders = new List<IHeaderProvider> { new MetaHeaderProvider() };
```

And `IHeaderProvider` since the instance we plan to maintain for the foreseeable future is the one producing: `x-elastic-client-meta"`. 

For users wanting to inject we already have `.Headers` on `ConnectionSettings` and `RequestConfiguration`.

src/Elasticsearch.Net/Connection/MetaData/MetaDataHeaders.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+	internal sealed class MetaDataHeaders : IMetaDataHeaders
+	{
+		public MetaDataHeaders(ClientVersionInfo clientVersionInfo)
+		{
+			AsyncMetaDataHeader = new MetaDataHeader(clientVersionInfo, "es", true);
+			SyncMetaDataHeader = new MetaDataHeader(clientVersionInfo, "es", false);
+		}
+
+		public MetaDataHeader AsyncMetaDataHeader { get; private set; }
```
Mpdreamz, line 15:
```suggestion
		public MetaDataHeader AsyncMetaDataHeader { get; }
```

src/Elasticsearch.Net/Connection/MetaData/MetaDataHeaders.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+namespace Elasticsearch.Net
+{
+	internal sealed class MetaDataHeaders : IMetaDataHeaders
```
Mpdreamz, line 7: both `internal` and `sealed` are used very sparingly. Siding on the side of being public by default has some benefits.

Not sure we need `IMetaDataHeaders` if this class is public? It also allows us to unit  test this in isolation.

src/Elasticsearch.Net/Connection/MetaData/MetaDataHeaders.cs
Mpdreamz: Maybe a good first `record` ? 😄

Mpdreamz (changes requested):

src/Elasticsearch.Net/Connection/MetaData/MetaDataHeaders.cs
stevejgordon: Yeah it could be. I want to read up on using records for older targets since I believe we may need to include code for a couple of attributes which may be required for them to work after compilation.

src/Elasticsearch.Net/Connection/MetaData/MetaDataHeaders.cs
stevejgordon: In this case, I think I probably agree 😄. I tend to lean the other way, internal and sealed until they need to be since we can change them freely without breaking changes to consumers. But when I've done that in other libs, it can cause some headaches for testing. Here, I think loosing the interface is better.

stevejgordon: @Mpdreamz In the latest update to this PR I've been able to avoid exposing this on the configuration entirely. Therefore I've left this internal now since it's used only within the provider.

stevejgordon:
@Mpdreamz I think we're ready for a final review on this please. I've refactored and simplified this further on the basis of YAGNI. 

I also added `MetaHeaderHelperTests` which uses a custom InMemoryConnection to simulate responses necessary to fully exercise the observables. This is mainly to support verification of expected helper on the `RequestMetaHeader` but has the nice side effect of exercising the flow of requests in general.

Mpdreamz (approved):
LGTM

PR #5199, opened 2020-12-17
Update generated low-level client code for 7.11

stevejgordon:
- Excludes QueryWatches from API renaming
- Includes a patch to correct _cat/_tasks so that they are marked experimental


Mpdreamz (approved):
LGTM 👍 

We'll need to backport the `QueryWatches` exception manually and rerun the code gen on them too.

This codegen PR's don't backport automatically nicely sadly 😄

PR #5207, opened 2020-12-22
Generate initial NEST code for 7.11.0

stevejgordon:
Includes a fix to the razor file to use correct removal version of 8.0


stevejgordon:
Will re-raise this once #5199 is merged and I can base of those changes

Issue #5198, opened 2020-12-17
7.11.0 Meta Ticket


stevejgordon:
### Breaking changes

**Machine Learning**

- [x]   Add new flag exclude_generated that removes generated fields in GET config APIs [#63899](https://github.com/elastic/elasticsearch/pull/63899) (issue: [#63055](https://github.com/elastic/elasticsearch/issues/63055))

### Deprecations

**Features/CAT APIs**

- [x]   Deprecate the 'local' parameter of /_cat/indices [#62198](https://github.com/elastic/elasticsearch/pull/62198) (issue: [#60718](https://github.com/elastic/elasticsearch/issues/60718))
- [x]   Deprecate the 'local' parameter of /_cat/shards [#62197](https://github.com/elastic/elasticsearch/pull/62197) (issue: [#60718](https://github.com/elastic/elasticsearch/issues/60718))

**MULTIPLE AREA LABELS**

- [x]   Deprecates _upgrade API (#47678) [#50484](https://github.com/elastic/elasticsearch/pull/50484) (issue: [#47678](https://github.com/elastic/elasticsearch/issues/47678))

**Mapping**

- [ ] ~~Emit a warning when an unknown geo strategy is passed in [#63975](https://github.com/elastic/elasticsearch/pull/63975)~~

### New features

**Aggregations**

- [ ]   Support global ords in top_metrics [#64967](https://github.com/elastic/elasticsearch/pull/64967) (issue: [#64774](https://github.com/elastic/elasticsearch/issues/64774))

**Features/Ingest**

- [x] URI parts ingest processor [#65150](https://github.com/elastic/elasticsearch/pull/65150) (issue: [#57481](https://github.com/elastic/elasticsearch/issues/57481)) - **PR #5304**

**Geo**

- [x]   Add geo_line aggregation [#41612](https://github.com/elastic/elasticsearch/pull/41612) (issue: [#41649](https://github.com/elastic/elasticsearch/issues/41649))

**MULTIPLE AREA LABELS**

- [ ]   Add doc_count field mapper [#64503](https://github.com/elastic/elasticsearch/pull/64503) (issue: [#58339](https://github.com/elastic/elasticsearch/issues/58339))

**Search**

- [x]   Async search status API [#62947](https://github.com/elastic/elasticsearch/pull/62947) (issue: [#57537](https://github.com/elastic/elasticsearch/issues/57537)) - **PR #5239**

### Enhancements

**Aggregations**

[...]

PR #5208, opened 2020-12-22
Support allow_duplicates on append processor

stevejgordon:
Related to https://github.com/elastic/elasticsearch/pull/61916
Contributes to #5198

### DIFF
--- a/tests/Tests/Ingest/ProcessorAssertions.cs
+++ b/tests/Tests/Ingest/ProcessorAssertions.cs
@@ -9,6 +9,7 @@ using System.Reflection;
 using Elastic.Elasticsearch.Xunit.XunitPlumbing;
 using Nest;
 using Tests.Core.Client;
+using Tests.Core.Extensions;
 using Tests.Core.Xunit;
 using Tests.Domain;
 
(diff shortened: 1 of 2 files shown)


tests/Tests/Ingest/ProcessorAssertions.cs
```diff
@@ -59,19 +60,41 @@ public static IPromise<IList<IProcessor>> Fluent(ProcessorsDescriptor d)
...
+			public override IProcessor Initializer
 			{
-				Field = "state",
-				Value = new object[] { StateOfBeing.Stable, StateOfBeing.VeryActive }
-			};
+				get
+				{
+					var ap = new AppendProcessor { Field = "state", Value = new object[] { StateOfBeing.Stable, StateOfBeing.VeryActive } };
+
+					if (TestClient.Configuration.InRange(">=7.11.0"))
```
Mpdreamz, line 80: Better to add a `SkipVersion` on the `Append` class and include the additional property. 

This way on master we can clean up the `SkipVersion` easier when we go over all the unnecessary skips.

Mpdreamz (changes requested):
I think its okay to simplify the `Append` cases to only run on `7.11.0` or include a `AppendWithAllowDuplicates` additional test class. 

Easier to track then dynamically built json. 


Also failing on the following `CodeStandards` tests:

```
Tests.CodeStandards.Descriptors.DescriptorMethodsTakingSingleValueTypeShouldBeNullable [FAIL]
  Failed Tests.CodeStandards.Descriptors.DescriptorMethodsTakingSingleValueTypeShouldBeNullable [166 ms]
  Error Message:
   Expected breakingDescriptors to be empty, but found {"allowDuplicates on method AllowDuplicates of Nest.AppendProcessorDescriptor`1 is not nullable"}.
  Stack Trace:
```

Descriptors taking value types should be nullable so that they be effectively reset.

tests/Tests/Ingest/ProcessorAssertions.cs
stevejgordon: I've done as suggested in the comment and added a separate test which covers allow duplicates on 7.11+

stevejgordon:
@Mpdreamz Ready for a re-review please

Mpdreamz (approved):
LGTM !

PR #5209, opened 2020-12-23
Fix failing transform tests and support dates_as_epoch_millis

### LINKED ISSUE #5185: [TESTS] TransformApiTests throwing JsonParsingException
It looks like some of the TransformApiTests are failing on CI due to a JsonParsingException:

https://github.com/elastic/elasticsearch-net/pull/5177/checks?check_run_id=1561871433

- Tests.XPack.Transform.TransformApiTests.StopTransformResponse
- Tests.XPack.Transform.TransformApiWithSettingsTests.ReturnsExpectedResponse
- Tests.XPack.Transform.TransformApiTests.DeleteTransformResponse
- Tests.XPack.Transform.TransformApiWithSettingsTests.ReturnsExpectedStatusCode

```
Elastic.Transport.UnexpectedTransportException : expected:'Number Token', actual:'"2014-12-29T00:00:00.000Z"', at offset:79
---- Nest.Utf8Json.JsonParsingException : expected:'Number Token', actual:'"2014-12-29T00:00:00.000Z"', at offset:79`
```

stevejgordon:
Tests were failing due to a change in the transform API which now prefers ISO dates, rather than epoch milliseconds for dates. This resulted in a serialisation failure.

This commit addresses that by conditionally producing the requests based on the version of ES being tested, so that a property of the appropriate type is used for deserialisation of the preview.

This also adds support for the now dates_as_epoch_millis setting which allows the original epoch millisecond format to be returned for ES 7.11 requests, when the option is true.

Fixes #5185
Contributes to #5198

### DIFF
--- a/src/Nest/XPack/Transform/TransformSettings.cs
+++ b/src/Nest/XPack/Transform/TransformSettings.cs
@@ -21,6 +21,9 @@ namespace Nest
 
 		[DataMember(Name = "max_page_search_size")]
 		public int? MaxPageSearchSize { get; set; }
+
+		[DataMember(Name = "dates_as_epoch_millis")]
+		public bool? DatesAsEpochMilliseconds { get; set; }
 	}
 
 	/// <inheritdoc />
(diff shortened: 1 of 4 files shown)


Mpdreamz (approved):
LGTM 👍

PR #5212, opened 2020-12-28
[Backport master] Fix failing transform tests and support dates_as_epoch_millis

Mpdreamz:
Tests were failing due to a change in the transform API which now
prefers ISO dates, rather than epoch milliseconds for dates. This
resulted in a serialisation failure.

This commit addresses that by conditionally producing the requests
based on the version of ES being tested, so that a property of the
appropriate type is used for deserialisation of the preview.

This also adds support for the now dates_as_epoch_millis setting which
allows the original epoch millisecond format to be returned for ES 7.11
requests, when the option is true.

(cherry picked from commit 7994d6fdf7734a5cb7b774e80dfc463fefd2b9cb)
Backport 7994d6fdf7734a5cb7b774e80dfc463fefd2b9cb from #5209

### DIFF
--- a/src/Nest/XPack/Transform/TransformSettings.cs
+++ b/src/Nest/XPack/Transform/TransformSettings.cs
@@ -21,6 +21,9 @@ namespace Nest
 
 		[DataMember(Name = "max_page_search_size")]
 		public int? MaxPageSearchSize { get; set; }
+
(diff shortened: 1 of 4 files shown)


Mpdreamz:
known flakey tests in `(WindowsIntegrationTests latest)`

PR #5213, opened 2020-12-28
Fix broken yaml test bootstrap

Mpdreamz:
Causing all the xpack tests to fail.


Mpdreamz:
`clients-ci` is happy, therefore I am 😸 . cc @stevejgordon self reviewing this so we can start the new year 🟢 🤞

PR #5216, opened 2020-12-28
Skip 20_fvh.yml temporarily while we wait for the fix to land

Mpdreamz:
Already fixed upstream but needs to find its way to the docker images.

Relates: elastic/elasticsearch#66794


Mpdreamz:
`clients-ci` happy, I'm happy 😸 . @stevejgordon self reviewing this to get us focusing on other things then 🟢 builds after the holidays :)

PR #5228, opened 2021-01-04
Fix deserialisation of aggs in Translate response

### LINKED ISSUE #5201: NEST SQL Translate no aggregations returned
**NEST/Elasticsearch.Net version**: 7.01

**Elasticsearch version**:

SQL:
"select UDFvarchar1, count(1) from interactions group by UDFvarchar1 order by UDFvarchar1 "
Kibana works with this SQL and translate returns correctly

Calling Translate:
var q2 = new TranslateSqlRequest()
{
Query = "select UDFvarchar1, count(1) from interactions group by UDFvarchar1 order by UDFvarchar1 "

};
var sqlResponse2 = _client.Sql.Translate(q2);
var reqs = sqlResponse2.Result

Reqs.Aggregations is null

stevejgordon:
The Translate API in Elasticsearch returns aggregations using the key "aggregations" rather than "aggs". This results in aggregations not appearing in the `SearchRequest` instance.

This fix adds a field so that aggregations deserialise correctly using UTF8Json in the `DefaultHighLevelSerializer`.

I'll open an issue in Elasticsearch to review whether in 8.0, it should prefer "aggs" by default.

Fixes #5201

### DIFF
--- a/src/Nest/Aggregations/AggregationContainer.cs
+++ b/src/Nest/Aggregations/AggregationContainer.cs
@@ -294,7 +294,15 @@ namespace Nest
 	{
 		public IAdjacencyMatrixAggregation AdjacencyMatrix { get; set; }
 
-		public AggregationDictionary Aggregations { get; set; }
+		// This is currently used to support deserializing the response from SQL Translate,
+		// which forms a response which uses "aggregations", rather than "aggs". Longer term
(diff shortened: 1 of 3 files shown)


Mpdreamz (approved):
LGTM :+1:

PR #5230, opened 2021-01-05
Support mapping runtime fields

stevejgordon:
This PR adds support to map runtime fields on CREATE index and PUT mapping requests. Support for searches using runtime fields will follow in another PR.

@Mpdreamz I made RuntimeFieldsDescriptor non-generic as it doesn't need anything from TDocument. This seems to work fine but I noticed that some other descriptors are generic without accessing `T` so I wanted to check if this is problematic here?

Contributes to #5198


Mpdreamz (approved):
LGTM :+1:

Mpdreamz:
> This seems to work fine but I noticed that some other descriptors are generic without accessing T so I wanted to check if this is problematic here?

If they do the might use it for inference of indices or types, or they are simply typed wrong!

PR #5239, opened 2021-01-06
Implement AsyncSearch Status API

stevejgordon:
Main change to existing is removing `Response` from `AsyncSearchResponseBase` and making `AsyncSearchResponseBase` non-generic as a result. This allows the base to be used for the status response which differs by not include the response itself.


src/Nest/XPack/AsyncSearch/AsyncSearchResponseBase.cs
```diff
@@ -1,11 +1,15 @@
...
+// See the LICENSE file in the project root for more information
+
 using System;
 using System.Runtime.Serialization;
 using Elasticsearch.Net.Utf8Json;
 
 namespace Nest
 {
 	[InterfaceDataContract]
-	public interface IAsyncSearchResponse<TDocument> : IResponse where TDocument : class
```
Mpdreamz, line 8: This is a breaking change we can't do in `7.11` since this was already released `7.7`

stevejgordon:
Revised the approach to avoid the breaking change.

Mpdreamz (approved):
LGTM :+1:

PR #5253, opened 2021-01-07
Code gen for NEST to support backporting APIs

stevejgordon:
Mostly seems to include format changes which is at least consistent with the other branches.


Mpdreamz (approved):
:+1: LGTM

stevejgordon:
Rebased onto master after Transport PR merged. Waiting for CI as a double check before merging.

PR #5259, opened 2021-01-13
Implement runtime fields on searches

stevejgordon:


Mpdreamz (approved):
LGTM! This is a great new feature to have exposed :)

PR #5275, opened 2021-01-15
Implement unsigned long numeric field

stevejgordon:
Add support for mapping an unsigned_long field. There is an issue currently with any mapping where the `null_value` needs to be provided. This is because we store that value in a `double?` property and currently, this causes a mapper_parser_exception. This has been raised for the ES team in https://github.com/elastic/elasticsearch/issues/67565

Contributes #5198


Mpdreamz (approved):
LGTM :+1:

PR #5286, opened 2021-01-29
Add geo-line aggregration

stevejgordon:
Implements geo line aggregation and required JSON formatters.

Since the geo-line aggregation tests required using XPack cluster, I have also adjusted `AggregationUsageTestBase` to make it generic so that the Cluster can be passed in as needed.


Mpdreamz (approved):
LGTM!

PR #5289, opened 2021-01-30
[Backport master] Add geo-line aggregration

stevejgordon:
Backport 3312afeb28bb9ea5d68646274b01dbd0498b57cf from #5286

@Mpdreamz This re-enables some previously skipped tests (on master). I've run integration locally and all seems fine. Happy for these to be enabled as per the 7.x branch?


Mpdreamz (approved):
LGTM :+1:


> This re-enables some previously skipped tests (on master). I've run integration locally and all seems fine. Happy for these to be enabled as per the 7.x branch?

100%, The presence of `SkipVersion` on `master` should probably be a compiler warning even.

stevejgordon:
@Mpdreamz I may have been over keen. Some of these do fail against latest. Just scrolled back from my local commands I'd ran latest-7 so will fix that up and push a new commit. I'll probably skip for now and add an issue to track what appears to be aggregation naming changes.

PR #5294, opened 2021-02-01
Fix size overflow in StringToDoubleConverter

stevejgordon:
Original code resized the static if too small but doesn't use the resized array.

Thanks to @benaadams for reporting this bug.

### DIFF
--- a/src/Elasticsearch.Net/Utf8Json/Internal/DoubleConversion/StringToDoubleConverter.cs
+++ b/src/Elasticsearch.Net/Utf8Json/Internal/DoubleConversion/StringToDoubleConverter.cs
@@ -540,30 +540,24 @@ namespace Elasticsearch.Net.Utf8Json.Internal.DoubleConversion
 
             buffer[buffer_pos] = (byte)'\0';
 
-            double? converted;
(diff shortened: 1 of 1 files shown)


src/Elasticsearch.Net/Utf8Json/Internal/DoubleConversion/StringToDoubleConverter.cs
```diff
@@ -540,30 +540,24 @@ private static bool AdvanceToNonspace(ref Iterator current, Iterator end)
...
 
             if (converted == null)
             {
                 // read-again
                 processed_characters_count = (current - input);
 
-                var fallbackbuffer = GetFallbackBuffer();
-                BinaryUtil.EnsureCapacity(ref _fallbackBuffer, 0, processed_characters_count);
+                var fallbackBuffer = GetFallbackBuffer();
+                BinaryUtil.EnsureCapacity(ref fallbackBuffer, 0, processed_characters_count);
```
Mpdreamz, line 553: :mag_right:  good catch

Mpdreamz (approved):
LGTM!

PR #5302, opened 2021-02-03
[Backport 6.x] Implement meta header for client requests

stevejgordon:
Backport fec16efac5b30c1bb11b866d71e40b40450cf2f0 from #5190

@Mpdreamz This backports the work in 7.x for the meta header to 6.x. Some changes had to be made to get it to compile and function. A few types needed to be made public rather than internal as it seems NEST cannot use Elasticsearch.NET internals in 6.x. Integration tests work as expected against latest-6 and all unit tests are passing. I observed the header on the integration test requests.


Mpdreamz (approved):
Awesome! LGTM :+1:

stevejgordon:
I've pushed one change to an unrelated test which is failing on 6.0.1 and needs to be skipped since it uses a parameter not available in that version. Once CI re-runs I'll merge!

PR #5303, opened 2021-02-04
Adds back additional information when running our tests

Mpdreamz:
Dumps all the configuration bits at the beginning of the run including
test mode and random seed information.

At the end of the run this now again produces an overview of how long
testing each cluster took.

And produce a reproduce command line that you can use to only rerun
failed collections/tests using the exact same configuration.

### DIFF
--- a/tests/Tests.Core/Client/Settings/TestConnectionSettings.cs
+++ b/tests/Tests.Core/Client/Settings/TestConnectionSettings.cs
@@ -55,6 +55,9 @@ namespace Tests.Core.Client.Settings
 			.ConnectionLimit(ConnectionLimitDefault)
 			.OnRequestCompleted(r =>
 			{
+				//r.HttpMethod;
+
+
 				if (!r.DeprecationWarnings.Any()) return;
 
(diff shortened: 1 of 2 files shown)


tests/Tests.Core/Client/Settings/TestConnectionSettings.cs
```diff
@@ -55,6 +55,9 @@ public class TestConnectionSettings : ConnectionSettings
 			.ConnectionLimit(ConnectionLimitDefault)
 			.OnRequestCompleted(r =>
 			{
+				//r.HttpMethod;
```
stevejgordon, line 58: In by accident?

stevejgordon (approved):
LGTM! I actually had this on my list for the week!

tests/Tests.Core/Client/Settings/TestConnectionSettings.cs
Mpdreamz: in by accident will remove thanks for catching this!

PR #5307, opened 2021-02-05
Update SDK to latest patch

stevejgordon:
May as well keep these updated since we updated the dockerfile for the YAML tests earlier this week.


Mpdreamz (approved):
Should we bump global.json as well? 

https://github.com/elastic/elasticsearch-net/blob/master/global.json#L3

stevejgordon:
@Mpdreamz  - I considered that, but I think the roll forward should get it? That way any 5.x version should work during dev.

Mpdreamz:
@stevejgordon that works for me too :)

PR #5313, opened 2021-02-10
Skip cat template YAML test

stevejgordon:

PR #5315, opened 2021-02-10
Fix parsing of some yaml tests on master

Mpdreamz:
- Skip cat template YAML test (#5313)
- No longer fail to parse when allowed_warnings is last in the do operation definition


stevejgordon (approved):
LGTM - Thanks

PR #5317, opened 2021-02-11
Support multiple expand_wildcards

stevejgordon:
Since 7.10.1 the watcher expand_wildcards property has accepted and returns an array for expanded_wildcards. This broke in our existing representation of IndicesOptions during deserialisation. This change is breaking but now supports either a single string value or an array in the JSON response.

I've added an additional test to run on 7.10.1 and later to PUT and GET a watch with multiple `expand_wildcards`.

I've updated existing tests to add the optional `allow_no_indices` property due to a regression in 7.10.1 and 7.10.2 which meant missing optional fields caused a validation error on the server. This ensures the tests can run against any 7.10.x version.


Mpdreamz (approved):
LGTM, thanks for catching this!

PR #5330, opened 2021-02-18
Bump testing abstraction and fix cluster run command

Mpdreamz:
- update to latest libs from elastic/elasticsearch-net-abstractions and remove nightlies from nuget.config
- Update cluster build command so you can run clusters

You should now be able to call e.g `./build.sh cluster readonly latest-7` in a terminal and use as the cluster to run your scoped integration tests against as your developing them.

Greatly reducing the time you have to wait for the cluster to start.


stevejgordon (approved):
LGTM!

PR #5378, opened 2021-03-01
Update get token response to include auth info

stevejgordon:
Updates the response as a result of this enhancement - https://github.com/elastic/elasticsearch/pull/62490

I've taken a similar approach as the server-side implementation by reusing the `AuthenticateResponse` type.

### DIFF
--- a/src/Nest/XPack/Security/User/GetUserAccessToken/GetUserAccessTokenResponse.cs
+++ b/src/Nest/XPack/Security/User/GetUserAccessToken/GetUserAccessTokenResponse.cs
@@ -2,7 +2,9 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
+using System.Collections.Generic;
 using System.Runtime.Serialization;
+using Elasticsearch.Net;
 
 namespace Nest
 {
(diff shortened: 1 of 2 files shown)


src/Nest/XPack/Security/User/GetUserAccessToken/GetUserAccessTokenResponse.cs
```diff
@@ -19,5 +19,8 @@ public class GetUserAccessTokenResponse : ResponseBase
 
 		[DataMember(Name ="type")]
 		public string Type { get; set; }
+
+		[DataMember(Name = "authentication")]
+		public AuthenticateResponse Authentication { get; set; }
```
Mpdreamz, line 24: While nice I am not sure this is the way to go, since this also inherits a lot from `ResponseBase` that makes no sense (e.g `.ApiCallDetails`, `.DebugInformation`).

Mpdreamz (changes requested):

src/Nest/XPack/Security/User/GetUserAccessToken/GetUserAccessTokenResponse.cs
stevejgordon: That's fair enough and I do agree it should probably be its own type for this purpose. Will apply that change.

Mpdreamz (approved):
LGTM!

PR #5380, opened 2021-03-02
Always include ReferenceAssemblies

Mpdreamz:
Due to difference in how design time and the dotnet CLI produce
package.json files.

This stabalizes the package json between builds


stevejgordon (approved):
LGTM

PR #5404, opened 2021-03-15
Handle "NaN" in sampling fields

### LINKED ISSUE #5007: ExtendedStats with one document -> UnexpectedElasticsearchClientException
7.9.0
**Elasticsearch version**:
7.9.1

Using ExtendedStats on a numeric value on one document generates always an exception:

```
    "message": "An error has occurred.",
    "exceptionMessage": "expected:'Number Token', actual:'\"NaN\"', at offset:179002",
    "exceptionType": "Elasticsearch.Net.UnexpectedElasticsearchClientException",
```

ps: I really don't know if the issue was present before, actually we working usually with a lot of documents, so this is the first time we see only one document after filtering our index and the first time in years that we see this exception.

**Expected behavior**
No exception but an empty result.

russcam:
Fixes #5007

### DIFF
--- a/src/Nest/Aggregations/AggregateFormatter.cs
+++ b/src/Nest/Aggregations/AggregateFormatter.cs
@@ -635,6 +635,8 @@ namespace Nest
 			extendedStatsMetric.SumOfSquares = reader.ReadNullableDouble();
 			reader.ReadNext(); // ,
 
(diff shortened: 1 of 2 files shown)


stevejgordon (approved):
LGTM! Thanks @russcam

PR #5409, opened 2021-03-16
Fix make assemble to produce zip

Mpdreamz:
- Updates make assemble to produce a zip file
- make sure the zip file is flattened


stevejgordon (approved):
LGTM!

PR #5412, opened 2021-03-16
Support case_insensitive on term-based queries

### LINKED ISSUE #5407: Couln'nt find case_insensitive option in High Level Client Term Query
Hi,
I am using ElasticSearch version 7.11. My TermQuery only returns case-sensitive matching results,to avoid that I set case_insensitive as true in DSL term query like below and it returns data as expected.

```
{
    "query": {
        "term": {
            "title": {
                "value": "BIG THEORY",
                "case_insensitive": true
            }
        }
    }
}
```
But when I am trying to implement in code using High Level Client [Nest], could'nt find the option to set case_insensitive as true.

Is there any way to set it as true?

Thanks in Advance

stevejgordon:
Resolves #5407

Note that `ISpanTermQuery` can no longer derive from `ITermQuery` as it doesn't support the `case_insenstive` option. This now manually includes and implements the value property.

### DIFF
--- a/src/Nest/QueryDsl/Span/Term/SpanTermQuery.cs
+++ b/src/Nest/QueryDsl/Span/Term/SpanTermQuery.cs
@@ -21,15 +21,18 @@ namespace Nest
 	{
 		public object Value { get; set; }
 		
-		protected override bool Conditionless => TermQuery.IsConditionless(this);
+		protected override bool Conditionless => IsConditionless(this);
 
 		internal override void InternalWrapInContainer(IQueryContainer c) => c.SpanTerm = this;
+
+		internal static bool IsConditionless(ISpanTermQuery q) => q.Value == null || q.Value.ToString().IsNullOrEmpty() || q.Field.IsConditionless();
 	}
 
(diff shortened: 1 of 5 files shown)


src/Nest/QueryDsl/Span/Term/SpanTermQuery.cs
```diff
@@ -9,18 +9,33 @@ namespace Nest
...
+		protected override bool Conditionless => IsConditionless(this);
 
 		internal override void InternalWrapInContainer(IQueryContainer c) => c.SpanTerm = this;
+
+		internal static bool IsConditionless(ISpanTermQuery q) => q.Value == null || q.Value.ToString().IsNullOrEmpty() || q.Field.IsConditionless();
 	}
 
-	public class SpanTermQueryDescriptor<T> : TermQueryDescriptorBase<SpanTermQueryDescriptor<T>, ISpanTermQuery, T>, ISpanTermQuery
-		where T : class { }
+	public class SpanTermQueryDescriptor<T> : FieldNameQueryDescriptorBase<SpanTermQueryDescriptor<T>, ISpanTermQuery, T>, ISpanTermQuery
```
Mpdreamz, line 31: This is technically breaking but worth doing IMO

Mpdreamz (approved):

PR #5413, opened 2021-03-16
Add explicit stop method to Auditable

### LINKED ISSUE #5363: `Audit.Ended` value is not set for events coming through `RequestPipelineDiagnosticObserver`
**NEST/Elasticsearch.Net version**: 7.11.1

**Elasticsearch version**: 7.11.1

Diagnostic events that are coming through `RequestPipelineDiagnosticObserver` have their `Audit.Ended` values not set.

**Steps to reproduce**:
 1. Provide diagnostic listener and subscribe to `DiagnosticSources.RequestPipeline.SourceName` event using `RequestPipelineDiagnosticObserver` with an `Action` for `onNextEnd`.
 2. Create a new instance of `ElasticLowLevelClient` and make a request to Elastic, i.e. `IndexAsync`.
 3. Check incoming `IApiCallDetails` payloads and their `AuditTrail` objects.

**Expected behavior**
`Audit.Ended` value should be set correctly.

----

I'm the maintainer of https://github.com/romansp/MiniProfiler.Elasticsearch. My little plugin pushes audit events from `ElasticClient` or (its low-level version) into MiniProfiler as custom timings.

I would like to provide one more way of integrating via diagnostic listeners. It works fine but incoming `AuditTrail` events don't have a correct `Ended` value so I can't compute how much time did request take.

I did some digging into transport implementation here: https://github.com/elastic/elastic-transport-net/blob/85c235850a126a161560d4754941db377ca1c7e9/src/Elastic.Transport/Components/Pipeline/RequestPipeline.cs#L173. Seems that `IApiCallDetails` is pushed into diagnostics observables _before_ `Auditable` gets disposed thus its `Ended` isn't set yet.

Though events that are coming through `ConnectionSettings.OnRequestCompleted` are correct.

<details>
  <summary>Silly repro</summary>

  ```c#
internal class TestDiagnosticListener : IObserver<DiagnosticListener>, IDisposable {
    private ConcurrentBag<IDisposable> Disposables { get; } = new ConcurrentBag<IDisposable>();

    public Action<IApiCallDetails> OnEnded { get; }

    public TestDiagnosticListener(Action<IApiCallDetails> onEnded) {
        OnEnded = onEnded;
    }

    public void OnError(Exception error) { }
    public void OnCompleted() { }

    public void OnNext(DiagnosticListener value) {
        TrySubscribe(DiagnosticSources.RequestPipeline.SourceName,
            () => new RequestPipelineDiagnosticObserver(null, v => OnEnded(v.Value)), value);
    }

    private void TrySubscribe(string sourceName, Func<IObserver<KeyValuePair<string, object>>> listener, DiagnosticListener value) {
        if (value.Name != sourceName) return;
        var d = value.Subscribe(listener());

        Disposables.Add(d);
    }

    public void Dispose() {
        foreach (var d in Disposables) {
            d.Dispose();
        }
    }
}

[Fact]
public async Task DiagnosticListener_AuditTrailIsValid() {
    // Arrange
    using var listener = new TestDiagnosticListener((data) => {
        var auditTrailEvent = data.AuditTrail[0];
        Assert.True((auditTrailEvent.Ended - auditTrailEvent.Started) >= TimeSpan.Zero);
    });
    using var foo = DiagnosticListener.AllListeners.Subscribe(listener);
    var connectionPool = new SingleNodeConnectionPool(new Uri("http://localhost:9200"));
    var settings = new ConnectionConfiguration(connectionPool, new InMemoryConnection());

    var client = new ElasticLowLevelClient(settings);
    var person = new { Id = "1" };

    await client.IndexAsync<BytesResponse>("test-index", PostData.Serializable(person));
}
  ```
</details>

stevejgordon:
Resolves #5363

Allows consumers of `Auditable` to explicitly call `Stop` to set the ended `DateTime` on the `Audit` object. This ensures that by the time this is available in a DiagnosticsListener `Audit` has a valid end `DateTime` set. Otherwise, it is the default until the Auditable is disposed.

### DIFF
--- a/src/Elasticsearch.Net/Auditing/Auditable.cs
+++ b/src/Elasticsearch.Net/Auditing/Auditable.cs
@@ -21,8 +21,7 @@ namespace Elasticsearch.Net
 			_dateTimeProvider = dateTimeProvider;
 			var started = _dateTimeProvider.Now();
 
-			_audit = new Audit(type, started);
-			_audit.Node = node;
+			_audit = new Audit(type, started) { Node = node };
 			auditTrail.Add(_audit);
 			var diagnosticName = type.GetAuditDiagnosticEventName();
 			_activity = diagnosticName != null ? DiagnosticSource.Diagnose(diagnosticName, _audit) : null;
@@ -43,9 +42,11 @@ namespace Elasticsearch.Net
 			set => _audit.Path = value;
 		}
 
+		public void Stop() => _audit.Ended = _dateTimeProvider.Now();
+
 		public void Dispose()
 		{
-			_audit.Ended = _dateTimeProvider.Now();
+			_audit.Ended = _audit.Ended == default ? _dateTimeProvider.Now() : _audit.Ended;
 			_activity?.Dispose();
 		}
 	}
--- a/src/Elasticsearch.Net/Transport/Pipeline/RequestPipeline.cs
+++ b/src/Elasticsearch.Net/Transport/Pipeline/RequestPipeline.cs
@@ -161,6 +161,7 @@ namespace Elasticsearch.Net
 					var response = _connection.Request<TResponse>(requestData);
 					d.EndState = response.ApiCall;
 					response.ApiCall.AuditTrail = AuditTrail;
+					audit.Stop();
 					ThrowBadAuthPipelineExceptionWhenNeeded(response.ApiCall, response);
 					if (!response.ApiCall.Success) audit.Event = requestData.OnFailureAuditEvent;
 					return response;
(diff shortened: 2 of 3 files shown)


src/Elasticsearch.Net/Auditing/Auditable.cs
```diff
@@ -43,9 +42,11 @@ public string Path
 			set => _audit.Path = value;
 		}
 
+		public void Stop() => _audit.Ended = _dateTimeProvider.Now();
```
Mpdreamz, line 45: LGTM, do we want this to be called multiple times to override or not? 

I think its fine as is.

Mpdreamz (approved):

PR #5417, opened 2021-03-20
enhancement/ienumerable searchafter descriptor

### LINKED ISSUE #5416: SearchAfter method overload with params argument translated into wrong json request
7.6.1

**Elasticsearch version**:
7.6.0

Usage of method overload `SearchAfter(params object[] searchAfter)` results in assignment of array of arrays to `search_after` field in request.

It looks like this
```
    "search_after": [
        [
            ... sorts ...
        ]
    ]
```
Intead of being like this
```
    "search_after": [
            ... sorts ...
    ],
```
As elastic API expects
https://www.elastic.co/guide/en/elasticsearch/reference/7.11/paginate-search-results.html#search-after

As result API returns this error
```
Elasticsearch.Net.ElasticsearchClientException: Request failed to execute. Call: Status code 400 from: POST /{index-name}/_search?typed_keys=true. ServerError: Type: parsing_exception Reason: "Expected [VALUE_STRING] or [VALUE_NUMBER] or [VALUE_BOOLEAN] or [VALUE_NULL] but found [START_ARRAY] inside search_after."
```

**Steps to reproduce**:
You can reproduce this case by passing any type except IList in order to trigger `SearchAfter(params object[] searchAfter)` overload.

Having 
```
class Doc
{
    public DateTime AddDate { get; set; }
}
```

This will use overload with `SearchAfter(IList<object> searchAfter)` and will generate correct request.
```
ISearchRequest SetupSearchRequest<T>(
    SearchDescriptor<T> descriptor, 
    IReadOnlyCollection<object> cursor) where T: Doc
{
    descriptor = descriptor.Sort(s => s.Ascending(f => f.AddDate)).Size(10);
    if (cursor != null)
    {
        descriptor = descriptor.SearchAfter(cursor.ToList()); //explicit cast to list
    }
    return descriptor;
}
```

This will use overload `SearchAfter(params object[] searchAfter)` and will generate incorrect request
```
ISearchRequest SetupSearchRequest<T>(
    SearchDescriptor<T> descriptor, 
    object cursor) where T: Doc
{
    descriptor = descriptor.Sort(s => s.Ascending(f => f.AddDate)).Size(10);
    if (cursor != null)
    {
        descriptor = descriptor.SearchAfter(cursor);
    }
    return descriptor;
}
```

stevejgordon:
Fixes #5416.

Since the sort from a previous search is represented as `IReadOnlyCollection<object>` it's reasonable that consumers may want to pass this unaltered into the next search. The `SearchRequest` defines this property as an `IList<object>` which we can't address without breaking changes. However, this PR can improve the fluent usage by providing an `IEnumerable<object>` overload, avoiding the need to call `ToList()` or `ToArray()` on the previous sort.

### DIFF
--- a/src/Nest/Search/Search/SearchRequest.cs
+++ b/src/Nest/Search/Search/SearchRequest.cs
@@ -453,6 +453,9 @@ namespace Nest
 			Assign(selector, (a, v) => a.Sort = v?.Invoke(new SortDescriptor<TInferDocument>())?.Value);
 
 		/// <inheritdoc cref="ISearchRequest.SearchAfter" />
+		public SearchDescriptor<TInferDocument> SearchAfter(IEnumerable<object> searchAfter) => Assign(searchAfter, (a, v) => a.SearchAfter = v?.ToListOrNullIfEmpty());
+
+		/// <inheritdoc cref="ISearchRequest.SearchAfter" />
 		public SearchDescriptor<TInferDocument> SearchAfter(IList<object> searchAfter) => Assign(searchAfter, (a, v) => a.SearchAfter = v);
 
 		/// <inheritdoc cref="ISearchRequest.SearchAfter" />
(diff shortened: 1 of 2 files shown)


Mpdreamz (approved):
LGTM

PR #5418, opened 2021-03-22
Update getting-started.asciidoc

LuboTheLupus:
The search query is not working with firstName, the prop in the person class is "FirstPerson"


stevejgordon:
Thanks for taking the time to contribute @LuboTheLupus. Unfortunately, this doesn't appear valid for a couple of reasons. This asciidoc is generated from `GettingStarted.doc.cs` which would be the correct place to make the change. However, there is no need to pascal case that value in the anonymous type as we expect the serialised field name to be camel case.

PR #5438, opened 2021-03-23
Introduce compatibility header

Mpdreamz:
Changes our default `Accept` and `Content-Type` request headers over to
the new vendor specific headers that includes compatible with
informations to start powering structurally versioned API's

```
Accept:       application/vnd.elasticsearch+json;compatible-with=7
Content-Type: application/vnd.elasticsearch+json; compatible-with=7
```

On 7.x Elasticsearch does not return the vendor type we send for
`Accept` but is still hardwired to return:

```
content-type: application/json; charset=UTF-8
```

We are therefor lenient in our Accept to Response ContentType
validations.

We do these assertions to prevent the client from de-serializing HTML
from possible proxies in the middle of the client and server.

NOTE this is for `7.13`

### DIFF
--- a/src/Elasticsearch.Net/Transport/Pipeline/RequestData.cs
+++ b/src/Elasticsearch.Net/Transport/Pipeline/RequestData.cs
@@ -13,10 +13,12 @@ namespace Elasticsearch.Net
 {
 	public class RequestData
 	{
-		public const string MimeType = "application/json";
-		public const string MimeTypeTextPlain = "text/plain";
 		public const string OpaqueIdHeader = "X-Opaque-Id";
 		public const string RunAsSecurityHeader = "es-security-runas-user";
+
+		public const string MimeTypeOld = "application/json";
+		public const string MimeTypeTextPlain = "text/plain";
+		public const string MimeType = "application/vnd.elasticsearch+json;compatible-with=7";
 
 		public RequestData(HttpMethod method, string path, PostData data, IConnectionConfigurationValues global, IRequestParameters local,
 			IMemoryStreamFactory memoryStreamFactory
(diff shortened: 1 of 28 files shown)


src/Elasticsearch.Net/Transport/Pipeline/RequestData.cs
```diff
@@ -151,6 +160,37 @@ IMemoryStreamFactory memoryStreamFactory
...
+
+		public static bool ValidResponseContentType(string acceptMimeType, string responseMimeType)
+		{
+			if (string.IsNullOrEmpty(acceptMimeType)) return false;
+			if (string.IsNullOrEmpty(responseMimeType)) return false;
+
+			// we a startswith check because the response can return charset information
+			// e.g: application/json; charset=UTF-8
+			if (acceptMimeType == RequestData.MimeTypeOld)
+				return responseMimeType.StartsWith(RequestData.MimeTypeOld);
```
stevejgordon, line 178: May want to force this as Ordinal-based since StartsWith and EndsWith are annoying culture-aware by default.

src/Elasticsearch.Net/Transport/Pipeline/RequestData.cs
```diff
@@ -151,6 +160,37 @@ IMemoryStreamFactory memoryStreamFactory
...
+			if (acceptMimeType == RequestData.MimeType)
+				// we check both vendored and nonvendored since on 7.x the response does not return a
+				// vendored Content-Type header on the response
+				return
+					responseMimeType == RequestData.MimeType
+					|| responseMimeType == RequestData.MimeTypeOld
+					|| responseMimeType.StartsWith(RequestData.MimeTypeOld)
+					|| responseMimeType.StartsWith(RequestData.MimeType);
+
+			return responseMimeType.StartsWith(acceptMimeType);
```
stevejgordon, line 190: Ordinal here too!

stevejgordon (changes requested):
Other than a couple of nits, LGTM!

PR #5440, opened 2021-03-23
Add support for datastream rollover api

ptasz3k:
Implement [manual datastream rollover API](https://www.elastic.co/guide/en/elasticsearch/reference/current/use-a-data-stream.html#manually-roll-over-a-data-stream)


stevejgordon:
@ptasz3k Thanks for taking the time to submit a PR. This API is already implemented and available under `Indices.Rollover` and `Indices.RolloverAsync`.

PR #5445, opened 2021-03-24
Introduce automated scheduled codegeneration

Mpdreamz:
- Allow codegenerator to be run from the root
- Allow codegen to be controlled through build.sh
- update command line argument parsing
- Paths are resolved relative so make sure we have a trailing / for absolute root
- Add github scheduled tasks to regenerate integration branches
- experiment with dynamic matrix


Mpdreamz:
Pair reviewed with @stevejgordon the code generations are failing because the target branches do not have the `make.sh` updates. 

:chicken: meets :egg:

PR #5478, opened 2021-03-25
Fixes the passing of args to assemble

Mpdreamz:
The difference a character makes.

Also added an assertion to our github action that `.ci/output` is
actually created


Mpdreamz:
Our unified build assemble run is happy:

https://github.com/elastic/elasticsearch-net/pull/5478/checks?check_run_id=2196435626

I added a check to ensure we actually produce output to `.ci/output`. Because passing args was broken the output was not produced in a folder that was mounted and thus never survived the `docker run`.

PR #5483, opened 2021-03-26
Move tests to rest-resources-zip.zip

Mpdreamz:
This moves our yaml test runner away from manually downloading the test files from github to sourcing the tests through the `rest-resources-zip.zip` artifact. 

We currently hardcode what is a platinum folder and what isn't. This still needs updating in the artifact and when it does we can switch our logic over.

This also moves our cleanup routine to the one @ezimuel extracted and formulized so we are line with th other clients.

@stevejgordon I stubbed but did not implement `wipeClusterSettings()` and therefor ignored the `cluster.put_settings` folder. This and all the other skips need to be re-evaluated.


stevejgordon (approved):
LGTM although I need to up my F# game for sure!

PR #5487, opened 2021-03-28
Add NOTICE file

Mpdreamz:
We already have license headers for the MIT projects we source includes

This PR adds a NOTICE.txt for our code and the code we source included


stevejgordon (approved):
LGTM

PR #5489, opened 2021-03-29
Update make.sh with skeleton from clients team

stevejgordon:
Updating to combine with the other items in the client's team skeleton.


Mpdreamz (approved):
Lgtm, i removed releasenotes from this task for now

Issue #5484, opened 2021-03-26
Make sure integrate `latest-7` works on `7.11`


Mpdreamz:
It's unlikely we do another 7.11 release but its one of the active integration branches.

Currently the `integrate latest 7` is broken (9 test failures)

https://github.com/elastic/elasticsearch-net/runs/2201937450

It would be good to turn this green so we are notified if `make assemble` turns red.

stevejgordon:
Resolved in #5504

PR #5504, opened 2021-03-31
Update 7.11 to support tests against latest-7

stevejgordon:
Not sure we want these lock changes but without them, it wouldn't compile locally for me.

Addresses #5484


stevejgordon:
Yep! It's not happy with the lock files. I'll pull those from this PR. I thought we'd seen the end of that!

PR #5558, opened 2021-04-20
Fixes incorrect reference in documentation

MEmanuelsson:
Fixes incorrect reference to "MyFirstCustomJsonNetSerializer" in documentation

### DIFF
--- a/docs/client-concepts/high-level/serialization/custom-serialization.asciidoc
+++ b/docs/client-concepts/high-level/serialization/custom-serialization.asciidoc
@@ -180,7 +180,7 @@ public class MyFirstCustomJsonNetSerializer : ConnectionSettingsAwareSerializerB
 }
 ----
 
-Using `MyCustomJsonNetSerializer`, we can serialize using
+Using `MyFirstCustomJsonNetSerializer`, we can serialize using
 
 * a Json.NET `NamingStrategy` that snake cases property names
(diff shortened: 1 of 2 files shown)


stevejgordon (changes requested):
This change will need to be made at source in `CustomSerialization.doc.cs` which is used to generate this asciidoc file using `.\build.bat documentation`.

stevejgordon:
Thanks for submitting this PR @MEmanuelsson. This is a generated file so the change would need to be made in `CustomSerialization.doc.cs` instead. If you're up for that, could you please amend this PR. You can run the docs generator afterwards to produce the asciidoc. If you're unable to make that change, no worries. Please let me know and I can close this and open a PR with the fix into the cs file.

MEmanuelsson:
No problem @stevejgordon, pushed an update to `CustomSerialization.doc.cs` just now. 

Ran `.\build.bat documentation` after the update to verify the change, this also results in some other asciidoc-files being updated but decided not to push these updates to the repo, right or wrong? 

image

stevejgordon:
This is perfect, thanks @MEmanuelsson! You made the correct call on the other generator changes. Some of those went into the branch yesterday and the deletions is a bug. I'll merge this in and handle backporting this to our other branches. Your help is very much appreciated.

Issue #5584, opened 2021-04-21
7.13.0 Meta Ticket


stevejgordon:
# Priority Items for 7.13.0

These are the priority items for 7.13.0. The need further review and refinement to identify if changes are required in the client to fully support these use cases.

- [x] Registered domain processor [#67611](https://github.com/elastic/elasticsearch/pull/67611) - **PR #5616**
- [x] Convert processor supports validation of IPv4/IPv6 addresses [#69989](https://github.com/elastic/elasticsearch/pull/69989) (issue: [#36145](https://github.com/elastic/elasticsearch/issues/36145)) - **PR #5588**
- [x] Introduce combined_fields query [#71213](https://github.com/elastic/elasticsearch/pull/71213) (issue: [#41106](https://github.com/elastic/elasticsearch/issues/41106)) - **PR #5619**
- [x] Network direction processor additions [#68712](https://github.com/elastic/elasticsearch/pull/68712) - **PR #5603**
- [x] Summary option for listing ingest pipelines without their definitions [#69756](https://github.com/elastic/elasticsearch/pull/69756) (issue: [#31954](https://github.com/elastic/elasticsearch/issues/31954))
- [x] Adds runtime_mappings to SQL requests [#71356](https://github.com/elastic/elasticsearch/pull/71356) (issue: [#68116](https://github.com/elastic/elasticsearch/issues/68116)) - **PR #5626**
- [x] Add calculated numeric fields [#69531](https://github.com/elastic/elasticsearch/pull/69531) - **PR #5622**
- [x] New queryable "_tier" metadata field [#69288](https://github.com/elastic/elasticsearch/pull/69288) (issue: [#68135](https://github.com/elastic/elasticsearch/issues/68135)) - **PR #5606**
- [x] Add positive_score_impact to rank_features type [#69994](https://github.com/elastic/elasticsearch/pull/69994) (issue: [#68619](https://github.com/elastic/elasticsearch/issues/68619)) - **PR #5607**
- [ ] ~~Add support for geo_line aggregation in pivot function [#69299](https://github.com/elastic/elasticsearch/pull/69299)~~
- [ ] ~~Support fetching _tier field value [#71379](https://github.com/elastic/elasticsearch/pull/71379) (issues: [#63569](https://github.com/elastic/elasticsearch/issues/63569), [#68135](https://github.com/elastic/elasticsearch/issues/68135))~~
[...]

PR #5616, opened 2021-04-23
Add registered domain ingest processor

stevejgordon:
Contributes to #5584


stevejgordon:
CI failure is an intermittent and unrelated issue.

PR #5658, opened 2021-04-30
Add NEST support for EQL Search API

stevejgordon:
Adds support for searching using EQL from the NEST client.

Contributes to #5584

@Mpdreamz - As this is a bit meaty, would you mind casting an eye over it? I opted not to create an interface for `EqlSearchResponse` and the types used for `Event` and `Sequence` which differs a little from the existing search code elsewhere.


stevejgordon:
Gonna assume the license header failure is a line endings thing. Will review on Tuesday.

Mpdreamz (approved):
LGTM, the inferfaces for `Search` are there to facilitate mocking. This is the only API we support mocking to an extend so no interfaces for the other responses makes sense here too.

PR #5659, opened 2021-05-01
Set no BOM in editorconfig

stevejgordon:
The `.editorconfig` change is applied by ReSharper code cleanup and avoid BOM changes affecting the check license headers action.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,6 +5,7 @@ trim_trailing_whitespace=true
 insert_final_newline=true
 
 [*]
+charset = utf-8
 indent_style = tab
 indent_size = 4
 


.editorconfig
```diff
@@ -5,6 +5,7 @@ trim_trailing_whitespace=true
 insert_final_newline=true
 
 [*]
+charset = utf-8
```
Mpdreamz, line 8: :heart: grave omission !

Mpdreamz (approved):
Almost LGTM, I don't think we need to change our header to the long form. We have explicit permission to use the short format given that's what most .NET repositories use (e.g the dotnet and microsoft repositories).

PR #5691, opened 2021-05-14
Fix ReSharper's "Convert 'as' expression type check and the following…

marodev:
Hello,

This pull request is _very simple_ and aims to reduce the technical debt. It specifically targets the "Convert 'as' expression type check and the following null check into pattern matching" [rule](https://www.jetbrains.com/help/resharper/UsePatternMatching.html) reported by ReSharper.

Let me know if you are interested in similar contributions.

### DIFF
--- a/src/DocGenerator/XmlDocs/XmlDocsVisitor.cs
+++ b/src/DocGenerator/XmlDocs/XmlDocsVisitor.cs
@@ -42,9 +42,7 @@ namespace DocGenerator.XmlDocs
 					_labeledListItem.Add(new Paragraph(content));
 				else
 				{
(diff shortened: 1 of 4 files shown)


stevejgordon:
Thanks for submitting this @marodev. Sorry for the delay in getting to this. I'm running the tests locally as I think the CI is a bit flaky at the moment. I'll merge if it looks good on my machine. 

I'm doing some major work right now to code-gen much of the vNext code, so it may be worth holding off on any further PRs for ReSharper suggestions. I tend to tackle these as I find them in specific files but with the codebase potentially changing a lot, the effort is probably not worthwhile at the moment.

stevejgordon (approved):
LGTM!

PR #5695, opened 2021-05-18
typo fixed - low priority

zeecorleone:
Observed a typo while reading documentation so wanted to fix it along


zeecorleone:
Signed agreement, but observed no change to existing PR

PR #5715, opened 2021-06-03
Typo Fixed (low priority) - on 7.x

zeecorleone:
fixed typo while reading documentation


stevejgordon:
Thanks for catching and reporting this @zeecorleone. The asciidoc files are actually generated from a corresponding C# file, in this case, `TermsQueryUsageTests.cs`. That requires amending the file and regenerating the docs from the CLI so it is a bit more involved. Did you want to try to tackle that? No worries if not, I can make the change and regen everything but I would need to close this PR.

PR #5728, opened 2021-06-15
[WIP] Pre-flight product check

stevejgordon:
Adds support for running a pre-flight check on first use to check the product and version.

This check is part of the request pipeline and runs once, before any other requests to check the validity of the product, issuing appropriate warnings and errors when failures are detected.


stevejgordon:
CI build is unrelated and fixed in #5729

glautrou:
This is a breaking change, not an enhancement.

After upgrading from NuGet "NEST 7.13.2" to "NEST 7.14.0" I get the following error:
`Elasticsearch.Net.UnsupportedProductException: The client noticed that the server is not a supported distribution of Elasticsearch.`

Associated code:
```
client.Ingest.PutPipeline(PipelineName, p => p
    .Description(PipelineDescription)
    .Processors(pr => pr
        .Attachment<GedIndex>(a => a
            .Field(f => f.ContenuBase64)
            .TargetField(f => f.Attachment)
        )
        .Remove<GedIndex>(r => r
            .Field(f => f
                .Field(j => j.ContenuBase64))
        )
    )
);
```
ES instance = v.7.9.2.

Note : I use the official ES [Ingest Attachment Processor Plugin](https://www.elastic.co/guide/en/elasticsearch/plugins/master/ingest-attachment.html).

jbrouwer1:
We have the same problem when updating Elasticsearch.Net to 7.14.0 and NEST to 7.14.0
Our AWS ES instance is v7.9 but what versions are supported for the new check to pass?
@stevejgordon

stevejgordon:
@glautrou With the `7.11` release of Elasticsearch, we no longer release oss-only distributions, as stated in the [FAQ on license change](https://www.elastic.co/pricing/faq/licensing). The recommendation is to upgrade to the latest default distribution of Elasticsearch, which is free to use under [Elastic License v2](https://www.elastic.co/licensing/elastic-license). Otherwise, you can pin the client version to <7.14.

We have labelled this as an enhancement since it affects only unsupported client and server combinations. In supported configurations, the change is not breaking. The intent here is to make this incompatibility explicit by failing fast to avoid consumers incorrectly assuming they are running in a supported configuration which is not tested and may not function as expected.

@jbrouwer1 The Elasticsearch clients go through rigorous testing and validation against the official Elasticsearch distribution coming from Elastic. Since the AWS Elasticsearch Service has incompatible APIs with Elasticsearch itself, either by missing APIs or has incompatible format, we do not support it. Please consider using [Elastic Cloud](https://cloud.elastic.co/) for an official and formally supported hosted Elasticsearch distribution.

PR #5755, opened 2021-06-29
Add readonly property to FS repository settings

### LINKED ISSUE #5753: Mechanism to set file system snapshot repository in ReadOnly mode
**NEST/Elasticsearch.Net version**: **7.x**
**Elasticsearch version**: **7.12**
Our application is using file system snapshot repository to take snapshot of index data. The location needs to be backup by third party backup solution. As Elasticsearch recommends to set repository in ReadOnly mode before backing up location. Our application is planning to set  that repository in ReadOnly mode; but there is no way to set snapshot repository in ReadOnly mode using NEST API. No 'ReadOnly' in 'FileSystemRepositorySettings' like there is in 'S3RepositorySettings'
Wanted to know way to set snapshot repository in backup mode before taking backup by third party solution

stevejgordon:
Fixes #5753


stevejgordon:
Unrelated CI failures which I'm investigating outside of this PR

PR #5844, opened 2021-07-12
Fix ApiUrlsLookup to include deprecated URLs

stevejgordon:


stevejgordon:
Manually porting to 7.x

PR #5902, opened 2021-07-15
Fix index exists response validity

### LINKED ISSUE #5901: Responses for `Indices.Exists` should be valid when a 404 is returned.
**NEST/Elasticsearch.Net version**: 7.13.2

When an index does not exist, the response should be valid when a 404 is returned. Currently, this is not the case.

**Steps to reproduce**:
 1. Send an `Indices.Exists` request to a non-existent index.

**Expected behavior**
The response should be valid and `Exists` should be false.

stevejgordon:
Fixes #5901


stevejgordon:
Not required

PR #5926, opened 2021-07-30
- then to than

eben-roux:


stevejgordon:
Good spot @eben-roux. Thanks for creating this PR. This ASCIIDOC is actually generated from a file (src/Tests/Tests/ClientConcepts/HighLevel/Inference/FieldInference.doc.cs) which is where this fix would need to be made. Once that's done, the docs can be re-generated. If you fancy it, would you mind making the change in that file? I can handle the re-gen once it's merged. Alternatively, I can port this and close this PR.

eben-roux:
I'd be happy to make the change in the source file :)

I'll submit a new PR for that change.

PR #5932, opened 2021-08-09
Update compatibility matrix to reflect supported versions

stevejgordon:


stevejgordon:
Clients CI failure is unrelated and needs further investigation.

PR #6003, opened 2021-09-22
Update index and add compatibility mode to connection section

stevejgordon:
- Update connection section of getting started
- Generate updated getting started doc
- Add release notes to index


stevejgordon:
Clients CI failure is unrelated.

PR #6015, opened 2021-10-07
Customize examples for 7.x

wrong42:
Remove not supported custom types from bulk insert and search paths.


stevejgordon:
@wrong42 Thanks for submitting this. Those files are generated by of a docs generator from source files. That process may change for 8.x work we are undertaking. Unfortunately, we cannot take this PR right now until we redesign a few things about that process. A major docs rework will happen with 8.0 before final release.

PR #6106, opened 2022-02-02
Update SharpZipLib

stevejgordon:


stevejgordon:
Test failures relate to the v2 index template API but don't occur locally. Suspect this is a timing issue. Will investigate outside of this PR.

PR #6124, opened 2022-02-09
Re-enable locked package restores

stevejgordon:
As titled.


stevejgordon:
CI failure is unrelated

PR #6171, opened 2022-03-23
[DOCS] Adds new Intro and Install chapters

szabosteve:
## Overview

This PR adds two new pages to the .NET documentation:
* Introduction
* Installation.

It also modifies: 
* the index file to point to the new pages instead of the old ones
* the section IDs in the old pages.

### Preview

* [Introduction]()
* [Installation]()

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -1,7 +1,7 @@
 :title-separator: |
 
 [[elasticsearch-net-reference]]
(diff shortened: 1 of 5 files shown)


stevejgordon (approved):
Thanks for this @szabosteve.

We can probably remove `low-level.asciidoc` entirely for v8 but let's pick that up in future work.

PR #6188, opened 2022-03-24
[DOCS] Removes low-level client docs and references

szabosteve:
## Overview

This PR removes the low-level client-related documentation pieces from .NET docs and also deletes the references to that book from the rest of the docs pieces.


szabosteve:
@elasticmachine test this please

docs/client-concepts/connection-pooling/building-blocks/connection-pooling.asciidoc
```diff
@@ -29,9 +29,7 @@ this is https://blogs.msdn.microsoft.com/adarshk/2005/01/02/understanding-system
 
 So, what is a connection pool in NEST responsible for? It is responsible for managing the nodes in an Elasticsearch
 cluster to which a connection can be made and there is one instance of an `IConnectionPool` associated with an
-instance of `ConnectionSettings`. Since a <<lifetimes,single client and connection settings instance is recommended for the
-life of the application>>, the lifetime of a single connection pool instance will also be bound to the lifetime
-of the application.
+instance of `ConnectionSettings`.
```
stevejgordon, line 32: I think we want to leave in the note about the lifetime of the pool here.

stevejgordon (changes requested):
One note about a removal I don't think we need. Otherwise LGTM thanks!

The naming for some of the types needs updating but I can follow up with a PR to start correcting those as necessary.

stevejgordon (approved):
Thanks!

Issue #6258, opened 2022-04-07
Version 6.x: Wrong JsonProperty name on AnalyzeToken. Expecting position_length when it should be positionLength


Ferran54:
**NEST/Elasticsearch.Net version**: 6.8.10
**Elasticsearch version**: 6.8

**Description of the problem including expected versus actual behavior**:
JsonProperty name is wrong
https://github.com/elastic/elasticsearch-net/blob/0aaf7c76fe24df9ca1d23df6c99256af005fecc8/src/Nest/Indices/Analyze/AnalyzeToken.cs#L14

![analyzerresponse](https://user-images.githubusercontent.com/25817096/162190198-db4988dd-e82b-495a-9510-7f98b434cd82.PNG)

The Eastic api is returning **positionLength** and NEst is expecting **position_length** so, NEst can't serialize the response properly and always return that null

**Steps to reproduce**:
 1.  call AnalyzeAsync on NEST 
 2. observe results are always returning PositionLength = null

**Expected behavior**
PositionLenght should be the same that is returned by the Elastic Analyzer api

stevejgordon:
Thanks for raising this and the PR to fix it @Ferran54. I have merged the PR and this morning I released `6.8.11` which includes your contribution.

PR #6259, opened 2022-04-07
Fix PositionLength JSON name on AnalyzeToken

Ferran54:
api return positionLength but nest is expecting a different name, so it is always null on nest response.


stevejgordon (approved):
Thanks for catching this and fixing it with this PR. LGTM!

stevejgordon:
Fixes #6258

stevejgordon:
Tests are passing locally so will ignore the CI issues here which are unrelated.

PR #6285, opened 2022-04-13
Add component template APIs

stevejgordon:
Adds code generates requests and responses for the follow component template APIs:

- Put Component Template
- Get Component Template
- Delete Component Template
- Exists Component Template


brenthompson2:
I don't see these APIs in the generated Cluster class for the v8 client. Did they get moved? Are they coming back?

https://github.com/elastic/elasticsearch-net/blob/main/src/Elastic.Clients.Elasticsearch/_Generated/Client/ElasticsearchClient.Cluster.g.cs

stevejgordon:
@brenthompson2 They will return. We dropped them in [the release](https://www.elastic.co/guide/en/elasticsearch/client/net-api/current/release-notes-8.0.0.html) as we found a few spec issues and needed time to port/introduce tests.

PR #6418, opened 2022-05-24
[DOCS] Adds Examples section to .NET docs

szabosteve:
## Overview

This PR adds an `Examples` section to the .NET documentation that contains code snippets showing how to execute CRUD operations.

### Preview

[Examples](https://elasticsearch-net_6418.docs-preview.app.elstc.co/guide/en/elasticsearch/client/net-api/master/examples.html)


docs/examples.asciidoc
```diff
@@ -0,0 +1,91 @@
...
+----
+var tweet = new Tweet
+{
+    Id = 1,
+    User = "kimchy",
+    PostDate = new DateTime(2009, 11, 15),
+    Message = "Trying out NEST, so far so good?"
+};
+
+var response = client.Index(tweet, idx => idx.Index("mytweetindex")); //or specify index via settings.DefaultIndex("mytweetindex");
```
stevejgordon, line 24: I'd like to drop an sync examples if possible. It's generally bad and for most modern situations not recommended. Could we default to the async example and maybe just mention once, the sync alternative if there's no alternative?

docs/examples.asciidoc
```diff
@@ -0,0 +1,91 @@
...
+All the calls have async variants:
+
+var response = client.IndexAsync(tweet, idx => idx.Index("mytweetindex")); // returns a Task<IndexResponse>
```
stevejgordon, line 27: We shouldn't show this example. Always show the `await`ed one.

stevejgordon (commented):
The API is a little in-flux still so we will need to revisit the samples. I can assist with filling in the update and delete documents. A few initial comments on know changes based on current alpha code.

stevejgordon (approved):
LGTM!

PR #6427, opened 2022-06-01
Update autolabel version on the 8.3 branch

Mpdreamz:
It took the version number from `main` when it was automatically branched. However the version number on main should have been `v8.3.0-alpha.1` and should be `v8.4.0-alpha.1` now


sethmlarson (approved):
Thanks!

PR #6479, opened 2022-06-28
Fix duplicate object serialization of aggregations

stevejgordon:
When using the object initializer syntax, aggregations can appear in the request twice. This is harmless but wastes bytes and has an added serialization overhead.

This PR ensures we retain the ability to deserialize from either name ("aggs" or "aggregations") but all serialization using "aggs" as standard.


stevejgordon:
Failures are unrelated.

PR #6510, opened 2022-07-11
Bump Newtonsoft.Json version

oldmine:
Upgrade Newtonsoft.Json to latest version
[Changelog](https://github.com/JamesNK/Newtonsoft.Json/releases/tag/13.0.1)

Add ability to use old elasticsearch and client version in new projects

### DIFF
--- a/build/scripts/scripts.fsproj
+++ b/build/scripts/scripts.fsproj
@@ -74,7 +74,7 @@
     <PackageReference Include="Fake.IO.Zip" Version="5.15.0" />
     <PackageReference Include="ILRepack" Version="2.1.0-beta1" />
     <PackageReference Include="FSharp.Core" Version="4.6.2" />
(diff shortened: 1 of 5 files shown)


oldmine:
Hi, @stevejgordon thank you for answer

I understand that 5.x version is no longer supported, but applications that use old Elasticsearch version and client is still exists
I can't reference a newer Newtonsoft.Json version in my project, because receive nuget error about NEST 5.6.6 require Newtonsoft.Json version in range 12.0.1 .. 13.0.0, but I try to install 13.0.1 (it's only one existing version with 13 major)

I think bump Newtonsoft.Json version in package helps many people who can't right now update Elasticsearch but want to use newer packages

stevejgordon:
@oldmine Ok I understand the problem. Unfortunately that is the trade-off of remaining on an out of support version. These versions will receive no new releases.

PR #6671, opened 2022-09-26
Sets the shared attribute ref_current to ref. (#6665)

stevejgordon:
Backport 2528bc01725e64bfc70a6cb147cd05183dcc04c0 from #6665


stevejgordon:
Seems this is not required as auto-backport worked.

PR #6808, opened 2022-10-19
Update ubuntu version for backport action

stevejgordon:
As titled.


stevejgordon:
Picked up the wrong commits

Issue #6647, opened 2022-08-23
BulkAll should classify retried and failed documents as dropped.


Mpdreamz:
BulkAll has an option to continue after dropped documents: https://github.com/elastic/elasticsearch-net/blob/b7dd9cf58c67cccc6d7f679b5883a155e112f5e2/src/Elastic.Clients.Elasticsearch/Helpers/IBulkAllRequest.cs#L17

HandleDroppedDocuments takes this setting into account and only throws an exception if we are not continueing after seeing dropped documents https://github.com/elastic/elasticsearch-net/blob/584edb8b4e8a9ec47bbb53d4f3c1a3a4c80db8f1/src/Elastic.Clients.Elasticsearch/Helpers/BulkAllObservable.cs#L173

However later after retrying we through if we still hold retryableDocuments. This should also check this configuration setting and treat them as dropped documents:

However later we don't classify retried to completion bulk items as dropped and do not make this check conditional: https://github.com/elastic/elasticsearch-net/blob/584edb8b4e8a9ec47bbb53d4f3c1a3a4c80db8f1/src/Elastic.Clients.Elasticsearch/Helpers/BulkAllObservable.cs#L178

PR #6815, opened 2022-10-19
Fix dropped doc callback for retried docs

stevejgordon:
Contributes to #6647


stevejgordon:
Failures are unrelated to this change

PR #6888, opened 2022-11-03
Apply codegenNames for union parameters

stevejgordon:
As titled.


stevejgordon:
Docs failure is unrelated

PR #7329, opened 2023-03-14
Add support for geo_bounding_box query.

stevejgordon:
As titled. Includes code-gen for multi-item unions to support geo types.


stevejgordon:
Test failure in actions seems unrelated.

PR #7539, opened 2023-04-04
Made ElasticsearchClient disposable, disposed of activity source

mlof:
In high throughput applications the ActivitySource does not get disposed of properly. By adding IDisposable to the ElasticsearchClient, this can hopefully be remedied.

### DIFF
--- a/src/Elastic.Clients.Elasticsearch/Client/ElasticsearchClient.cs
+++ b/src/Elastic.Clients.Elasticsearch/Client/ElasticsearchClient.cs
@@ -18,7 +18,7 @@ namespace Elastic.Clients.Elasticsearch;
 /// <summary>
 /// A strongly-typed client for communicating with Elasticsearch server endpoints.
 /// </summary>
-public partial class ElasticsearchClient
+public partial class ElasticsearchClient : IDisposable
 {
 	private readonly HttpTransport<IElasticsearchClientSettings> _transport;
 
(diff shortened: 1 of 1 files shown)


stevejgordon:
Hi, @mlof. I don't think we should make this change before reviewing the scenario more closely. 

Strictly speaking, yes, the client technically should dispose of the ActivitySource, but the client is intended to be used as a singleton for the lifetime of applications. We don't anticipate more than one client being used (per cluster). Are you able to open an issue describing the usage scenario where you encounter problems?

Making it disposable starts to get us into the same set of problems HttpClient experiences where confusion over its intended use occurs because it is `IDisposable`.

mlof:
While I don't necessarily agree (When used in a DI pipeline it ensures that the cleanup happens correctly on application shutdown) looking a bit further in the memory dumps it's actually an underlying problem. 
I don't have 36 million instances of `new ActivitySource("Elastic.Clients.Elasticsearch.ElasticsearchClient")`. They're actually sneaky `new ActivitySource("Elastic.Transport.RequestPipeline")` instances. Bears looking into, but i'll make sure the problem isn't just on my end before making a pull request. 😅

mlof:
If I can help in any way, feel free to reach out. For now, I suppose I can close the PR.

stevejgordon:
Thanks for highlighting this issue @mlof. I will have PRs in to fix this today and likely get an 8.0.10 release out tomorrow which should avoid the issue.

mlof:
Brilliant. Thank you.

stevejgordon:
@mlof I've released [8.0.10](https://github.com/elastic/elasticsearch-net/releases/tag/8.0.10) if you want to try it and verify it resolves the allocation of `ActivitySource` instances.

mlof:
So far so good!

PR #7731, opened 2023-05-19
No allocations in `ResponseItem.IsValid` property

YohDeadfall:
`ToLower`/`ToUpper` methods are allocating which isn't good for properties which imply performance and no allocations (lazy is fine, but not repeating as here). For `Status` check I used `is` keyword which under the hood creates a local variable. That's not necessary since `Status` is a simple property with a backing field and JIT would be happy to read its value once into a register (different on .NET Fx). So let's think about it as a compensation for added lines by removing a some characters.


technige:

flobernd (approved):
LGTM! Thanks!

PR #7733, opened 2023-05-19
Fixed the equality contract on Metrics type

YohDeadfall:
`HashSet<T>` doesn't override `GetHashCode`. Therefore, the default provided algorithm for reference types is used which doesn't take fields into account as it happens for value types. That breaks the equality contracts on Metrics.

Ideally `HashCode` type should be used for more correct computations with less number of collisions, but that requires increasing the minimal target framework.

`SetEqual` is available since .NET Fx 3.5 and is destined for equality comparison of two different hash set instances. Since both instances of the `Metrics` type has the same default comparer, the method works as a sugar over a length check followed by iteration over one set and probing the second one for containing the current element through the `Contains` method.

### DIFF
--- a/src/Elastic.Clients.Elasticsearch/Core/Infer/Metric/Metrics.cs
+++ b/src/Elastic.Clients.Elasticsearch/Core/Infer/Metric/Metrics.cs
@@ -70,7 +70,16 @@ public sealed class Metrics : IEquatable<Metrics>, IUrlParameter
 	}
 
 	/// <inheritdoc />
-	public override int GetHashCode() => Values != null ? Values.GetHashCode() : 0;
+	public override int GetHashCode()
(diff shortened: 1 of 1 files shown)


technige:

flobernd:
Thanks as well for this contribution!

It might be worth adding a conditional reference to `Microsoft.Bcl.HashCode` for the `net462` and `netstandard2.0` targets. There are around ~150 custom `GetHashCode` implementations in the project (a lot of them are auto-generated) that could benefit from using `HashCode`. I will check internally, if this is something we might consider.

For now I'm going to approve the PR to have this fix included in the next patch release. The `HashCode` optimization can be done later.

flobernd (approved):

PR #7735, opened 2023-05-22
Removed unused `JsonIgnore`

YohDeadfall:
Since the `Field` type has a JSON converter specified through an attribute, member attributes are ignored. Therefore, `JsonIgnore` has no effect at all and can be removed.


technige:

flobernd (approved):
LGTM! Thanks!

YohDeadfall:
You're welcome, but to complete it pipelines have to be allowed to run since I contributed here first time.

PR #7736, opened 2023-05-22
Optimized `FieldConverter`

YohDeadfall:
This pull request has three commits:

1. Optimizes read and writes using prencoded property names:
1.1. Reading doesn't allocate and decode a JSON property name.
1.2. Writing doesn't encode property names.
2. Writing has two length checks less, one for each property name write since there's a limit for non-encoded literals.
3. Since any interpolated string using static values is preprocessed and turned into a constant string type names where replaced by `nameof(Field)` expressions.
4. `null` values are handled by the writer itself ([see this](https://github.com/dotnet/runtime/blob/bdf5df39c4749dd88cf45de008f9f3c2d5905fba/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonConverterOfT.cs#L188-L189)).

### DIFF
--- a/src/Elastic.Clients.Elasticsearch/Core/Infer/Field/FieldConverter.cs
+++ b/src/Elastic.Clients.Elasticsearch/Core/Infer/Field/FieldConverter.cs
@@ -96,10 +96,8 @@ internal sealed class FieldConverter : JsonConverter<Field>
 		else
 		{
 			writer.WriteStartObject();
-			writer.WritePropertyName(FieldProperty);
-			writer.WriteStringValue(fieldName);
-			writer.WritePropertyName(FormatProperty);
-			writer.WriteStringValue(value.Format);
+			writer.WriteString(FieldProperty, fieldName);
+			writer.WriteString(FormatProperty, value.Format);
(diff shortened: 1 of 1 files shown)


technige:

flobernd (approved):
LGTM! Thanks!

src/Elastic.Clients.Elasticsearch/Core/Infer/Field/FieldConverter.cs
```diff
@@ -48,19 +51,19 @@ private static Field ReadObjectField(ref Utf8JsonReader reader)
...
-				var propertyName = reader.GetString();
-				reader.Read();
-
-				switch (propertyName)
+				if (reader.ValueTextEquals(FieldProperty.EncodedUtf8Bytes))
+				{
+					reader.Read();
+					field = reader.GetString();
+				}
+				else if (reader.ValueTextEquals(FormatProperty.EncodedUtf8Bytes))
```
YohDeadfall, line 59: For your info, I proposed an ability to pass an encoded text directly to the reader, but it was declined. See dotnet/runtime#86591.

PR #7763, opened 2023-06-29
Bump SDK and tooling version to `net6`

flobernd:
This PR bumps the SDK version to `6.0.4**` and changes the tooling target-framework to `net6` accordingly.

Solves an issue that prevented `Microsoft.CodeCoverage` from being correctly restored.


flobernd:
The failed test reports are from previous integration test runs (these had random failures, but succeeded in a second run).

stevejgordon (approved):
LGTM

PR #7769, opened 2023-07-04
Add Windows test workflow and net462 test target

flobernd:


tests/Tests.ClusterLauncher/Tests.ClusterLauncher.csproj
```diff
@@ -1,9 +1,15 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>net462;netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
     <CheckEolTargetFramework>false</CheckEolTargetFramework>
   </PropertyGroup>
+  <PropertyGroup>
```
flobernd, line 7: Workaround for:

```
rror NU1004: The project's runtime identifiers have changed from. Project's runtime identifiers: , lock file's runtime identifiers win7-x86.The packages lock file is inconsistent with the project dependencies so restore can't be run in locked mode. Disable the RestoreLockedMode MSBuild property or pass an explicit --force-evaluate option to run restore to update the lock file.
```

This is a known long standing issue (see comments).

Another workaround that does not involve disabling lock files, would be this one:
https://github.com/NuGet/Home/issues/9195#issuecomment-1547896705

tests/Tests.Core/Xunit/ElasticsearchXunitRunOptions.cs
```diff
@@ -29,7 +31,16 @@ public ElasticsearchXunitRunOptions()
 			Generators.Initialize();
 		}
 
-		public override void OnBeforeTestsRun() => TestConfiguration.Instance.DumpConfiguration();
+		public override void OnBeforeTestsRun()
+		{
+			VerifierSettings.DerivePathInfo(
```
flobernd, line 36: `[ModuleInitializer]` is not available prior to NET5.

This snippet is moved from there.

tests/Tests.Domain/Tests.Domain.csproj
```diff
@@ -11,7 +11,7 @@
 
   <ItemGroup>
     <PackageReference Include="Bogus" Version="22.1.2" />
-    <PackageReference Include="Elastic.Elasticsearch.Managed" Version="0.3.5" />
+    <PackageReference Include="Elastic.Elasticsearch.Managed" Version="0.4.1" />
```
flobernd, line 14: See https://github.com/elastic/elasticsearch-net-abstractions/pull/58 for details.

tests/Tests/_VerifySnapshots/FunctionScoreQueryUsageTests.VerifyDescriptorJson.verified.txt
```diff
@@ -1,7 +1,7 @@
 {
   query: {
     function_score: {
-      boost: 1.1,
```
flobernd, line 4: NETFX is not fully IEEE compliant which messes with our roundtrip serialization tests. Particularily `float` values seem to cause problems. As a workaround I changed these tests to use values that can be represented as integers.

More details: https://github.com/dotnet/runtime/issues/435#issuecomment-560503912

PR #7811, opened 2023-07-11
Fix localization issue with floating-point deserialization from string

flobernd:
Fixes deserialization of "floating-point strings" on machines with German localization:

```
 FAIL  Tests.Serialization.SourceSerializationForNumericPropertiesTests.DoubleAsString_DeserializesCorrectly
       Expected deserialized.Double to be 1.0, but found 10.0.
..........................................................................
 FAIL  Tests.Common.Fields.FieldValueTests.CanSerialize_DoubleFieldValueKind
       Expected result.FieldValue.ToString() to be "1.1", but "1,1" differs near ",1" (index 1).
```


stevejgordon (approved):
LGTM

PR #7835, opened 2023-07-12
Regenerate client using 8.9 specification

flobernd:


Mpdreamz (changes requested):
Do we maintain an allow/block list to include API's? 

Do we want to include all the new API's that are coming in from this PR?

In the past we would add each API in follow up commits with integration tests.

Mpdreamz (approved):

flobernd:
I've created this from the wrong base branch. Will open another PR soon.

PR #7880, opened 2023-08-08
Refactor client overloads

stevejgordon:
- Simplify overloads and avoid repeated code
- Ensure we pass the cancellation token to transport

We had a chunk of repeated code which was hard to keep in sync. As these are all internal methods, I've restructured the overloads and removed a redundant one. Currently, we don't use the `forceConfiguration` overload, which we should probably review and either remove those overloads or reintroduce NEST force config behaviour.

This also fixes a bug where we don't always pass the cancellation token on transport calls making it impossible for some requests to be cooperatively cancelled.


flobernd (approved):
LGTM! Nice refactoring, thanks!

PR #7883, opened 2023-08-09
Remove leading `/` character from API urls

### LINKED ISSUE #7878: Relative URI path of Elasticsearch is removed from the base URI
**Elastic.Clients.Elasticsearch version**: 8.9.0

**Elasticsearch version**: 8.8.1

**.NET runtime version**: 6.0

**Operating system version**: Windows 10

We have Elasticsearch hosted on a relative path, such as: https://mysite.com/elastic/
When we try to make any search request from the .NET client, we are getting 404 error because the client uses only the base uri: https://mysite.com/

**Steps to reproduce**:
 1.  Set URI of elasticsearch: `var uri = new Uri("https://mysite.com/elastic/");`
 2. Initialize node pool: `var nodePool = new SingleNodePool(uri);`
 3. Initialize settings: `var settings = new ElasticsearchClientSettings(nodePool, ...);`
 4. Initialize client: `var client = new ElasticsearchClient(settings);`
 5. Make any request, e.g. `var info = client.Info();`
 In the response (via debug information enabled, and via OnRequestCompleted logging) we can see that the requested URL was: GET https://mysite.com/?pretty=true&error_trace=true (_/elastic/_ path is missing).

**Expected behavior**
We expect the .NET client not to alter the original elasticsearch URI path provided.

**More info**
In the old NEST Elastic .NET Client (e.g. v7.15.0) this bug doesn't occur. I found this connected issue, which deals with the very same bug we experience: https://github.com/elastic/elasticsearch-net/issues/3691

I believe the bug is connected to Uri construction, see Microsoft documentation remarks [here](https://learn.microsoft.com/en-us/dotnet/api/system.uri.-ctor?view=net-7.0#system-uri-ctor(system-uri-system-string)):

> if the relativeUri begins with a slash, then it will replace any relative part of the baseUri

In the NEST v7 client, _ApiUrlsLookups_ class contains all paths **without** leading slash (thus it won't replace the relative part of the base uri): 

image

However, in the current v8 client, _ApiUrlLookup_ class contains all paths **with** leading slash (thus it will replace the relative part of the base uri):

image

flobernd:
Closes #7878


flobernd:
@zsonglorman Hi! I planned a release for tomorrow 🙂

flobernd:
@zsonglorman 8.9.2 has been released on NuGet

PR #7894, opened 2023-08-17
Reintroduce suggestion feature

### LINKED ISSUE #7390: Add "suggest" to SearchResponse / Error during CompletionSuggester (8.x Client)
1. It looks like the `SearchResponse` is missing serialization for a suggestion response yet, e.g. https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters.html#querying

2. Usage of the `Prefix` property at  `CompletionSuggester` level leads to an error -> `[9:9] [completion] unknown field [prefix]`

Not working example:
```
new CompletionSuggester
{
    Field = "myfield", 
    Size = 20, 
    Prefix = "Act",
    SkipDuplicates = true,
}
```

Working example:
```
var suggester = FieldSuggester.Completion(new CompletionSuggester
{
    Field = "myfield", 
    Size = 20, 
    SkipDuplicates = true,
});
suggester.Prefix = "Act";
```

flobernd:
Implements a hand-crafted version of `SuggestDictionary<TDocument>` and adds the - slightly modified - generated "suggest" variants.

The next steps will involve improving the generator in order to generically support these constructs (aggregations and suggestions).

Closes #7390


src/Elastic.Clients.Elasticsearch/Serialization/SingleOrManyCollectionConverter.cs
```diff
@@ -16,4 +16,6 @@ internal class SingleOrManyCollectionConverter<TItem> : JsonConverter<ICollectio
 
 	public override void Write(Utf8JsonWriter writer, ICollection<TItem> value, JsonSerializerOptions options) => 
 		SingleOrManySerializationHelper.Serialize<TItem>(value, writer, options);
+
+	public override bool CanConvert(Type typeToConvert) => true;
```
stevejgordon, line 20: We should probably have `CanConvert` check the type implements ICollection<T> here.

src/Elastic.Clients.Elasticsearch/Types/Core/Search/SuggestDictionary.cs
```diff
@@ -0,0 +1,215 @@
...
+
+using Elastic.Clients.Elasticsearch.Serialization;
+
+namespace Elastic.Clients.Elasticsearch.Core.Search;
+
+[GenericConverter(typeof(SuggestDictionaryConverter<>), unwrap:true)]
+public sealed partial class SuggestDictionary<TDocument> :
+	IsAReadOnlyDictionary<string, IReadOnlyCollection<ISuggest>>
+{
+	public SuggestDictionary(IReadOnlyDictionary<string, IReadOnlyCollection<ISuggest>> backingDictionary) :
```
stevejgordon, line 19: Could we make this ctor internal as we only expect to deserialize into this type ourselves and create it from the converter?

stevejgordon (changes requested):
LGTM with a couple of minor suggestions.

src/Elastic.Clients.Elasticsearch/Types/Core/Search/SuggestDictionary.cs
flobernd: The code generator seems to generate most constructors as `public` at the moment. I tried to keep this hand-crafted class as close as possible to the generator output in order to avoid breaking changes in the future like we discussed. We could potentially improve the generator for outputting `protected/internal` constructors for types that are exclusively used responses.

src/Elastic.Clients.Elasticsearch/Serialization/SingleOrManyCollectionConverter.cs
stevejgordon: That's fair.

src/Elastic.Clients.Elasticsearch/Types/Core/Search/SuggestDictionary.cs
stevejgordon: Also true. For some of these types used only on responses, it shouldn't generate them with public ctors. We try to determine which types are used where in the generator so we can make response-only types read-only and init-only. I probably missed that in some places for these special cases. Personally, I'd start with it internal as making it public later if that happens in code-gen is non-breaking.

flobernd: Sounds good! Will change it to `internal` for now.

PR #7982, opened 2023-11-10
[CI] Sets up Buildkite configuration

picandocodigo:
It looks like it's running the same build as in Jenkins. `run-elasticsearch` is failing for `free` in both Jenkins and Buildkite, but it runs ok for `platinum`. It's probably some setup issue missing at the moment with security in 8.x. I'm not sure there are any tests actually running though, but we can look into that and once it's working backport this to 7.x.


flobernd (approved):
LGTM! Thanks a lot 🙂

PR #8052, opened 2024-03-07
Update TargetFrameworks

### LINKED ISSUE #8050: Remove netcoreapp3.1 and net5.0 from TargetFrameworks and add net8.0
**Is your feature request related to a problem? Please describe.**
Currently, I see in the csproj files `<TargetFrameworks>netstandard2.0;net461;netstandard2.1;netcoreapp3.1;net5.0;net6.0</TargetFrameworks>`

**Describe the solution you'd like**
Remove `netcoreapp3.1` and `net5.0` and add `net8.0`.

**Describe alternatives you've considered**

**Additional context**

gpetrou:
Resolves #8050.

### DIFF
--- a/.buildkite/DockerFile
+++ b/.buildkite/DockerFile
@@ -1,5 +1,7 @@
-ARG DOTNET_VERSION=6.0.403
+ARG DOTNET_VERSION=8.0.100
 FROM mcr.microsoft.com/dotnet/sdk:${DOTNET_VERSION} AS elasticsearch-net-build
+
+ENV NUGET_SCRATCH="/tmp/NuGetScratch"
 
 ARG USER_ID
 ARG GROUP_ID
--- a/.buildkite/DockerFile
+++ b/.buildkite/DockerFile
@@ -44,6 +44,7 @@ RUN chown -R $GROUP_ID:$USER_ID /tmp/NuGetScratch
 COPY . .
 
 # making sure enough git info is available inside the container
+RUN git config --global --add safe.directory '*'
 RUN git rev-parse HEAD .
 
 # USER user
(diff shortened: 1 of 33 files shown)


gpetrou:
I used `nuget restore -ForceEvaluate` to update the packages.lock.json files.

flobernd:
Seems like the "Unified Release / Assemble" workflow is failing because of the Docker container using the wrong .NET SDK version.

flobernd:
Thanks for updating!

I think there is/was a bug in the way we pass the argument in the Docker commandline:
https://github.com/elastic/elasticsearch-net/blob/4a6b9e8ebb443f26d3ce9cdfa494292397b4052c/.ci/make.sh#L136

flobernd:
Thanks again for updating. Seems like we are going down a rabbithole... The newer SDKs seem to use different temp paths which causes building the docker image to fail.

gpetrou:
> The newer SDKs seem to use different temp paths which causes building the docker image to fail.

I think we need to append the user name at the end of the `NuGetScrarch` folder name. If I understand correctly, this should be `user`.
https://learn.microsoft.com/en-us/nuget/consume-packages/managing-the-global-packages-and-cache-folders

flobernd:
Maybe we can just set the `NUGET_SCRATCH` environment variable to `/tmp/NuGetScratch` to restore the old behavior. Inside the Docker container we only have a single active user anyways.

flobernd:
Looking good now! Thanks again for the PR and the updates. I'll merge on Monday and include the changes in the next patch release.

flobernd (approved):
LGTM! Thank you 🙂

landlord-matt:
@flobernd Both .NET 7 and .NET 8 focused on JSON serialisation performance. Did you notice any difference?

Will the patch release also target Elasticsearch 8.6? When do you expect the patch release?

flobernd:
@landlord-matt I don't think there will be any noticeable performance benefits as most of the optimizations are related to the codegen features which we don't use yet (they require code changes).

The patch will only target 8.12, but you should be able to update your 8.6 library to 8.12.1 when it's released. I think we had no major breaking changes since then.

A patch release will happen until end of March.

PR #8053, opened 2024-03-07
Use JsonPropertyName in Ranges

### LINKED ISSUE #8044: Range types do not serialize with the correct property names
**Elastic.Clients.Elasticsearch version**: 8.12.0

**Elasticsearch version**: 8.12.1

**.NET runtime version**: .NET 8

**Operating system version**: Windows 11

The [Range types](https://github.com/elastic/elasticsearch-net/blob/main/src/Elastic.Clients.Elasticsearch.Shared/Types/Ranges.cs) are using `[DataMember]` attributes to define the JSON property names but these aren't used by System.Net.Json.  The `JsonPropertyName` attribute should be used instead.  This causes an error when trying to index a document because the JSON looks like `{ "greaterThanOrEqualTo": 1, "lessThanOrEqualTo": 2 }` instead of  `{ "gte": 1, "lte": 2 }`

**Expected behavior**
Using a range type should properly serialize with the correct JSON property names

gpetrou:
Resolves #8044.

### DIFF
--- a/src/Elastic.Clients.Elasticsearch.Shared/Types/Ranges.cs
+++ b/src/Elastic.Clients.Elasticsearch.Shared/Types/Ranges.cs
@@ -3,7 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
-using System.Runtime.Serialization;
+using System.Text.Json.Serialization;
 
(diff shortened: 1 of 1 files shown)


flobernd (approved):
LGTM! Thank you 🙂

PR #8062, opened 2024-03-25
Implement proper deserialization of `EsqlQueryResponse`

flobernd:
Uses `CustomResponseBuilder` to property deserialize `EsqlQueryResponse`.

Additional changes:

- Mark all `RequestParameters` with the `partial` modifier.
- Regenerate client using the latest spec


src/Elastic.Clients.Elasticsearch.Shared/Api/Esql/EsqlQueryResponse.cs
```diff
@@ -0,0 +1,10 @@
+// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+namespace Elastic.Clients.Elasticsearch.Esql;
+
+public sealed partial class EsqlQueryResponse
+{
+	public byte[] Data { get; init; }
```
Mpdreamz, line 9: Are we gonna expose this more concretely in the future?

flobernd: @Mpdreamz The response is modelled like this in the specification:

```ts
export class Response {
  body: EsqlColumns
}

export type EsqlColumns = ArrayBuffer
```

Depending on the "format" query parameter, this buffer might contain JSON, CSV, ... (and even a binary Arrow dataframe in the future). Internally, we call this the "raw API".

The plan is to provide helper functions on top of the raw API for different usecases. Not all clients will implement the same high level helpers, but for .NET I'll definitely want to add at least the "Object API" (aka simple deserialization into source classes) and the "Arrow Dataframe API".

PR #8090, opened 2024-04-05
Improve serialization of "single-or-many" properties

### LINKED ISSUE #8056: Empty should collection is serialized improperly
**Elastic.Clients.Elasticsearch version**:  8.12.0

**Elasticsearch version**:  8.3.1

**.NET runtime version**: 8.0

**Operating system version**: MacOs Sonoma

**.Should** method with empty collection of actions generates incorrect json in the request to ES which leads to 400 response. NEST worked properly in the same case and sent an empty array

**Steps to reproduce**:
 1.
Do something like this
```            
var response = await this.Client8.SearchAsync<Contact>(
       sd => sd.Routing($"{contact.OrgId}").Index("contacts-t").Query(q => q.Bool(b => b.Should(Array.Empty<Action<QueryDescriptor<ContactProjection>>>()))));
```
 2.
The following body will be generated

```
# Request:
{
  "query": {
    "bool": {
      "should": 
    }
  }
}
```

 3. The response will fail with 400: Unexpected character ('}' (code 125)): expected a value

**Expected behavior**
The following request should be sent

```
# Request:
{
  "query": {
    "bool": {
      "should": []
    }
  }
}
```

flobernd:
Closes #8056

### DIFF
--- a/src/Elastic.Clients.Elasticsearch.Serverless/_Generated/Api/AsyncSearch/SubmitAsyncSearchRequest.g.cs
+++ b/src/Elastic.Clients.Elasticsearch.Serverless/_Generated/Api/AsyncSearch/SubmitAsyncSearchRequest.g.cs
@@ -1745,14 +1745,14 @@ public sealed partial class SubmitAsyncSearchRequestDescriptor<TDocument> : Requ
 		else if (KnnDescriptorActions is not null)
 		{
 			writer.WritePropertyName("knn");
-			if (KnnDescriptorActions.Length > 1)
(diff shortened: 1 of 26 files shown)


stevejgordon (commented):
The serialization changes look okay to me. It would probably be worth adding some test cases for these now we have a reasonable understanding of the nuances of the API and the expected formats of the data.

flobernd:
That's definitely a good idea. I'll add some tests in a following step after I updated the tests for 8.13 in general 🙂

PR #8129, opened 2024-04-16
Regenerate client using the latest spec

### LINKED ISSUE #8015: A call to client.Cluster.StatsAsync fails with JSON serialization error
**Elastic.Clients.Elasticsearch version**: 8.11.0

**Elasticsearch version**: 8.11.0

**.NET runtime version**: net7.0, net6.0

**Operating system version**: Windows 10

Run a plain request to get cluster stats:

`var stats = await client.Cluster.StatsAsync(new ClusterStatsRequest()).ConfigureAwait(false);`

It fails with the following error:
> The JSON value could not be converted to System.Int32. Path: $.indices.query_cache.total_count

**Notes**
This is only happening when the `total_count` property of the `query_cache` object of the target cluster exceeds the `int.MaxValue`. In my case, if I run the same query in the ES console, I get `2153022551` for the `.indices.query_cache.total_count`, which is obviously bigger than `int.MaxValue` (`2147483647`).
 The type of [the `TotalCount` property](https://github.com/elastic/elasticsearch-net/blob/main/src/Elastic.Clients.Elasticsearch/_Generated/Types/QueryCacheStats.g.cs#L78) is `int`, but should probably be `long`.

### LINKED ISSUE #8124: Unable to perform knn query [knn] unknown field [k
**Elastic.Clients.Elasticsearch version**:

**Elasticsearch version**:
8.13.0
**.NET runtime version**:
.NET 8
**Operating system version**:
Windows 11
I've created a knn query with the following syntax

```
        var knnOuterQuery = new QueryDescriptor<object>().Knn(knn => knn
            .Filter(query)
            .Field("vector")
            .NumCandidates(limit * 2)
            .QueryVector(coll));
```

Then I got an error:  ServerError: Type: x_content_parse_exception Reason: "[9:7] [knn] unknown field [k]"

Looking for generated query it contains a k value that it is not supported: this is the request where you can see a k : 0 that is not correct. If I remove that parameter and try to execute the query everything seems to work just fine.

```
{
  "from": 0,
  "query": {
    "knn": {
      "field": "vector",
      "filter": {
        "match_all": {}
      },
      "k": 0,
      "num_candidates": 10,
      "query_vector": [
        -0.0546875,
        -0.53125,
        0.453125,
        -0.484375
      ]
    }
  },
  "size": 5
}
```

Indeed the driver has that parameter in the request, but seems not to be understood by the server. I'm constructing the query from a QueryDescriptor<object>() object.

image

Am I doing something wrong or the driver indeed add a wrong "k" argument to the query?

flobernd:
Closes #8124
Closes #8015


alkampfergit:
Great I'm waiting for the nuget package to appear (even in prerelease). Many thanks.

PR #8185, opened 2024-05-10
fix: object initializer syntax for source filtering

EugeneGritsina:
Found the same syntax bug in at least 7.17 documentation also, but I'm not sure whether the syntax is correct for other versions. We use 6.0 in our project, so I provided a fix for it.


flobernd:
Thanks for the PR! Closing it since NEST is end of life.

Issue #8273, opened 2024-07-28
DefaultFieldNameInferrer (ToCamelCase) doesn't work like CamelCase of serializers


cwuethrich:
**Elastic.Clients.Elasticsearch version**: 8.14.6

**Elasticsearch version**: 8.14.3

**.NET runtime version**: all

**Operating system version**: all

**Description of the problem including expected versus actual behavior**:
The `ToCamelCase` extension as backup of the `DefaultFieldNameInferrer` doesn't work like the CamelCase namigs of `System.Text.Json` or `Newtonsoft.Json`. For example a property named `DBName` ends as `dBName` instead of `dbName`.
This behavior ends in exceptions if you use strict mapping.

**Steps to reproduce**:
```csharp
var settings = ***
var client = new ElasticsearchClient(settings);

string indexName = "test";
var createIndexResponse = await client.Indices.CreateAsync<MyDocument>(indexName, d => d
	.Mappings(m => m
		.Dynamic(DynamicMapping.Strict)
		.Properties(p => p
			.Keyword(d => d.DBName)
		)
	));
if (!createIndexResponse.IsSuccess())
{
	if (createIndexResponse.TryGetOriginalException(out Exception ex))
		throw ex;
}

var indexResponse = await client.IndexAsync(
	document: new MyDocument() { DBName = "database" },
	index: indexName,
	id: null);
if (!indexResponse.IsSuccess())
{
	if (indexResponse.TryGetOriginalException(out Exception ex))
		throw ex; // throws strict_dynamic_mapping_exception 
}
```

**Expected behavior**
The property name `DBName` should ends as `dbName` in property mapping.

flobernd:
Hi @cwuethrich, thanks for reporting! I'll have a look.

The current implementation of `ToCamelCase` is a very simple stub, if I remember correctly.

cwuethrich:
Hi @flobernd 
This got merged impressively fast. Thanks a lot!

flobernd:
No problem at all @cwuethrich ! Thanks again for the PR. A new release with the fix included will happen later this week 🙂

PR #8274, opened 2024-07-29
Adjust ToCamelCase Extension

cwuethrich:
Adjust `ToCamelCase` Extension of `string` to match the behavior of `System.Text.Json` and `Newtonsoft.Json`.
Bug: #8273

### DIFF
--- a/src/Elastic.Clients.Elasticsearch.Shared/Core/Extensions/StringExtensions.cs
+++ b/src/Elastic.Clients.Elasticsearch.Shared/Core/Extensions/StringExtensions.cs
@@ -18,10 +18,25 @@ internal static class StringExtensions
 		if (!char.IsUpper(s[0]))
 			return s;
 
-		var camelCase = char.ToLowerInvariant(s[0]).ToString();
-		if (s.Length > 1)
-			camelCase += s.Substring(1);
+		var chars = s.ToCharArray();
 
-		return camelCase;
+		for (var i = 0; i < chars.Length; i++)
+		{
+			var current = chars[i];
+
+			if (char.IsSeparator(current))
+				break;
+
+			if (0 < i && i + 1 < chars.Length)
+			{
+				var next = chars[i + 1];
+				if (!char.IsUpper(next) && !char.IsSeparator(next))
+					break;
+			}
+
+			chars[i] = char.ToLowerInvariant(current);
+		}
+
+		return new string(chars);
 	}
 }
--- a/tests/Tests/Core/Extensions/StringExtensionsTests.cs
+++ b/tests/Tests/Core/Extensions/StringExtensionsTests.cs
@@ -1,0 +1,19 @@
+// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+namespace Tests.Core.Extensions;
+
+public class StringExtensionsTests
+{
+	[U]
+	public void ToCamelCase()
+	{
+		"camelCase".ToCamelCase().Should().Be("camelCase");
+		"CamelCase".ToCamelCase().Should().Be("camelCase");
+		"CAMELCase".ToCamelCase().Should().Be("camelCase");
+		"CamelCASE".ToCamelCase().Should().Be("camelCASE");
+		"Camel Case".ToCamelCase().Should().Be("camel Case");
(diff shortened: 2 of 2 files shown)


src/Elastic.Clients.Elasticsearch.Shared/Core/Extensions/StringExtensions.cs
```diff
@@ -18,10 +18,25 @@ internal static string ToCamelCase(this string s)
...
-		if (s.Length > 1)
-			camelCase += s.Substring(1);
+		var chars = s.ToCharArray();
 
-		return camelCase;
+		for (var i = 0; i < chars.Length; i++)
+		{
+			var current = chars[i];
+
+			if (char.IsSeparator(current))
```
flobernd, line 27:
```suggestion
			if (!char.IsLetter(current))
```

src/Elastic.Clients.Elasticsearch.Shared/Core/Extensions/StringExtensions.cs
```diff
@@ -18,10 +18,25 @@ internal static string ToCamelCase(this string s)
...
+				break;
+
+			if (0 < i && i + 1 < chars.Length)
+			{
+				var next = chars[i + 1];
+				if (!char.IsUpper(next) && !char.IsSeparator(next))
```
flobernd, line 33:
```suggestion
				if (char.IsLetter(next) && !char.IsUpper(next))
```

src/Elastic.Clients.Elasticsearch.Shared/Core/Extensions/StringExtensions.cs
```diff
@@ -18,10 +18,25 @@ internal static string ToCamelCase(this string s)
 		if (!char.IsUpper(s[0]))
 			return s;
 
-		var camelCase = char.ToLowerInvariant(s[0]).ToString();
-		if (s.Length > 1)
-			camelCase += s.Substring(1);
+		var chars = s.ToCharArray();
```
flobernd, line 21:
```suggestion
#if NET6_0_OR_GREATER
		Span<char> chars = stackalloc char[s.Length];
		s.CopyTo(chars);
#else
		var chars = s.ToCharArray();
#endif
```

flobernd: This requires an additional:

```csharp
#if NET6_0_OR_GREATER
using System;
#endif
```

flobernd (commented):
Thanks for the PR 🙂 

I suggested a few adjustments. 

One suggestion is a minor performance optimization using a stack allocated `Span` instead of allocating a new array on the heap.

The other suggestion is a functional one. I think we want to handle e.g. digits the same way as separators so that for instance `SHA1` is converted to `sha1` instead of `shA1`. To save some checks, I reversed the logic to check for `char.IsLetter()` (or `!char.IsLetter()`) instead of explicitly excluding separators, digits, etc.

flobernd:
Actually, let's just copy the `System.Text.Json` implementation (MIT license):

```csharp
    internal static string ToCamelCase3(this string name)
    {
        if (string.IsNullOrEmpty(name) || !char.IsUpper(name[0]))
        {
            return name;
        }

#if NET
        return string.Create(name.Length, name, (chars, name) =>
        {
            name.CopyTo(chars);
            FixCasing(chars);
        });
#else
            char[] chars = name.ToCharArray();
            FixCasing(chars);
            return new string(chars);
#endif
    }

    private static void FixCasing(Span<char> chars)
    {
        for (var i = 0; i < chars.Length; i++)
        {
            if (i == 1 && !char.IsUpper(chars[i]))
            {
                break;
            }

            var hasNext = (i + 1 < chars.Length);

            // Stop when next char is already lowercase.
            if (i > 0 && hasNext && !char.IsUpper(chars[i + 1]))
            {
                // If the next char is a space, lowercase current char before exiting.
                if (chars[i + 1] == ' ')
                {
                    chars[i] = char.ToLowerInvariant(chars[i]);
                }

                break;
            }

            chars[i] = char.ToLowerInvariant(chars[i]);
        }
    }
```

cwuethrich:
Wasn't sure about the license. But if we are allowed to use it, it would be the best. Then the properties for sure gonna match. Sad they didn't make theirs classes public.

flobernd:
@cwuethrich I can open a new PR, if you don't mind 🙂 Thanks for creating this one!

PR #8290, opened 2024-08-08
ignore obsolete warning in tests

svalbuena:


flobernd:
Thank you for the PR. I think adding this directive to `.editorconfig` is okay for now. A lot of tests have not been adjusted after the breaking 8.13 release yet. This is definitely something I would like to fix in a timely manner.

PR #8376, opened 2024-10-09
Move away from shared project

Mpdreamz:
This is an alternative approach to share code between the two clients. This is a tad more inline with the `dotnet` CLI as we no longer depend on legacy projects. 

The impetus for this change was to improve cold compilation times. 

My local jetBrains Rider kept locking up trying to do the initial compile. 

- Moves us over to `<ArtifactsPath>` (set to `$root\.artifacts` like other .NET projects)
- Removes a leftover `<OutputPath>` for certain projects (this was tripping up deps.json resolving HARD).  
- No longer use nuget package locking
  - nuget resolves to lowest anyway and we never float dependencies
  - no other .NET repository at Elastic does this.


flobernd:
This looks great ❤️ Thanks a lot for creating this PR.

I will need some time to review in detail, but already have one comment regarding the locked restore:

Would keeping the package lock have any other noticable drawback? I usually like that approach as it e.g. allows for easy package caching in CI (based on a hash of all `package.lock.json` files). This speeds up build times on the slow GitHub workers quite a bit.

To mitigate some of the "developer experience issues" like having to use `--force-evaluate`, my usual approach is to guard the condition like this:

```xml
<RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
<RestoreLockedMode Condition="'$(ContinuousIntegrationBuild)' == 'true'" />
```

The CI passes that argument whenever the project must be built or packaged:

https://github.com/elastic/elasticsearch-net/blob/5cde9691b5a4613068ae1c97e8c6268575f77d57/.github/workflows/release.yml#L86

What do you think?

Mpdreamz:
We simply cache by `[cf]proj` files. Slightly higher cache busting rate but in practice not a big issue :)

https://github.com/elastic/apm-agent-dotnet/blob/3ffe9b4f24f2e3cfdc5e82c1a42e346d64d36fb9/.github/workflows/bootstrap/action.yml#L39

flobernd:
No super strong opinion here, but given that the `package.lock.json` approach has no real drawbacks (AFAIK), I think it's the more robust approach. These files always represent the final package graph.

Hashing just `[cf]proj` won't capture changes in includes for these files as well as `Directory.Packages.props`, `Directory.Build.props`, `Directory.Build.targets`, etc.

`Directory.Build.props` for example is currently used to include the `DotNet.ReproducibleBuilds` package.

PR #8405, opened 2024-11-14
Update `Elastic.Transport` to 0.5.2

flobernd:
Update `Elastic.Transport` to 0.5.2


stevejgordon (approved):
LGTM - Not as bad as I feared that would be!

Mpdreamz (approved):
🥳

PR #8434, opened 2024-12-20
Update getting-started.asciidoc

predogma:
The connection should be using the cloud id and API key, not the cloud endpoint.

The image unfortunately will need to be updated to as it obfuscates the cloud id and highlights the endpoint copy.

Sample Screenshot
Screenshot 2024-12-20 at 9 51 28 AM

This would be consistent with our docs here
https://www.elastic.co/guide/en/elasticsearch/client/net-api/current/connecting.html#cloud-deployment

Which states always use the cloud id, for our Elastic Cloud Hosted deployments to connect.

### DIFF
--- a/docs/getting-started.asciidoc
+++ b/docs/getting-started.asciidoc
@@ -27,16 +27,17 @@ Refer to the <<installation>> page to learn more.
 === Connecting
 
 You can connect to the Elastic Cloud using an API key and the Elasticsearch 
-endpoint. 
+cloud id. 
 
 [source,net]
 ----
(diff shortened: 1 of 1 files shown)


marciw:
Thanks! I pushed a commit that fixes capitalization and adds the corrected image. I also borrowed the API key phrasing from the Python client docs for now.

PS: ideally this would target `main` rather than `8.17`... but because I'm planning other improvements soon, I can just bring these changes along at that point. But FYI for future PRs :)

marciw (approved):
LGTM with edits

cc @flobernd

Issue #8339, opened 2024-09-05
8.15.6 - General API Feedback from NEST.


niemyjski:
**Elastic.Clients.Elasticsearch version**: 8.15.6

**Elasticsearch version**: 8.15.1

**.NET runtime version**: 8.x

**Operating system version**: Any

**Description of the problem including expected versus actual behavior**:

While upgrading I've noticed a few things that really made the conversion harder than it had to be and I feel like the feedback below would have made things easier. I get that the purpose of the rewrite was to make maintenance easier, but on the other end is consumers/customers that have large investments in existing api's / elastic.

1. Kept the existing interfaces, they were not hurting anything especially around code gen.
2. leverage interfaces more for very common fields like IProperty.Fields (tons of properties have fields, other common properties) having something like IHaveField or anything... This change with concrete types forces us to check tons of concrete types which makes maintaining it even harder as new types are added. No common interface for the tons of aggregation types seems like a miss?
3. There is 0 point to have such an internalized api (worst case make the setters internal / private). This has made things much harder than it should be especially around aggregations / sorting. The property names chosen for some like VariantName kinda adds to the confusion.
4. 1 way implicit conversions for aggregations requires lots of boxing, type casting, and harder to go back the other way, especially with internalized. Also this method might make your serializer implementation slightly easier to generate, but harder to work with. It also means the aggregations don't actually represent what is supported (e.g., meta, sub aggs).
5. Leverage debugger displays / record types for better dx.
6. I've logged other issues below which seem like bigger problem areas to solve.

**Expected behavior**

Easier to use api's and less internalized things leading to a better experience.

Reference: https://github.com/FoundatioFx/Foundatio.Parsers/pull/84

Also apologies if the labels are incorrect for or this issue:

- https://github.com/elastic/elasticsearch-net/issues/8334
- https://github.com/elastic/elasticsearch-net/issues/8335
- https://github.com/elastic/elasticsearch-net/issues/8336
- https://github.com/elastic/elasticsearch-net/issues/8337
- https://github.com/elastic/elasticsearch-net/issues/8338

Thank you.

flobernd:
Hi @niemyjski, thank you for the detailed feedback. 

Quite a few of these design decisions were made before I took over the maintainer role for the .NET clients. It probably makes sense to discuss some of these points during the call together with Martijn.

1. If I understood correctly, you are mainly missing a way to implement custom aggregation variants (custom name, custom aggregation type) with the new client. Is that correct?
[...]

PR #8448, opened 2025-01-21
Improve serialization

flobernd:
Changes:

- [x] Improve custom converter performance.
  - [x] Replace nested `JsonSerializer.{Serialize|Deserialize}` calls with direct converter access (`JsonSerializerOptions.GetConverter` -> `converter.{Read|Write}`) to prevent frequent creation of top-level JSON reader stacks.
  - [x] Prefer direct `ValueSpan` access instead of allocating intermediate `string` values (e.g. when comparing property names or parsing/formatting values).
  - [x] Use pre-encoded `JsonEncodedText` constants for compile-time constant literals (e.g. property names).
  - [x] Use `Utf8Parser`/`Utf8Formatter` for primitive value conversions. 
  - [x] Envelope-stype marker types like e.g. `SourceMarker<T>` should only (de-)serialize the wrapped type to prevent unnecessary allocations.
- [ ] Implement round-trip support for all types (limited to all JSON serializable types).
  - [ ] Requests
    - [x] Regular requests
    - [x] Requests with required path-parameters (deserialization works, but we need a generic way to set the path-parameters)
    - [ ] NDJSON requests (`IStreamSerializable`)
  - [ ] Responses
    - [x] Regular responses
    - [x] Dictionary responses
    - [ ] NDJSON responses (`IStreamSerializable`)
  - [x] Regular classes
  - [x] Unions
  - [ ] Complex Unions
  - [x] Enums
  - [x] Internal variants
  - [x] External variants
  - [x] Untagged variants
  - [x] Container variants
  - [ ] `SortOptions` special case
- [ ] Change descriptors to wrap the corresponding object representations in order to re-use serialization code.

Breaking user facing changes:

1. :boom: Removed `Field.Format` property and corresponding constructor and inferrer overloads.
   This property was not used anymore for quite some time (got replaced by the `FieldAndFormat` type).
2. :boom: `Field`/`Fields` static factory methods and conversion operators no longer return nullable references but throw exceptions instead (`Field`) if the input `string`/`Expression`/`PropertyInfo` argument is `null`.
   This especially makes implicit conversions to `Field` more easily usable without having to apply the null-forgiveness operator (`!`) all the time.
3. :boom: Removed `FieldValue.IsLazyDocument`, `FieldValue.IsComposite` and the corresponding members in the `FieldValue.ValueKind` enum. These values were not used anymore for quite some time.

Non-breaking user facing changes:

1. All requests do now expose read/write access to path properties (like `index`, `id`, ...) which previously could only be initialized from the constructor.

Later/bonus:

- [ ] Do not generate custom converters when property attributes are sufficient.
- [ ] Serialization AOT support.
- [x] Get rid of `ISelfTwoWaySerializable` (used for source-body/array/dictionary requests/responses).
- [ ] Get rid of `ISelfSerializable` (used for descriptors).
- [ ] Get rid of `ElasticsearchClient.SettingsTable` (used for descriptors and in some other places). Replace with `ContextProvider` access.
- [ ] Validate required properties and nullability.
- [ ] Replace globally registered `IsADictionaryConverterFactory` with case-by-case converters.

Related to:

- #8339
  All variants now have either a public `string Type` (internal, external, untyped) property or public `VariantType/Variant` properties (container). For non-exhaustive container variants like e.g. `Aggregation`, these properties are settable.


flobernd:
Closed in favor of #8485

PR #8450, opened 2025-01-29
Update CrudExamplesTests.cs

RobinSoenen:
Small change in the documentation:
 
var response = await client.IndexAsync(tweet, "my-tweet-index"); // <2>

To 
var response = await client.IndexAsync(tweet, (IndexName)"my-tweet-index"); // <2>

in order to avoid an ambiguous call:

error CS0121: The call is ambiguous between the following methods or properties: 'ElasticsearchClient.IndexAsync(TDocument, IndexName, CancellationToken)' and 'ElasticsearchClient.IndexAsync(TDocument, Id?, CancellationToken)'


RobinSoenen:
Contributor agreement signed in the meantime

PR #8466, opened 2025-03-04
Fix #8340: Prevent null reference exception in SearchResponse<T>

### LINKED ISSUE #8340: SearchResponse.Documents throws null reference exception on successful call
**Elastic.Clients.Elasticsearch version**:
8.15.5

SearchResponse<T>.Documents throws a null ref exception because HitsMetadata is null (and/or possibly HitsMetadata.Hits).

Stacktrace:
at Elastic.Clients.Elasticsearch.SearchResponse`1.get_Documents() in /_/src/Elastic.Clients.Elasticsearch.Shared/Api/SearchResponse.cs:line 21

**Expected behavior**
Should not throw an exception. HitsMetadata/Hits being null should be handled gracefully.
The Documents property and also Hits property should handle HitsMetadata being null and also declare correct nullability.

sadeghpour:
- Added null checks for `HitsMetadata` in `SearchResponse<T>.Documents`
- Ensured `HitsMetadata` and `Hits` are properly validated before accessing them
- Prevented potential crashes when the response contains no hits

Closes #8340

### DIFF
--- a/src/Elastic.Clients.Elasticsearch/_Shared/Api/SearchResponse.cs
+++ b/src/Elastic.Clients.Elasticsearch/_Shared/Api/SearchResponse.cs
@@ -2,6 +2,7 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.Json.Serialization;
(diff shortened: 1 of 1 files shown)


sadeghpour:
Hi team, I have signed the CLA, but the check still seems to be failing. Could you please verify? Let me know if I need to take any further action. Thanks!

pquentin:
Hello, and thank you for your contribution! I have no idea about the code itself. I can confirm that you signed the CLA check correctly, but the issue is that your commit does not use an email address. See https://github.com/elastic/elasticsearch-net/commit/4a91b26f1095bd82681c825b441960f2dba0df07.patch.

flobernd:
Thanks for the PR @sadeghpour !

Besides this fix, I would like to investigate the case a little bit more in depth. Would it be possible for you to send me the JSON request and response that triggers this behavior?

In our specification, both, `ResponseBody<TDocument>.hits` (this is `HitsMetadata` in the .NET client) and `HitsMetadata<T>.hits` are not marked as optional. I wonder if this should be changed cc @pquentin @l-trotta

l-trotta:
@flobernd never found any evidence that `hits` could be null, is there a special case where it happens?

sadeghpour:
@l-trotta You are right! After further investigation following @flobernd  previous comment, I also couldn't find any evidence that hits could be null. It seems there's no special case where this happens.

PR #8505, opened 2025-04-28
9.0.1 fixes

### LINKED ISSUE #8502: 9.0 - GetAliasResponse is initialized with Aliases set to null instead of empty
**Elastic.Clients.Elasticsearch version:** 9.0.0

**Elasticsearch version:** 9.0.0

**.NET runtime version:** net9.0

**Operating system version:** Windows for client, docker for elasticsearch

**Description of the problem including expected versus actual behavior:**
GetAliasResponse is now initialized with Aliases being null instead of empty dictionary. This causes IsValidResponse and the depending DebugInformation to fail with System.NullReferenceException when no matching alias exists.
Expected result is for IsValidResponse not to throw, either by dictionary being empty, or by checking for nulls.

### LINKED ISSUE #8503: 9.0.0 - [inner_hits] fields doesn't support values of type: VALUE_STRING
**Elastic.Clients.Elasticsearch version**: 9.0.0

**Elasticsearch version**: 9.0.0

**.NET runtime version**: 9.0.4

**Operating system version**: macOS

Incorrect query serialization for `inner_hits.fields`.
According to spec it can be just string (not an array): https://github.com/elastic/elasticsearch-specification/blob/main/specification/_types/common.ts#L137

but elastic reply with error `[inner_hits] fields doesn't support values of type: VALUE_STRING` when `fields` serialized as `string` instead of `arrray`.

**Steps to reproduce**:

Execute query with `fields` serialized as `string`
```json
"inner_hits": {
   "fields": "passages.text",
   "name": "elser-chunk",
   "size": 1,
   "_source": false
},
```

**Expected behavior**

Server should not return an error in this case, or serializer should always serialize `fields` as array. Following example work as expected:

```json
"inner_hits": {
   "fields": ["passages.text"],
   "name": "elser-chunk",
   "size": 1,
   "_source": false
},
```

Image

Image

// cc @flobernd please take a look, it looks like a significant regression.

flobernd:
Closes #8502 
Closes #8503

### DIFF
--- a/src/Elastic.Clients.Elasticsearch/_Generated/Api/MachineLearning/GetDataFrameAnalyticsStatsRequest.g.cs
+++ b/src/Elastic.Clients.Elasticsearch/_Generated/Api/MachineLearning/GetDataFrameAnalyticsStatsRequest.g.cs
@@ -109,7 +109,7 @@ internal sealed partial class GetDataFrameAnalyticsStatsRequestConverter : Syste
 
 /// <summary>
 /// <para>
-/// Get data frame analytics jobs usage info.
+/// Get data frame analytics job stats.
 /// </para>
 /// </summary>
 [System.Text.Json.Serialization.JsonConverter(typeof(Elastic.Clients.Elasticsearch.MachineLearning.GetDataFrameAnalyticsStatsRequestConverter))]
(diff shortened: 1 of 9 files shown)


sergey-tihon:
LGTM! Thank you 😆

sergey-tihon:
Thank you, I just wanted to check if I can trick the bot with LGTM.

PR #8628, opened 2025-07-17
Fix missing quote in docs example

florent-leborgne:
Fixes a missing quote reported in https://github.com/elastic/docs-content/issues/2164

Closes: https://github.com/elastic/docs-content/issues/2164


pquentin (approved):
Thanks! LGTM.

PR #8661, opened 2025-08-12
feature/aot continued

Mpdreamz:
- **Fix remaining AOT compiler errors**
- **More AOT annotations**
- **Id and Routing converters need to be public in order to be used by static serializer contexts outside of the client.**
- **Add Serializer context**
- **Add overload taking multiple IJsonTypeResolvers and ensure we use JsonTypeInfoResolver.Combine**


src/Elastic.Clients.Elasticsearch/_Shared/Serialization/DefaultSourceSerializer.cs
```diff
@@ -94,10 +100,13 @@ private static IReadOnlyCollection<JsonConverter> CreateDefaultBuiltInConverters
 
 	[UnconditionalSuppressMessage("AOT", "IL3050:Calling members annotated with 'RequiresDynamicCodeAttribute'", Justification = "Always using explicit TypeInfoResolver")]
 	[UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute'", Justification = "Always using explicit TypeInfoResolver")]
-	private static void MutateOptions(JsonSerializerOptions options, IJsonTypeInfoResolver? typeInfoResolver, Action<JsonSerializerOptions>? configureOptions)
+	private static void MutateOptions(JsonSerializerOptions options, IJsonTypeInfoResolver[]? typeInfoResolvers, Action<JsonSerializerOptions>? configureOptions)
```
flobernd, line 103: Do we really need the multi-resolver overload?

I was under the impression that we can chain multiple `JsonTypeInfoResolver.Combine()` calls (given that the `System.Text.Json` API as well only accepts a single `IJsonTypeInfoResolver` in all places). This should give the user already all the required flexibility.

Mpdreamz: Talked offline, reverted this!

flobernd (approved):
Thank you 🙂

PR #8701, opened 2025-09-11
Default DOTNET_VERSION=9.0.305 for unified-release workflow

MindaugasButkus:
This should resolve the unified-release workflow failures.

The errors were most likely caused by the discard operators in `JsonUnionSelector.T1` and `JsonUnionSelector.T2`.

`DOTNET_VERSION=9.0.100 ./.ci/make.sh assemble 9.0.0-SNAPSHOT` fails.
`DOTNET_VERSION=9.0.305 ./.ci/make.sh assemble 9.0.0-SNAPSHOT` succeeds.

Went through release notes of different versions of .NET 9. Couldn't find any mentions of this issue. But hey - it works.

### DIFF
--- a/.ci/make.sh
+++ b/.ci/make.sh
@@ -42,7 +42,7 @@ OUTPUT_DIR="$repo/${output_folder}"
 REPO_BINDING="${OUTPUT_DIR}:/sln/${output_folder}"
 mkdir -p "$OUTPUT_DIR"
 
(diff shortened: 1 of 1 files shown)


flobernd:
Hi @MindaugasButkus, thanks again for creating a PR!

I tracked down the issue to the discard operators as well, but ran out of time to investigate the root cause. Good thing you found a solution.

PR #8792, opened 2025-12-08
Use UTF-8 literals in `JsonEncodedText.Encode`.

teo-tsirpanis:
Will save a conversion from UTF-16. The generator will have to be updated accordingly.


flobernd:
Good catch, thank you!

PR #8835, opened 2026-03-04
Fix mutable Metrics.All singleton and constructor bugs

### LINKED ISSUE #7732: Singleton `Metrics.All` is mutable
The idea of any singleton to be immutable if it serves an idea of a constant value, but unfortunately, it doesn't apply to `Metric` exposing values as a `HashSet<string>`:

https://github.com/elastic/elasticsearch-net/blob/9e58d38688dfb1f15ba6868379c769adf488225b/src/Elastic.Clients.Elasticsearch/Core/Infer/Metric/Metrics.cs#L22

A potential solution is to use an immutable collection or `IReadOnlySet` interface which limits mutability, but that can be allowed through methods added to `Metrics`.

amitkanfer:
## Summary

Fixes #7732

- **Immutability**: Changed the backing type of `Metrics.Values` from `HashSet<string>` to `string[]`. Arrays are fixed-size and cannot have elements added or removed, making `Metrics.All` and all other `Metrics` instances genuinely immutable.
- **Constructor bug**: The null/empty guard in both constructors assigned `Values = EmptyMetrics` but then fell through to overwrite it unconditionally. Added `return` after the guard so it takes effect.
- **Shared mutable field removed**: Removed the `private static readonly HashSet<string> EmptyMetrics` shared field; replaced with array literals `[]`.

## Test plan

- [ ] Verify `Metrics.All.ToString()` returns `"_all"` after this change
- [ ] Verify `new Metrics(string.Empty).ToString()` returns `""` (was broken before due to constructor bug)
- [ ] Verify `new Metrics((string)null)` no longer silently produces a set containing `null`
- [ ] Existing unit tests pass

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/Elastic.Clients.Elasticsearch/_Shared/Core/Infer/Metric/Metrics.cs
+++ b/src/Elastic.Clients.Elasticsearch/_Shared/Core/Infer/Metric/Metrics.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 
 using Elastic.Transport;
 
@@ -20,8 +21,6 @@ public sealed class Metrics :
 	, IParsable<Metrics>
 #endif
 {
-	private static readonly HashSet<string> EmptyMetrics = new();
-
 	/// <summary>
 	/// An instance of <see cref="Metrics"/> representing all statistics.
 	/// </summary>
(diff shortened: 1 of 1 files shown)


flobernd:
The `Values` property is `private` and no mutation happens inside the `Metrics` class. The original issue is invalid.

PR #8836, opened 2026-03-04
Fix Query operators to reflect nullable parameters

### LINKED ISSUE #8807: Query operator handles null but no nullable in method siguature resulting in warnings
Any reason why these are not nullable types? The operator methods (.And() and .Or()) seems to be handling null args so it appears that nulls are allowed?

https://github.com/elastic/elasticsearch-net/blob/de0517da98ac751d15b8ea02fc901c4cbca0b0f0/src/Elastic.Clients.Elasticsearch/_Shared/Types/QueryDsl/Query.cs#L17C33-L17C39

Without this using Nullable<Query> results in a compile time warning unless pragma is used
`# pragma warning disable CS8604`

amitkanfer:
## Summary

Fixes #8807

The `&`, `|`, `!`, and `+` operators on `Query` silently handle `null` arguments in their implementations, but the method signatures declared non-nullable parameters. This caused spurious `CS8604` (possible null reference argument) warnings when combining `Query?` variables with these operators under nullable reference types.

Changes:
- `operator &` and `operator |`: parameters changed from `Query` to `Query?`
- `And()` and `Or()` internal helpers: parameters changed to `Query?`, non-null assertions added where the logic guarantees non-null after null checks
- `operator !` and `operator +`: return types changed to `Query?` to reflect they return `null` when the input is `null`

## Test plan

- [ ] Project compiles without `CS8604` warnings when using `Query?` variables with `&` and `|`
- [ ] `null & query` and `query & null` return the non-null operand (existing behavior preserved)
- [ ] `!((Query?)null)` returns `null` (existing behavior preserved)
- [ ] Existing tests pass

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/Elastic.Clients.Elasticsearch/_Shared/Types/QueryDsl/Query.cs
+++ b/src/Elastic.Clients.Elasticsearch/_Shared/Types/QueryDsl/Query.cs
@@ -14,10 +14,10 @@ public partial class Query
 
 	public static bool operator true(Query _) => false;
 
-	public static Query operator &(Query leftContainer, Query rightContainer) =>
+	public static Query operator &(Query? leftContainer, Query? rightContainer) =>
 		And(leftContainer, rightContainer);
 
-	internal static Query And(Query leftContainer, Query rightContainer)
+	internal static Query And(Query? leftContainer, Query? rightContainer)
 	{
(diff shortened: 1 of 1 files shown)


flobernd (approved):
LGTM! Thank you.

PR #8901, opened 2026-04-29
[DOCS] Connect via endpoint URL (fix Cloud ID mismatch)

### LINKED ISSUE #2431: [Website]: elasticsearch/clients/dotnet/getting-started
### Type of issue

Missing information

### What documentation page is affected

https://www.elastic.co/docs/reference/elasticsearch/clients/dotnet/getting-started

### What happened?

What is `CLOUD_ID` in https://www.elastic.co/docs/reference/elasticsearch/clients/dotnet/getting-started#_connecting?

### Additional info

_No response_

charlotte-hoblik:
Same direction as https://github.com/elastic/docs-content/pull/5414: prefer the Elasticsearch endpoint URL in connection examples so steps match the Cloud UI, curl, and other client docs.

This addresses https://github.com/elastic/docs-content/issues/2431 :
- text pointed at the endpoint while the sample still used `CLOUD_ID`. 
- Connecting page now treats endpoint + API key as default and documents Cloud ID only as an optional path for hosted Elastic Cloud.


docs/reference/connecting.md
```diff
@@ -17,15 +17,27 @@ Client applications should create a single instance of `ElasticsearchClient` tha
...
 
-The following snippet demonstrates how to create a client instance that connects to an {{es}} deployment in the cloud.
+var client = new ElasticsearchClient(settings);
+```
+
+1. Replace `<ELASTICSEARCH_ENDPOINT>` with your cluster’s HTTPS URL and `<API_KEY>` with the API key configured for your application.
+
+### Using a Cloud ID (Elastic Cloud hosted only) [cloud-id]
+
+On hosted Elastic Cloud deployments you can alternatively pass the **Cloud ID** string from the deployment overview together with credentials. The client resolves the Cloud ID to cluster nodes and applies defaults suited to Elastic Cloud (including HTTPS and HTTP compression).
```
swallez: `to cluster nodes` should be `to the deployment URL` as all nodes are hidden behind Elastic Cloud's proxy.

We can also remove HTTP since it's HTTPS only. I would actually remove the entire `(including ... compression)` snippet as the goal of defaults is to simplify the user's life and not have them care about it!

swallez (commented):
Left a comment.

And a question that is probably more for @flobernd: there no mention of node sniffing in the node pool section. Should we add it, or is it documented elsewhere (and then should be linked here)?

docs/reference/connecting.md
flobernd: > The client resolves the Cloud ID to cluster nodes and applies defaults suited to Elastic Cloud (including HTTPS and HTTP compression)

This is actually not true 😅 These optimizations only get applied when using the Cloud ID (`CloudNodePool` in the client). 

The above `ElasticsearchClientSettings` constructor overload using an url will resolve to the `SingleNodePool` - which does not apply any cloud specific parameters.

The urls are not a reliable inidicator (since custom domains are often used) and therefore not used to change the behavior. The .NET client requires the user to express explicit intent:

```csharp
var pool = new CloudNodePool(new Uri("<ELASTICSEARCH_ENDPOINT>"), new ApiKey("<API_KEY>"));
var settings = new ElasticsearchClientSettings(pool);
```

flobernd: The node sniffing implementations live in `Elastic.Transport` (elastic/elastic-transport-net) and I'm afraid this is only documented in code, since we still did not release a documentation for the transport library (which technicall is still in alpha (`0.x` version)).

The user has to express explicit intent here as well by using either `SniffingNodePool` or `StickySniffingNodePool`.

```
  ┌───────────────────────────────────────────┬─────┬───────────────────────────────────────────────────────────┬────────────┬───────────────────────────────────────────┐
  │ Pool                                      │ Pin │ Sniff (reseed)                                            │ MaxRetries │ Selection                                 │
  │                                           │ gs  │                                                           │            │                                           │
  ├───────────────────────────────────────────┼─────┼───────────────────────────────────────────────────────────┼────────────┼───────────────────────────────────────────┤
  │ SingleNodePool                            │ off │ off (Reseed no-op)                                        │ 0          │ the single node                           │
  │ CloudNodePool (extends SingleNodePool)    │ off │ off (Reseed no-op)                                        │ 0          │ the cloud endpoint                        │
  │ StaticNodePool                            │ on  │ off (Reseed no-op)                                        │ nodes.Coun │ round-robin via global cursor (with       │
  │                                           │     │                                                           │ t          │ dead/alive tracking)                      │
  │                                           │     │                                                           │ - 1        │                                           │
  │ SniffingNodePool (extends StaticNodePool) │ on  │ on (writes new node list under a ReaderWriterLockSlim)    │ nodes.Coun │ round-robin via global cursor over the    │
  │                                           │     │                                                           │ t          │ sniffed list                              │
  │                                           │     │                                                           │ - 1        │                                           │
  │ StickyNodePool (extends StaticNodePool,   │ on  │ off (Reseed overridden as no-op)                          │ nodes.Coun │ always starts at cursor 0, sticks to the  │
  │ sealed)                                   │     │                                                           │ t          │ first alive node                          │
  │                                           │     │                                                           │ - 1        │                                           │
  │ StickySniffingNodePool (extends           │ on  │ on (with optional Func<Node, float> nodeScorer; default   │ nodes.Coun │ always starts at cursor 0, sticks to the  │
  │ SniffingNodePool, sealed)                 │     │ scorer returns 0f)                                        │ t          │ first alive node                          │
  │                                           │     │                                                           │ - 1        │                                           │
  └───────────────────────────────────────────┴─────┴───────────────────────────────────────────────────────────┴────────────┴───────────────────────────────────────────┘
```

flobernd:
@charlotte-hoblik Hey 👋 are you going to iterate on this? Feel free to contact me, if anything needs clarification.

charlotte-hoblik:
Thanks @swallez and @flobernd for the review and sorry for the slow follow-up on my side.

@flobernd especially: I really appreciate how much time you put into this. The pool behavior explanation and the sniffing context were a huge help. That kind of research makes the docs and the author better. Thank you.

I’ve updated `connecting.md` to reflect your feedback without turning the page into a transport deep-dive:

- Cloud ID: copy now talks about the deployment endpoint behind Elastic Cloud’s proxy 
- Main cloud example: still endpoint URL + API key as the default path, aligned with the console and other clients.
- Sniffing: a short pointer to Elastic.Transport / the [elastic-transport-net](https://github.com/elastic/elastic-transport-net) repo so we answer @swallez’s question without dropping the whole pool matrix in.

If you’d rather keep CloudNodePool called out explicitly for advanced readers, I’m happy to add a single sentence.

Thanks again—please take another look when you can.

docs/reference/connecting.md
```diff
@@ -17,15 +17,27 @@ Client applications should create a single instance of `ElasticsearchClient` tha
...
 
-The following snippet demonstrates how to create a client instance that connects to an {{es}} deployment in the cloud.
+var client = new ElasticsearchClient(settings);
+```
+
+1. Replace `<ELASTICSEARCH_ENDPOINT>` with your cluster’s HTTPS URL and `<API_KEY>` with the API key configured for your application.
+
+### Using a Cloud ID (Elastic Cloud hosted only) [cloud-id]
+
+On hosted Elastic Cloud deployments you can alternatively pass the **Cloud ID** string from the deployment overview together with credentials. The client turns that into your deployment’s {{es}} endpoint—the URL exposed through Elastic Cloud’s proxy, not a list of separate machines you connect to one by one.
```
flobernd:
```suggestion
On hosted Elastic Cloud deployments you can alternatively pass the **Cloud ID** string from the deployment overview together with credentials. The client turns that into your deployment’s {{es}} endpoint.
```

The "list of separate machines" part is an implementation detail that is probably not relevant to the user.

flobernd (approved):
Thank you! Just one /nit 🙂

Issue #7664, opened 2023-04-28
[PERF] `IdResolver` and `RoutingResolver` closure allocations


stevejgordon:
The `IdResolver` and `RoutingResolver` types cache lookup functions but, in doing so, cause closure classes to be allocated per document whenever we infer the Id or Route for an object instance. This is most evident during bulk operations. For example, here are the excess allocations caused by a BulkAll request that indexes 100k documents.

![image](https://user-images.githubusercontent.com/3669103/235137009-a5afec07-b33c-4c25-998c-63e6699fc0de.png)

We should revisit the code in these resolvers to review if we can:

a) avoid capturing local variables and therefore allocating closures
b) optimise further (when the platform supports it) using IL emit etc.

PR #8973, opened 2026-08-14
Restore delegate caching in IdResolver

fresh55:
IdResolver only caches the getter delegate when resolution fails. On success it
returns without caching, so the delegate is rebuilt on every call.

Dropped in 4117107d17 during the aot rework. RoutingResolver from the same
commit still caches both paths.

100k resolves of the same type: 64 ms / 55.2 MB before, 3.5 ms / 40 bytes after.

Partially addresses #7664. That issue also mentions closure allocations in
RoutingResolver which I left alone here, happy to do that in a separate pr.


flobernd:
Hi @fresh55 , thanks for the PR!

fresh55:
thanks for merging

is the routingresolver half of #7664 still worth a pr or has that moved on. happy to open one if it is useful

PR #8979, opened 2026-08-14
Avoid closure allocations in RoutingResolver

### LINKED ISSUE #7664: [PERF] `IdResolver` and `RoutingResolver` closure allocations

fresh55:
`Resolve()` passes two lambdas to `JoinField.Match()` that capture `instance` and `this`. The closure is allocated in the method prologue, so every call pays for it, even the ones that return before `Match()` is reached.

Swapped `Match()` for a check on the internal tag, both types are in the same assembly so nothing public changes.

100k resolves of the same type:

| path | before | after |
| --- | --- | --- |
| join field child | 8.9 ms / 160 B/op | 3.4 ms / 0 B/op |
| route property | 5.1 ms / 32 B/op | 3.6 ms / 0 B/op |
| no routing | 2.5 ms / 32 B/op | 2.0 ms / 0 B/op |

The 160 is the closure plus the two delegates, the 32 is the closure on its own on paths that never reach `Match()`.

Together with #8973 this covers both points of the issue.

Closes #7664


src/Elastic.Clients.Elasticsearch/_Shared/Core/Infer/RoutingResolver.cs
```diff
@@ -54,7 +54,16 @@ public RoutingResolver(IElasticsearchClientSettings settings, IdResolver idResol
...
+		var joinField = GetJoinFieldFromObject(type, instance);
+		if (joinField is null)
+		{
+			return null;
+		}
+
+		// Avoid 'JoinField.Match()', its lambdas capture 'instance' and allocate a closure per call.
+		return (joinField.Tag == 1)
+			? ResolveId(_settings, joinField.ChildOption.ParentId)
+			: _settings.Inferrer.Id(instance);
```
flobernd: Both `JoinField.Match()` and `JoinFieldConverter.Write` switch on `Tag` and throw on an unrecognized value. The ternary treats every tag other than 1 as a parent, so a future third variant would be routed by the document's own id instead of failing. Unreachable today, but a switch keeps the three sites consistent at the same cost:

```suggestion
		switch (joinField.Tag)
		{
			case 0:
				return _settings.Inferrer.Id(instance);

			case 1:
				return ResolveId(_settings, joinField.ChildOption.ParentId);

			default:
				throw new Exception($"Unrecognized tag value: {joinField.Tag}");
		}
```

src/Elastic.Clients.Elasticsearch/_Shared/Core/Infer/RoutingResolver.cs
```diff
@@ -54,7 +54,16 @@ public RoutingResolver(IElasticsearchClientSettings settings, IdResolver idResol
...
 
 		static string? ResolveId(IElasticsearchClientSettings settings, Id id)
```
flobernd: `ResolveId` duplicates `Id.GetString(IElasticsearchClientSettings)`, which `JoinFieldConverter.Write` already uses for the same `ParentId`. `IdResolver.Resolve` returns null for a null instance, so `joinField.ChildOption.ParentId.GetString(_settings)` gives the same result for both the document-backed and the literal case, and the local function can go.

src/Elastic.Clients.Elasticsearch/_Shared/Core/Infer/RoutingResolver.cs
fresh55: Applied

src/Elastic.Clients.Elasticsearch/_Shared/Core/Infer/RoutingResolver.cs
fresh55: Applied, thanks

PR #8995, opened 2026-09-13
Fix routing inference for bulk update and delete operations

### LINKED ISSUE #8978: Bulk update operations get routing wrong
**1. fluent Update and UpdateMany do not send routing at all**

BulkUpdateOperationDescriptor.IdFrom sets Instance.Id, but it never sets Instance.IdFrom. BeforeSerialize reads IdFrom to work out the routing, so it finds nothing and the routing is skipped. NEST 7.x did set it and routing worked there.

**2. update sends `"routing":null` when the type has no routing property**

BulkUpdateOperation<T, TPartial> does `Routing ??= new Routing(IdFrom)`, and that is never null. index and create already check for this, update does not. delete has the check too, but BulkDeleteOperationDescriptor sets Routing in its constructor, so the check never runs.

**repro**

with RoutingProperty set

```
fluent UpdateMany    {"update":{"_id":"doc-1"}}
initializer          {"update":{"_id":"doc-1","routing":"team-a"}}
fluent IndexMany     {"index":{"_id":"doc-1","routing":"team-a"}}
```

without it

```
fluent UpdateMany    {"update":{"_id":"doc-1"}}
fluent DeleteMany    {"delete":{"_id":"doc-1","routing":null}}
initializer update   {"update":{"_id":"doc-1","routing":null}}
```

the first one is the bad one. on an index that uses routing, the update goes to the wrong shard and you get document_missing_exception. nothing in the error points at the client.

fresh55:
closes #8978

fluent `IdFrom` set `Id` but not `IdFrom`, so update never inferred routing. update and the typed delete descriptor also wrote `"routing":null` when the type had no routing property.

```
before  {"update":{"_id":"doc-1"}}                     {"delete":{"_id":"doc-1","routing":null}}
after   {"update":{"_id":"doc-1","routing":"team-a"}}  {"delete":{"_id":"doc-1"}}
```

index and create already resolved routing before assigning it. that check is now one helper on `BulkOperation`, used by all four.

`Delete<T>(doc)` and `DeleteMany<T>(docs, selector)` built the untyped descriptor from an id alone and lost routing and `_index` too, they now go through the same path.

`WithScript(script, upsert)` had the same gap, it now infers routing from the upsert too.

`BulkUpdateOperationWithScript<TDocument>` still inherits `ClrType => null` so it skips `_index` inference, left alone here as it would shift index resolution for existing callers.


flobernd:
Thanks for the PR @fresh55 , I'll have a look 🙂

flobernd (commented):
Thanks for the fix and the thorough write-up. The root-cause analysis matches what I see, and I verified the serialized
output of the fluent and initializer paths against `main` and this branch: every difference matches your description,
explicit routing still wins, and the plain-id, index and create paths are unchanged.

One small thing before merging: the three-line comment above the routing call in `BulkIndexOperation.SetValues` (and
its copy in `BulkCreateOperation.SetValues`) describes the trade-off that now lives on `BulkOperation.InferRouting`,
which has its own comment. Please trim them to the `Id` part or drop them.

Optional, only if you feel like it:

- `BulkDeleteOperationDescriptor<TDocument>(TDocument)` re-implements what `BulkDeleteOperation<T>(T)` already does;
  `base(new BulkDeleteOperation<TDocument>(documentToDelete))` would be enough.
- In `BulkUpdateOperation<TDocument, TPartialDocument>.BeforeSerialize`, the "first non-null of `IdFrom` / `Upsert`"
  is computed twice. Computing it once and passing it to both `Id` and `InferRouting` also avoids the array
  allocation.

One related gap: `BulkUpdateOperationWithScript<TDocument>` (the `WithScript(script, upsert)` factory shape) still does
not infer routing from `Upsert`, so it has the same wrong-shard problem this PR fixes for the other update shape.

While testing I also noticed two pre-existing problems that I will take care of separately: the typed
`BulkDeleteOperationDescriptor<TDocument>` derives from the non-generic descriptor base, so its fluent methods throw,
and the unconstrained `Delete<TSource>(TSource)` overload captures scalar ids such as `long` or `Guid`. Nothing for you
to do there, just mentioning it so it does not surprise you if you touch those types.

No need to add tests: the serialization tests under `tests/Tests` are currently excluded from the build. The two
failing checks are label-related and on our side, nothing for you to do there either.

fresh55:
Thanks for the review! Addressed all four. 🙂

flobernd (approved):
LGTM, thank you @fresh55 !

